PR #1, opened 2017-11-16
Add textutils.prompt

hugeblank:
Probably _**the most**_ necessary PR from CC to get implemented into CC-Tweaked. as described in CC's PR:

textutils.prompt is the brainchild of numerous people that have attempted to contribute to ComputerCraft. SquidDev has mentioned this in various PR's as "super-duper READ-O-MATIC 4000 with turbo text entry", and to be quite honest, that's what it is. It is the read function on steroids, if you will.
 
Features:
- All arguments are put into a table, so that they don't have to be put in any particular order, like how it is in read().
- Type error protection has been condensed into a more understandable form.
- Adds a few arguments, including those from read(), and others that would never make it into read():
  - replaceChar (_sReplaceChar)
  -  history (_tHistory)
  - complete (_fnComplete)
  - prefix (_sDefault)
  - limit (number): As defined in #387
  - newline (boolean): As defined in #386
  - completeBGColor (number): Changes the Background color of the suggested completion values
  - completeTextColor (number): Changes the Text color of the suggested completion values
  - filter (function): takes the current input string and feeds it to a function. The function can then manipulate the string and return it.
  - customkeys (table): a table consisting of all the "special" keys, any and all of which can now be redirected to any other key. Keys that can be redirected (if not redirected, their default is just the key name, prefixed by "keys."):
    - enter
    - up
    - down
    - left
    - right
    - backspace
    - home
    - delete
    - tab
    - end (["end"])

What's really great about this function being used over read is that it allows for both command lines, and applications with a GUI alike to have greater customizability, over read(). Hopefully in the future this may be used in replacement of read(). 
Along with this, my hope with prompt is to have people that have wanted a specific feature implemented into read(), be put here. As there is no particular order to the arguments, it's all subject to the user's need of it.

EDIT: FIRST :P

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/textutils.lua
@@ -1,3 +1,322 @@
+function prompt( _tOptions )
+    local tVerify = {replaceChar = "string", history = "table", complete = "function", prefix = "string", limit = "number", newline = "boolean", completeBGColor = "number", completeTextColor = "number", filter = "function", customkeys = "table"}
+    if not _tOptions then
+        _tOptions = {}
+    end
+    for k, v in pairs(tVerify) do
+        if _tOptions[k] ~= nil and type( _tOptions[k] ) ~= v then
+            error( "bad argument " .. k .. " (expected " .. v .. " got ".. type( _tOptions[k] ) .. ")", 2)
+        end
+    end
+    
+    term.setCursorBlink( true )
+
+    local sLine
+    if type( _tOptions.prefix ) == "string" then
+        sLine = _tOptions.prefix
+    else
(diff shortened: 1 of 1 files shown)


SquidDev:
I'm closing this as when the time comes, I'll just merge from the corresponding PR on the main ComputerCraft repo. I don't really want to be duplicating PRs between the mods.

As to merging the original PR, I'll have a look later today. I remember not being super happy with the code duplication but I think you've updated it since then.

hugeblank:
Alright understandable, but in the case that the PR on CC doesn't get merged, what will happen then?

Really there's not much that I haven't put my grubby fingers on comparing read() to textutils.prompt, except for the autocompletion section. everything else I have at the very least modified and tweaked.

PR #5, opened 2018-01-20
Networking API

SquidDev:
Very much work in progress, but a useful starting point at least. I'll write up the rationale/design for this when closer to release.

 - [x] Initial implementation of api, cable support, etc...
 - [x] Full block modems
 - [x] Targeted and networked peripherals. These can probably be put on the `IPeripheral` interface.
   - [x] Implement targeted peripherals where required.
 - [ ] MCMultipart and CBMP support. Maybe.
 - [x] Optimise. Can we cache the active peripheral? It's recalculated _a lot_.
 - [x] Test everything. Seriously. Just go through CCTweak's bug list to see how broken everything is.
   - Full block → normal modem chaining. Ideally just disable it.
   - Cross thread network access
   - Chunk loading/unloading
   - Cable/modem breaking/placing. Especially within the modem part, as that is especially susceptible to breakage.
   - Breaking/replacing peripherals and checking ID changes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,6 +59,8 @@ dependencies {
     deobfProvided "mezz.jei:jei_1.12:4.7.5.86:api"
     runtime "mezz.jei:jei_1.12:4.7.5.86"
     shade 'org.squiddev:Cobalt:0.3.1'
+
+    testCompile 'junit:junit:4.11'
 }
 
 javadoc {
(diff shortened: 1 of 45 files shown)


SquidDev:
There are several issues with Computronics' MultiPeripheral system which we'll have to solve:

 - [ ] Modem peripherals are no longer an instance of themselves so we cannot prevent wrapping other modems. Ideally this wouldn't be an issue anyway, so something we should look into fixing.

 - [ ] Network nodes are not attached to the peripheral. I think it might be a good idea to move those methods to `IComputerAccess`, this way people know what "network" they are called from.

PR #12, opened 2018-02-12
Implemented energy peripheral

CJBok:
Allow computers and modems to get energy information from adjecent / connected blocks. Ignore if block already contains peripheral.

Supported energy types:
- Forge Energy
- Redstone Flux

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,8 @@ deploy
 luaj-2.0.3/lib
 luaj-2.0.3/*.jar
 *.DS_Store
+/bin/
+.classpath
+.project
+*.launch
+.settings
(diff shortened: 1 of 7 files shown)


SquidDev:
Thanks for the contribution!

Whilst I do think that these sorts of peripherals are useful, I'd rather avoid having them in CC: Tweaked. In the long term, I'd like to see a more comprehensive OpenPeripheral-esque mod  bundled with CC, but for the time being mod integration is just being bundled [with Plethora](https://github.com/SquidDev-CC/Plethora).

Sorry, and thanks again.

CJBok:
No problem, i was trying to avoid creating the need for an extra block. I Also made sure the mod isn't depending on Redstone Flux. If Redstone Flux isn't installed the mod still works and of course Forge Energy is part of Forge.

And thank you for keeping CC alive and updated!

PR #15, opened 2018-02-24
Maybe fix monitors

Lignum:
With this PR, we now generate individual display lists instead of three at once, because that way it's guaranteed that we always call the right one.


Lignum:
Also, there is a `displayLists` `TIntHashSet` in `ClientMonitor`, it doesn't appear to be used?

SquidDev:
> Also, there is a `displayLists` `TIntHashSet` in `ClientMonitor`, it doesn't appear to be used?

I added that to track buffers and dispose of them on world unload, but removed it before merging :/.

Lignum:
Shall I get rid of that then?

SquidDev:
Nah, I'll try to remember to actually add said feature.

Lignum:
Alright.

PR #31, opened 2018-05-08
Amend typo in keys.lua

hugeblank:
Skimming through the CC API for an emulator project and found this bug. I decided to take this game-breaking issue into my own hands and repair it. You're welcome CC.

In other words this is the worst reason to receive a contributor tag, so if you ( @SquidDev ) want to close this and fix it yourself and not actually give me a pity tag, feel free. XP


SquidDev:
It's probably going to be better to PR this into CC instead and I'll merge it over. It's probably also adding something like this to ensure backwards compatibility

```diff
 keys["return"] = keys.enter
+keys.scollLock = keys.scrollLock
```

hugeblank:
Alright, sounds good.

Issue #23, opened 2018-04-15
Mod integrations


SquidDev:
- [x] **EnderIO**: Bundled cable support. Possibly also a "wired network conduit".
 - [x] **Project: Red**: Bundled cable support.
 - [ ] **RSWires**: Bundled cable support.
 - [ ] **MCMultipart**: Support for wired modems and cables. I suspect this'll have to wait for 1.13, as `BlockGeneric`/`TileGeneric` rather makes this impossible.
 - [ ] **SimpleLogic wires**: Bundled cable support.

SquidDev:
Closing this for now: RSWires and SimpleLogic wires are sufficiently painful to integrate that I'm rather unwilling to put any more time into it.

It's worth noting that EnderIO is gone again SleepyTrousers/EnderIO@764789611d8aa83fe97c3e15a7a0e8a391f15b49. Ya to CC being maintained.

Issue #53, opened 2018-07-24
Reducing ticking


SquidDev:
Some commonly placed blocks in CC(:T) have tiles which implement `ITickable`. Despite they do very little work, it does adds substantial overhead to each tick which would be nice to avoid. The main culprits are as follows:

## `TileCable`
https://github.com/SquidDev-CC/CC-Tweaked/blob/b8957cab5c3c3829f4931dd21faf39323e45fa87/src/main/java/dan200/computercraft/shared/peripheral/modem/TileCable.java#L468-L486

Most of this is just "on first tick" logic, so could potentially be shifted somewhere else?

## `TileMonitor`
https://github.com/SquidDev-CC/CC-Tweaked/blob/b8957cab5c3c3829f4931dd21faf39323e45fa87/src/main/java/dan200/computercraft/shared/peripheral/monitor/TileMonitor.java#L148-L186

Only the primary monitor really needs to tick. While I'd like to avoid having "main" and "non-main" monitor TEs, we should investigate ways to only have one ticking monitor.

## `TilePrinter`
This currently doesn't tick, though may not actually be threadsafe - we're calling `.setAnim` on the computer thread.

## Others?
Could we potentially do the same for other modems - most just run logic on the first tick and then never again after that.

PR #55, opened 2018-07-27
WIP: The road to 1.13

SquidDev:
This is will never be merged, as development will happen on another branch, I just felt it was a good way of tracking progress for 1.13 and a way to document the process as it goes on. This PR will hopefully end up as a bit of a devlog in terms of what's being done.

The port is effectively split into three phases, "cleanup", "flattening" and "updating". The first two can be done on 1.12, which allows us to make a start before Forge is ready (and goodness, do we need it).

## Cleanup
Due to CC's lack of development, I'm no longer going to try to push my Minecraft related changes up to the original repo. As I no longer need to keep things in sync, this should allow me to make some more comprehensive changes to the code.

As the likelihood of being able to update the world is next to nothing, I'm taking this opportunity to remove a reasonable amount of "legacy" functionality. For instance, the original turtle item, integer upgrades id and legacy disks.

I'm also hoping to shift away from putting as much logic on the tile entity. `TileGeneric` does allow for some nifty things, but using the tile for all logic, instead of the block/block state, limits us (see #23). I think 6c6891d4ea6ab4b804c35ff4c756f9dec71022a1 is a great example of how moving stuff "down" in the hierarchy can allow us to simplify a lot of things.

## Flattening
As block metadata and item damage is effectively gone in 1.13, the idea of "sub-blocks" is mostly dead. This allows us to split some of our god-blocks up. For instance:

 - Computers can be split into normal/advanced, and use NBT for computer id (instead of damage)
 - Pocket computers can be split into normal/advanced
 - Printouts can be split into separate items
 - Peripherals can be split up. This is probably the most useful change, though also the hardest to implement.

It's worth noting that in some cases, we don't need separate item classes or anything - we can just instantiate the same item with different fields.

We are still heavily dependent on item damage in some places (`getFamily(int)`, etc...), so this will require some a lot of work and a lot of changes.

It may also be making translation keys be dependent on the resource name at this point, rather than switching on item/item type.

## Updating
Honestly, I've no clue how much effort this will require. I'd like to imagine that once the flattening is done, most changes will be purely cosmetic (mappings changes: ughr), but it is all very dependent on what Forge does.

I know rewriting CC:T's command system will be a bit of a pain, due to our home-grown command system. I haven't looked super deeply into brigadier so we'll see how that goes.


vpontin:
I just wanted to leave my eternal gratitude for what you are doing for CC. Thank you very much!
Man, even if someday dan returns and updates the original repo i still would use CC:T.

src/main/java/dan200/computercraft/shared/computer/blocks/BlockComputerBase.java
```diff
@@ -68,13 +75,43 @@ public final ComputerFamily getFamily( IBlockAccess world, BlockPos pos )
...
+    public final int getWeakPower( IBlockState state, IBlockAccess world, BlockPos pos, EnumFacing oppositeSide )
+    {
+        return getStrongPower( state, world, pos, oppositeSide );
+    }
+
+    @Override
+    public int getBundledRedstoneOutput( @Nonnull IBlockState state, @Nonnull IBlockAccess world, @Nonnull BlockPos pos, @Nonnull EnumFacing side )
+    {
+        TileEntity tile = world.getTileEntity( pos );
+        return tile instanceof TileGeneric && tile.hasWorld() ? ((TileComputerBase) tile).getBundledRedstoneOutput( side ) : 0;
```
SquidDev: This needs to be fixed.

SquidDev:
I thought I'd do a quick write up of what work I did yesterday, as well as some of the rationale and hopes for the 1.13 preliminary.

This is effectively going to act as a block, so here be walls of text! Any thoughts/feedback are always much appreciated, as I'm sure there's some issues I'm unaware of.

That being said, it's probably worth noting I'm only interested in cleaning up the _Minecraft_ side of things - the actual computer portions are an entirely separate project here.

## Legacy code
ComputerCraft is an old mod now and, in the interest of backwards compatibility, has an awful lot of legacy functionality. For instance, turtles originally had very few possible upgrades, and so the available upgrades (and computer ID) were stored in item damage. When upgrades were further developed, this was stored in item NBT instead, and thus a separate block/item was added to disambiguate the two.

In the interests of preserving worlds between versions, CC has never removed this code. We nominally have support for loading up Minecraft 1.2 worlds, even though the world formats are no longer compatible.

So far the following have been removed:
 - Legacy turtle item and block
 - Integer turtle upgrade ids (replaced with `ResourceLocation` around 1.8)
 - Legacy disk (stored disk id in item damage)
 - Computers storing their id in damage
 - Block/item name remapper. This was added in order to convert worlds between 1.10 and 1.11, but given the other breaking changes it can be considered redundant.

There's still a couple of other places where things could be cleaned up, though they can be handled at a later date:
 - Pocket computer upgrades being stored as a short
 - "Old style" computer ids on NBT
 - Incorrect and inconsistent spelling of upgrade/item/block names

## Flattening
CC stems from the beta days, where we were limited to 256 blocks and so one tried to cram as much functionality into one class. This has lead to a couple of "god blocks", most noticeably peripherals.

1.13 has changed two things as far as this is all concerned:
 - Block ids are effectively infinite, and so there is no longer a reason to cram everything together
 - Item damage is gone, so sub-items based on damage are no longer sensible.

Due to the lack of item damage, we now have to separate things into individual item. However, this does not mean we require separate item classes. For instance, printouts (page, pages and books), computers, pocket computers and turtles all have one `Item` class for their group, just instantiated with different arguments.

This actually allows us to strip out a lot of extra code, such as determining the translation name - we can just set it directly (or build it from resource location, which is what we and 1.13 does).

We also have to do the same for blocks. There's only two cases where this needs doing though: normal computers (which was incredibly easy) and peripherals (which really isn't).

### Peripherals
`computercraft:peripheral` currently handles the current sub-blocks:
 - Disk drives (direction is stored in block metadata)
 - Wireless modems (direction is stored in block metadata)
 - Monitor (direction stored in NBT)
 - Printer (direction stored in NBT)
 - Advanced monitor (direction stored in NBT)
 - Speaker

`BlockPeripheral` and `TilePeripheralBase` are an ungodly mess which try to resolve all these conflicts, while also adding lots of extra functionality which complicates anybody trying to subclass it.

It's worth noting that `BlockPeripheral` can only be rotated horizontally, meaning you have beautiful variants like `WirelessModem`, `WirelessModemUp`, etc...

Speakers and wireless modems were pretty easy to shift out. Speakers as they have very little special handling, modems as they can be merged with the existing ender modem code.

I'm currently trying to shift away from `TilePeripheralBase`, which introduces complexities in this porting process. We need to rethink how we handle the visual state of blocks (such as a modem's on/off state) and how that is synced to the client. Ideally we could reduce the number of ticking tiles (as most of them do little more than `if(changed) syncToClient();`), though that is most definitely a separate project.

## Reducing indirection
The third change, and one I've made less progress on, is reducing the amount of indirection within blocks and tiles. Many `BlockGeneric` methods defer their handling to the corresponding `TileGeneric`. While this does offer greater flexibility, it's often redundant. For instance, only computers respond to redstone, and only cables have custom collision boxes.

Further more, this often leads to a lot of bouncing back and forth between the two classes. For instance, in order to determine a cable's collision boundaries, we access the tile, get the current peripheral type (which requires a block and blockstate lookup), get the direction (another block state lookup), determine which neighbours are cables and so can be connected to, and then return the block. Meanwhile, if this was implemented on the block level, we could find all that information in the block state without having to even think about it!

Most of my current progress on this has been moving things "down" the hierarchy (so from `BlockGeneric`/`TileGeneric` to `BlockCable`/`TileCable`). The next step is to obviously move things exclusively into the block class. In some cases this is very easy (explosions), some a bit of a faff (bounds/collisions) and some entirely undesirable (redstone and neighbour changes on computers).

Lemmmy:
> We need to rethink how we handle the visual state of blocks (such as a modem's on/off state) and how that is synced to the client.

How do furnaces do it now? Is there any reason we can't do it a similar way here?

SquidDev:
> How do furnaces do it now? Is there any reason we can't do it a similar way here?

I believe that is stored in the block state. While this is the only way of presenting it, there's a few issues:

 - This _may_ be persisted to disk, which we really don't want as it resets on world load. I believe this can be avoided by using `.getExtendedState` though.
 - Furnaces do work every tick, so all visual state management is handled there. Most blocks don't do any work on their tick (it's all done in the CC thread), so we have to make them ticking _just_ to sync state, which is a little bit ugly. This is mostly an issue with cables and monitors, as you end up having a lot of them (see #53).

steamp0rt:
Up the darn event cache from 255 if possible

SquidDev:
> Up the darn event cache from 255 if possible

> That being said, it's probably worth noting I'm only interested in cleaning up the Minecraft side of things - the actual computer portions are an entirely separate project here.

Thanks for reading things in such detail.

steamp0rt:
Whoops. My bad. I didn't get any sleep, so I may have skipped over some things

_by some things, i mean pretty much everything_

Lemmmy:
@steamp0rt Make a separate issue.

SquidDev:
OK, another status update because apparently I like the sound of my own voice. There hasn't been much progress, though I have successfully split out monitors - only cables and printers to go!

Monitors were actually surprisingly easy. Storing orientation as part of the block state (and so using `EnumFacing` rather than magic integers) allowed us to do get rid of an awful amount of code, which is always nice.

One of worries about the flattening is that each possible block state takes up a new ID. This is fine most of the time, but is problematic when it comes to something like monitors - there's 192 possible states (3 * 4 * 16^4), and so 192 possible block ids. Cables are even worse - 2^6 * 7 * 4 (cable connections, modem direction and state states - this is hitting 2000 pretty fast.

It's possible Forge will provide a mechanism around that, or we should bodge some things (such as modem state) with the extended state.

Lemmmy:
ComputerCraft won't be the only mod to implement cables, so I'm sure the Forge guys will be thinking of something for this. Maybe there's somewhere you can drop by and ask?

SquidDev:
So MCP snapshots were released today and so I inevitably loaded this up under it. Some things I'd like to/might do before hand:

 - Reformat everything. This is going to make it impossible to move code between 1.12 and 1.13 versions, but when doing large refactors I end up doing it compulsively.
 - [ ] Remove `PeripheralType`. _Maybe_. It doesn't really need to exist any more.
 - [ ] Remove the block-state sensitive `TileEntity` getter.
 - [ ] Try to use extended states where possible for rendering.

I'll probably add to this list as time goes on. I'm probably going to wait a few more days for mappings to stabilise before doing any more work on this.

SquidDev:
So I've reached the point where I don't thin there's any immediate refactoring, so I guess now is the time to start testing these changes:

---

## Testing CC:T, flattened-edition
I'm mostly interested here in behaviour which is different between master and this branch. So the best thing here is probably to run master and this branch at the same time, perform an action in both, and verify the result is the same. The sorts of things I think it'd be worth testing:

 - Redstone interaction
 - Block drops/creative-mode picks
 - Placing blocks (especially wired modems, wireless modems, and cables).
 - Pretty much everything to do with peripherals. Inserting/extracting from printers and drives, monitors placement, etc...

If you find some inconsistency between the two versions, post a comment below (with screenshots if possible). If you find some behaviour which is a little odd (but works on master), then just create a new issue.

### Running CC:T
You'll need Git and the JDK installed in order to do this:

#### For `master`:
 - `git clone -b master --depth=1 https://github.com/SquidDev-CC/CC-Tweaked.git CCT-old`
 - `cd CCT-old`
 - `./gradlew runClient`

#### For `mc-1.13.x`:
 - `git clone -b mc-1.13.x --depth=1 https://github.com/SquidDev-CC/CC-Tweaked.git CCT-new`
 - `cd CCT-new`
 - `./gradlew runClient`

BombBloke:
Placed turtles always start out facing towards me (when they should be facing away). `forward`/`back`/`up`/`down` each orientate them to face *north* before performing the movement. This can cause turtles to teleport - eg, if you attempt to move east from x200/z200 to x201/z200, the turtle will warp to x201/z201 and then move north. I guess this all has something to do with `BlockDirectional` being turfed?

Peripheral calls seem to be working ok, best I can make out, including when moving turtles are being used as the peripherals.

Wired Modems are incorrectly labelled as Network Cables.

Actual Network Cables can't be placed against non-solid blocks, such as "other Network Cables". If I *do* place a Network Cable, the *next* block I place (of any type) is removed from my inventory as if I were in Survival, *even while in Creative.* This isn't triggered when I "merge" a cable with a modem, but is when placing cables on their own or next to other cables.

Destroying a cable block attached to an active modem no longer seems to de-activate that modem.

The Disk Drive GUI shows a title of `tile:computercraft:drive:name`.

Monitor "displays" have terrible Z-fighting issues that aren't present in the old build, but they otherwise appear to be ok. Tested rendering in all twelve orientations and with a few different sizes.



Bounding boxes look ok to me. Middle-click and named-device drops seem to work in Creative.

Stuff I haven't played with at this time:

* Redstone  
* Pocket computers  
* "Equipped" peripherals  
* Crafting / Recipes

redfast00:
Will turtles be able to be [waterlogged](https://minecraft.gamepedia.com/Waterlogging)? For example when they move into water?

SquidDev:
Oh wow, thank you so much @BombBloke! I could have sworn I tested turtle movement after the directional change, but highly likely I broke it doing something else. I'll get right on fixing those.

> Will turtles be able to be waterlogged? For example when they move into water?

I'm hoping all non-solid blocks (modems, cables and turtles) will be waterloggable, but I honestly haven't looked into it yet.

As a bit of a status updated, the 1.13 mappings were officially released a few days ago, so I've been going round tackling the low hanging fruit. I'm "down to" about 250 compile errors (from the original 700 odd), though it's hard to tell how much actual progress has been made until Forge is released. I might look into doing a POC build using [Rift](https://minecraft.curseforge.com/projects/rift), purely to test out some of the vanilla changes. That's still a _very_ long way off, so we'll see how this all goes :).

SquidDev:
@BombBloke I've pushed a commit which fixes most of the above issues. Monitors will require a little bit more of an investigation into what I've broken there.

> Destroying a cable block attached to an active modem no longer seems to de-activate that modem.

I'm unable to reproduce this, which is rather odd.

BombBloke:
> Monitors will require a little bit more of an investigation into what I've broken there.

They put me in mind of [this old bug](https://github.com/dan200/ComputerCraft/issues/87), which unfortunately was reported before Dan made his source available. I'm seeing the problem with all segments along all orientations this time, though.

>> Destroying a cable block attached to an active modem no longer seems to de-activate that modem.

> I'm unable to reproduce this, which is rather odd.

I can no longer reproduce it myself. There might be a condition I didn't notice before, but I suspect you may've resolved it as a side-effect of working on the other issues.

I do notice that whenever I re-load my world, all active modems appear de-activated until I disable and then re-activate them. This is purely cosmetic: the connected computer still detects the remote peripherals.

johnneijzen:
this may help it is list changes from 1.12 to 1.13 like metadata, rendering changes, and other stuff
https://gist.github.com/williewillus/353c872bcf1a6ace9921189f6100d09a

SquidDev:
I'm closing this, as much of this work has either been applied to master or my new 1.13 branch. I gained a lot of useful knowledge from this PR, so thank you everyone for helping test - there's definitely some things BombBloke found that I know are going to crop up again.

PR #64, opened 2018-08-14
Update edit.lua with my version

TagFreettle:


SquidDev:
Hai, thanks for the contribution. As the README says, my general stance here is for people to PR against the original CC repo if at all possible, and I'll merge it across. 

This means more people have an opportunity to comment, and there's a greater chance of your changes getting into the original mod.

TagFreettle:
What? What happened?

TagFreettle:
But I sent a pull request already and it never came through. What happened?

vpontin:
Link that pull request here and i think Squid can analyse

PR #65, opened 2018-08-22
Prevent stack overflows when using turtle.place() with a full inventory

apemanzilla:
Using `turtle.place()` (e.g. to shear a sheep) when the turtle cannot accept dropped items causes a stack overflow - the drop consumer will attempt (and fail) to store the item, then spawn a new item stack, triggering the drop consumer again. This PR fixes this issue by changing the `dropConsumer` to be a `Function<ItemStack, ItemStack>` returning the remaining items that couldn't be stored. A list of remaining items is then returned from `clearDropConsumer`, so that they can be spawned as dropped items in the world, after the consumer has been removed, preventing a stack overflow.

`turtle.dig()` and `turtle.attack()` were already safe, but this PR also modifies them slightly to use this new `Function` system, for consistency and to avoid code duplication.

### DIFF
--- a/src/main/java/dan200/computercraft/ComputerCraft.java
+++ b/src/main/java/dan200/computercraft/ComputerCraft.java
@@ -98,7 +98,7 @@ import java.net.MalformedURLException;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.*;
-import java.util.function.Consumer;
+import java.util.function.Function;
 import java.util.zip.ZipEntry;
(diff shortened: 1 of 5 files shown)


SquidDev:
Good catch, thank you! Not sure how I hadn't picked that up before.

As I'm sure you could tell, we handled this for turtle digging/attacking, though didn't fix placing at the same time. What you've done is a much nicer fix anyhow 👍.

PR #68, opened 2018-10-02
Create de_de.lang

Vexatos:
How dare I. Absolute blasphemy.

Roughly based on https://github.com/dan200/ComputerCraft/pull/516 except not really.


SquidDev:
Thank you for putting this together (and to Lignum for putting an eye over it). 

As Lig says, adjectives are a tad sub-optimal, but the existing system is really Anglo-centric, and I'm not sure to work around it in a general way. Happy to chat on IRC if you've got ideas!

Vexatos:
The adjective cheat is what I came up with in cases where custom grammar isn't implemented.

PR #86, opened 2018-12-17
WIP: 1.14 Fabric port

SquidDev:
Like #55, unlikely to ever merge this, but more of a tracking issue to see what still needs to be done. Off the top of my head:

### Bug fixing and missing features
 - [ ] Cables/wired modem multipart functionality (block breaking, bounding box, block broken)
 - [ ] Proper wired modem functionality when breaking neighbouring blocks: we need to drop an item and update the network.
 - [ ] Inventory desyncs on printers and disk drives. Honestly, can we clean up inventory code even a little?
 - [ ] `/computercraft copy`
 - [ ] Go through all times we request a block update, and see if we can narrow the scope (we're more reliant on block states, so can get rid of explicit "send TE to client"/re-render calls?).
 - [ ] Add dan200/ComputerCraft#331 into the main `IPocketAccess`.
 - [ ] Turtle model transforms. It might be nice to clean up how we do loading too.

### Rethink
 - [ ] How do we handle command computers under the new command system?
 - [ ] Can we better handle mounting the ROM/bios from data packs? Ideally, we'd actually pay attention to whether they are enabled or not.
 - [ ] What API compatibility, if any, are we prepared to break. At this point it's unlikely there will be another CC fork, so it's possible to take some liberties, but I'm not sure how comfortable I am with doing that, so it's worth discussing.


DaveyL2013:
How do I run this in 18w50a? Modlist does not see it... I compiled it from the instructions on the main page for 1.14, and it opens in 1.12? Thx, DaveyL2013.

SquidDev:
@DaveyL2013 You're better off grabbing a build from [CurseForge](https://minecraft.curseforge.com/projects/cc-tweaked/files?filter-status=1&filter-game-version=1738749986%3A64806) instead. We're using Fabric as our modloader, so you'll need to [install that too](http://fabric.asie.pl/use/).

Issue #33, opened 2018-05-14
Limiting HTTP


SquidDev:
ComputerCraft added very rudimentary limiting system to the filesystem (a fixed number of handles per computer), but no similar system was added to the HTTP API. It would be nice to add some more fine grained control to the `http` API, to prevent abuse by malicious/badly implemented programs.

One nice thing about HTTP is that, as it is asynchronous, we can delay requests instead of outright rejecting them. I suspect this will be required as many installers/updaters will download multiple files in parallel.

Configuration options could be server wide and per-computer. While the latter is more important (normally it will be a few computers playing up), it may be useful to enforce some global constraints too. If so, we will have to establish some level of scheduling, prioritising requests from computers which don't access the network very often, etc...

## Request limits
These limits are designed to control individual requests. They are not going to prevent abuse of the HTTP API, but should stop people doing insanely stupid things with it.

### HTTP Requests
 - **Max requests:** The maximum number of requests a computer can be making at one time. Additional requests can be pushed to a queue.
 - **Max download size:** The maximum size an incoming response can be before the connection is terminated.
 - **Max upload size:** The maximum size an outgoing request can be.

### Websockets
 - **Max websockets:** The maximum number of open websockets a computer can have. Unlike HTTP requests, it is probably save to error if this limit is reached.
 - **Max packet size:** The maximum size a websocket message can be (incoming and outgoing).

It's also worth noting that many of these options could be configured on a per-domain/per-IP-range level. For instance, you could allow downloading larger files from GitHub, or limit the upload size to Pastebin.

## Rate limiting
While the above has some utility, it doesn't prevent people making lots of requests in a short period of time. Instead, we should also investigate ways of limiting the amount of bandwidth a computer may consume.

--- 

If people have ideas, suggestions or if there's anything else we could look into limiting, do comment!

hugeblank:
An obvious first question: 
What would each of those limits be by default, and would the limits be enabled at all by default?

LemADEC:
We need to limit the request rate and bandwidth usage, at least globally. When a global limit is reached, I would advise to prioritize the 'good' computers. For example, the computer doing the most request/bandwidth throws a LUA exception, and a console log reports it. That way the player and server owners are aware of the issue.

nevercast:
[...]

PR #98, opened 2019-01-06
WIP: Http rework

SquidDev:
The aim of this is to rewrite the implementation of the HTTP API to be more powerful and efficient. There are several key changes to be made here:

 - [x] Make HTTP requests using Netty, instead of a `HttpURLConnection`. While this does involve writing more code, it provides greater flexibility in handling redirects (we can now do http to https) and methods.
 - [x] Implement the restrictions mentioned in #33. For now, we'll just have per-computer limits.

### Further steps to be made before release
 - [ ] Think about how we should handle cleaning up resources. I do worry that `ManagedResource` is a bit over-kill, but we need some way of measuring resource lifetimes.
 - [x] Verify that all resources are cleaned up as expected. Make sure all `ByteBuf`s are released as expected.
 - [x] Test thoroughly - we need to make sure there's no regressions!
 - [ ] Can we unify our `WeakReference`/resource handling with that of file systems?


SquidDev:
It looks like http calls are no longer tracked - only upload and download.

SquidDev:
One other thing worth adding may be returning the response status text along with the code. I discussed this in dan200/ComputerCraft#515, but didn't add it in 5bf9f9e3c5ae56e7fb4bcac3efbaaf7eb7d58094.

SquidDev:
One of these days I'll remember to remove the `WIP: ` prefix first.

Issue #114, opened 2019-02-14
Rate limits, scheduling and preemption


SquidDev:
This has been kicking around my head for a couple of weeks, especially since 12e82afad235f507d9dc2d118914c6aa4da72118 dropped, and given we just had an Indicent™ on SwitchCraft, I thought it might be as good a time as any to write up an issue on this:

Basically, we need a better way to prevent computers running amok. Or rather, when computers run amok, we need to reduce the impact that has on the rest of the server (and ideally on other computers). I think the best way in order to do this is as follows:

 - Introduce global limits for the amount of server time ComputerCraft can consume in a tick.
 - Allow pre-empting computers before they actually yield, reducing the risk of people blocking the computer thread.
 - Add some form of scheduling, so computers which do little work get server or computer time are not pushed aside by those which do a lot.

As far as work on the computer thread goes, it should be pretty trivial in a conceptual sense. The actual `Computer` implementation is a bit of a concurrency mess right now, so it'd be good to clean that up first. As far as scheduling goes, I think we can get away with something simple like waited-fair queuing (as used by Linux's CFS) - effectively pick whoever has used the least resources so far.

Server time is a little more complex. Unlike computers, tasks are executed in multiple places - turtles (on TE tick, executes commands for that computer), `MainThread` (in one batch for multiple computers) and on Plethora peripherals (on TE tick, commands tied to multiple computers). We need a way to spread time fairly between computers, while keeping as close to our budget as possible\*.

I haven't found a satisfactory solution to this yet, without introducing high latencies or massive unfairness. Though, as hopefully most of the time we'll be well within our budget, I think we mostly need to worry about the case where a couple of computers are hogging all execution time.

\* We're always going to go _above_ our budget, as we don't know how long something will take before we do it.

It should also be possible to extend some of this code to the bandwidth limiting areas of #33, though I haven't fully thought though all of that yet.

SquidDev:
We've implemented the computer thread portions of this issue. I need to have a serious think about how best to handle the server tick scheduling - any ideas are welcome!

SquidDev:
Well, this'll do for now.

PR #119, opened 2019-02-18
Preempting computer execution

SquidDev:
This implements computer-thread preemption or pausing, as discussed in #114.

## Summary of changes
 - `CobaltLuaMachine`/`ComputerExecutor` can now be paused - this suspends the machine via a debug hook. When doing work again, we resume the machine, rather than starting a new task.

 - `ComputerThread` now has an implementation of CFS:

   We aim to execute one task from every compute within a 50ms window (`ComputerThread.DEFAULT_LATENCY`). Thus, each computer is run with a timeout of (`latency / (tasks + 1)`). Once that timeout is reached (or the task is finished), we pause the machine, and add it to the queue.

   In order to ensure fair sharing of resources, the queue is sorted by the "virtual runtime", which is effectively computed by how much time it has received, weighted by how many other computers also want to do work. The more tasks to be queued, the less weight the current time is worth (as it is given a smaller share of the "ideal" processor).

## To improve before merging
 - [x] Find good defaults for the above parameters.
 - [x] Rework the documentation of `ComputerThread` and `TimeoutState`, as we've tweaked these a little.

### DIFF
--- a/src/main/java/dan200/computercraft/core/computer/ComputerExecutor.java
+++ b/src/main/java/dan200/computercraft/core/computer/ComputerExecutor.java
@@ -104,6 +104,20 @@ final class ComputerExecutor
     volatile boolean onComputerQueue = false;
 
     /**
+     * The amount of time this computer has used on a theoretical machine which shares work evenly amongst computers.
+     *
+     * @see ComputerThread
+     */
+    long virtualRuntime = 0;
+
+    /**
+     * The last time at which we updated {@link #virtualRuntime}.
+     *
+     * @see ComputerThread
+     */
+    long vRuntimeStart;
+
+    /**
      * The command that {@link #work()} should execute on the computer thread.
      *
      * One sets the command with {@link #queueStart()} and {@link #queueStop(boolean, boolean)}. Neither of these will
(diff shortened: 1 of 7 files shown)


SquidDev:
I'm probably going to spin the `ComputerExecutor` changes off into their own branch first, so it's easier to review preemption when it's finally production-ready.

However, before then, I think it's worth reviewing how `IPeripheral.attach` and `IPeripheral.detach` are called. `PeripheralAPI` schedules them to be attached/detached on the computer thread. I'd assume this was to ensure thread safety, however cables will sometimes attach/detach on the main thread, so that's probably not the case.

I'll try to get in touch with Dan about this, and see if there's a reason for it. If not, we're probably safe to run them when the event handler is called. Note, this will still sometimes call it on the computer thread, mostly when unloading the world.

SquidDev:
Small summary of the changes which have occurred since I last commented:

 - The computer thread changes have been added to master. It's worth noting that the timeout code is very different to before - soft aborts (and so pauses) are handled within the `CobaltLuaMachine`.
    Related to that, peripherals are now attached on the main thread.
 - Cobalt now allows suspending anywhere, which means we can also look into pausing things within the `poll` method.

I'm going to look into seeing what I can cherry-pick from 7004b65 and e8542088c6ba018f553c88fd5266b8115dfdcb59, but this may end up being a manual job given how much has changed.

PR #120, opened 2019-02-19
Minor palette improvements

SquidDev:
This cherry picks the backwards compatible parts of dan200/ComputerCraft#433, namely:

 - `term.nativePaletteColo(u+ε)r`. Note, this has been changed slightly: the methods are defined on the main `term` API, rather than the redirect object.
 - Split `colours.rgb8` into `colours.packRGB` and `colours.unpackRGB`. We still keep the original method for the sake of backwards compatibility.

To discuss:
 - Is it sane to put `nativePaletteColour` on the root term object? I can't see a situation where a redirect would have different colours, so this seems sensible to me.
 - `packRGB` or `packRgb`? We use all capitals for initialisms already (`os.loadAPI`), so it's consistent, but IMO looks a little ugly.


Lemmmy:
> Is it sane to put nativePaletteColour on the root term object? 

Yes. It's just like any other constant, I guess

> packRGB or packRgb?
IMHO, packRGB is fine, packRgb is ugly.

Lignum:
Looks good to me.

PR #123, opened 2019-02-21
Make Multishell Scrollable

Wilma456:
You can now Scroll through the Programlist, if the list bigger than your Screen.

multiscroll

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/advanced/multishell.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/advanced/multishell.lua
@@ -88,8 +88,8 @@ local function cullProcess( nProcess )
                 selectProcess( 1 )
             end
         end
-        if nScrollPos > #tProcesses then
-            nScrollPos = #tProcesses
+        if nScrollPos ~= 1 then
+            nScrollPos = nScrollPos - 1
         end
         return true
     end
--- a/src/main/resources/assets/computercraft/lua/rom/programs/advanced/multishell.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/advanced/multishell.lua
@@ -8,6 +8,8 @@ local nCurrentProcess = nil
 local nRunningProcess = nil
 local bShowMenu = false
 local bWindowsResized = false
+local nScrollPos = 1
+local bScrollRight = false
 
 local function selectProcess( n )
     if nCurrentProcess ~= n then
@@ -86,6 +88,9 @@ local function cullProcess( nProcess )
                 selectProcess( 1 )
             end
         end
+        if nScrollPos > #tProcesses then
+            nScrollPos = #tProcesses
+        end
         return true
     end
     return false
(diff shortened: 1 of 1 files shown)


SquidDev:
This looks to be the same as dan200/ComputerCraft#408 - it's definitely something I'd like to have, but would it be possible to implement [this comment](https://github.com/dan200/ComputerCraft/pull/408#issuecomment-348596130)?

Wilma456:
It is the same PR. I'm moving my PR's to this Repo, because Dan200 stoped working on CC.

I will look how I can make your requested change.

SquidDev:
> I'm moving my PR's to this Repo, because Dan200 stoped working on CC.

If there's a PR you want me to review on the original repo, I can merge it from there. I've done that for several others already.

Wilma456:
Should be fixed now.

Here's a List of my PR's from the original Repo:
dan200/ComputerCraft#498
dan200/ComputerCraft#495
dan200/ComputerCraft#468
dan200/ComputerCraft#437
dan200/ComputerCraft#428
dan200/ComputerCraft#412
dan200/ComputerCraft#390

SquidDev:
Looks good to me, thanks!

SquidDev:
I think I've merged a couple of those already, the list I've unmerged ones I've got is as follows:

 - dan200/ComputerCraft#422 - See my comment on the PR there.
 - dan200/ComputerCraft#437 - I'd really like to improve Lua's pretty printer, but I'd really like to handle things like recursive tables too. [I've made some attempts in MBS](https://github.com/SquidDev-CC/mbs#improved-serialisation) to make a nicer printer, but it's a long way from being ready for CC.
 - dan200/ComputerCraft#462 - I need to revisit this one. On one hand, I'm rather loathe to leak internal IDs to the user. On the other, it's definitely useful.
 - dan200/ComputerCraft#468 - See my comment there. We still need to think about how this works with typing latency and whatnot.
 - dan200/ComputerCraft#495 - Maybe use `multishell.max_title_length`, and then take into account the whole length when trimming the file name? Otherwise it's definitely a good change!
 - dan200/ComputerCraft#498 - In principle it's a good change, but I might want to look into defining a "standard" for how we handle switches/flags first. CC historically hasn't had programs which take arguments like that.
 - dan200/ComputerCraft#509 - I'm really not sure about this, as the method for detecting functions is a little fragile. Better autocompletion would be fantastic, but worth discussing what we want from it first.

Lemmmy:
Only looked at the first 3 PRs for now, will look at the rest later. I can't comment on the original repo.

* [dan200/ComputerCraft#422](https://github.com/dan200/ComputerCraft/pull/422) - I personally feel like this is the completely wrong approach to take here. I feel a better solution would be making the shell scrollable, like MBS does.
* [dan200/ComputerCraft#462](https://github.com/dan200/ComputerCraft/pull/462) - Assuming that the internal IDs are the same as the item IDs pre-equip, e.g. `minecraft:diamond_pickaxe`, then I don't think there's anything wrong with exposing them; the values would be no different do `turtle.inspect()` or `turtle.getItemDetail()`, both of which are exposed. If they are unique internal IDs though, then I agree this would be bad to expose. However, I think for a lot of cases, relying on the IDs of the equipped items isn't a good solution; you should rather rely on the behaviour that they provide, because different items can provide the same behaviour. For example, all pickaxes provide the `.dig()` function. You would have to hardcode a list of pickaxes, or use some naïve string searching (which isn't guaranteed to be accurate), to rely on getEquip for most cases.

Wilma456:
I will look after the PR's and I may open new PR's here.

PR #124, opened 2019-02-21
Improved autocompletion for set.lua

JakobDev:
If a setting is boolean, set.lua will now complete true or false.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/startup.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/startup.lua
@@ -163,9 +163,14 @@ local function completeChat( shell, nIndex, sText, tPreviousText )
         return completeMultipleChoice( sText, tChatOptions )
     end
 end
+local tBoolean = { "true", "false" }
 local function completeSet( shell, nIndex, sText, tPreviousText )
     if nIndex == 1 then
         return completeMultipleChoice( sText, settings.getNames(), true )
+    elseif nIndex == 2 then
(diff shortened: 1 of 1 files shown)


SquidDev:
I think my one worry here is that there are some (non-CC) programs that have settings which take `false` or some string, rather than being `false` or `true`. I don't want to offer completions for values which aren't actually legal!

Honestly, I'd really like to rewrite the whole settings API so that you register settings with validation and everything else, but that's probably a way off.

Wilma456:
true or false is only autocompleted, if the setting is a boolean. If the type of the setting is a string, nothing will happen.

Lemmmy:
Yes, but the type can still be `false|string` - when false, this would parse as a boolean, but string values would not parse as booleans. If a value is already set to `false`, it will suggest `true`, when that is not really a valid value for some settings.

Wilma456:
@Lemmmy What do you mean? Can you give me examples?

SquidDev:
@Wilma456 Not Lem (obviously), but in [mbs](https://github.com/SquidDev-CC/mbs) there's a couple of settings which are either false or a string.

One example of this is the "history save path": This can either be a file path (`string`), or you can set it to `false` to disable the behaviour. Here you wouldn't want to be passing `true`, as then it'll try to use that as the save path, and then error!

SquidDev:
I'm going to close this for the time being. Auto-completion of `set` is definitely something I want to improve on, but doing it correctly is going to require some further improvements to the `settings` API.

PR #127, opened 2019-02-24
Handle Pastebin spam protection and add a cache buster

Lemmmy:
If you `pastebin put <filename>` when the file contains a URL, you will trigger Pastebin's spam protection:

The user will have to manually navigate to the paste in their browser and complete the captcha before the paste can then be retrieved with `pastebin get`. This PR adds an additional check to `get()` that ensures the content is plaintext, otherwise it will inform the user that they need to complete the captcha in a web browser. It does assume that the reason for failure was a captcha, but it does provide the URL for them to check themselves if it was anything else.

As well as this, it adds a cache buster onto the end of the URL when getting. This is because ComputerCraft caches the redirect to the HTML page (`/raw/paste` → `/paste`) if they tried to retrieve it during spam protection. The cache buster fixes this issue.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
@@ -78,8 +78,6 @@ if sCommand == "put" then
     if response then
         print( "Success." )
 
-        print(response.getResponseHeaders())
-
         local sResponse = response.readAll()
         response.close()
 
--- a/src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
@@ -20,11 +20,20 @@ end
 
 local function get(paste)
     write( "Connecting to pastebin.com... " )
+    -- Add a cache buster so that spam protection is re-checked
+    local cacheBuster = tostring(math.random())
     local response = http.get(
-        "https://pastebin.com/raw/"..textutils.urlEncode( paste )
+        "https://pastebin.com/raw/"..textutils.urlEncode( paste ).."?cb="..cacheBuster
     )
 
     if response then
+        -- If spam protection is activated, we get redirected to /paste with Content-Type: text/html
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
```diff
@@ -20,11 +20,20 @@ end
 
 local function get(paste)
     write( "Connecting to pastebin.com... " )
+    -- Add a cache buster so that spam protection is re-checked
+    local cacheBuster = tostring(math.random())
```
SquidDev: Can we make this `("%x"):format(math.random(0, 2^30))` instead? I realise nobody will ever see this, but just feels a little nicer.

src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
```diff
@@ -20,11 +20,20 @@ end
...
     local response = http.get(
-        "https://pastebin.com/raw/"..textutils.urlEncode( paste )
+        "https://pastebin.com/raw/"..textutils.urlEncode( paste ).."?cb="..cacheBuster
     )
 
     if response then
+        -- If spam protection is activated, we get redirected to /paste with Content-Type: text/html
+        local headers = response.getResponseHeaders()
+        if not headers["Content-Type"] or not headers["Content-Type"]:find( "^text/plain" ) then
+            printError( "\nFailed to get paste. Please complete the captcha in a web browser: https://pastebin.com/"..textutils.urlEncode( paste ) )
```
SquidDev: Just a couple of things:
 - Can you use `io.stderr:write` instead? I'd quite like to move most programs over to use `io` in the future, so might as well make a start now.
 - Thoughts on making this print `Failed.`, and then the reason why on a new line? Could also do that for the `Failed.` below too.

Lemmmy: Yep

SquidDev:
Thank you very much!

PR #129, opened 2019-02-25
Rework computer execution and ComputerThread

SquidDev:
Extracts the non-preemption related code from #119.

### Highlights
 - Timeouts are implemented by a POD class, which means you can set a timeout before the machine has been booted up, and it'll actually apply.
 - Computer execution code (turn on/off, event handling) is now done within `ComputerExecutor`.
 - Task queue is handled by the executor, rather than having a map of computers to queues.

### Still to do
 - Possibly remove the whole task queue, and just have `.work` either run the command or pull an event.
 - Document and test everything. I'd like this to be both more solid, and more understandable when people approach it in the future.
 - Shutdown computers if they hard abort.


SquidDev:
I've implemented an alternative in 67af7a698bff0881076e9d267dc276b94c1ddcc6 and 06e76f9b157f4d08bf33f7a2ba1e59a88cad1070.

PR #151, opened 2019-03-19
Improved code quality in go.lua

UlisseMini:
Stylistic changes to make the code easier to read, I've tested it and it works the same.

Might be better to keep the quotes like so
```lua
local tHandlers = {
    ["fd"]       = turtle.forward,
    ["forward "] = turtle.forward,
    ["forwards"] = turtle.forward,
    ["bk"]       = turtle.back,
    ["back"]     = turtle.back,
    ["up"]       = turtle.up,
    ["dn"]       = turtle.down,
    ["down"]     = turtle.down,
    ["lt"]       = turtle.turnLeft,
    ["left"]     = turtle.turnLeft,
    ["rt"]       = turtle.turnRight,
    ["right"]    = turtle.turnRight,
}
```
Tell me if you like this better and i'll change it.

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/programs/turtle/go.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/turtle/go.lua
@@ -5,18 +5,18 @@ if #tArgs < 1 then
 end
 
 local tHandlers = {
-    ["fd"] = turtle.forward,
-    ["forward"] = turtle.forward,
-    ["forwards"] = turtle.forward,
-    ["bk"] = turtle.back,
-    ["back"] = turtle.back,
-    ["up"] = turtle.up,
-    ["dn"] = turtle.down,
-    ["down"] = turtle.down,
-    ["lt"] = turtle.turnLeft,
-    ["left"] = turtle.turnLeft,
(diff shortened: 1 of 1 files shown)


SquidDev:
Thanks for the pull request! 

Honestly, I'd prefer it if it wasn't aligned (so just keep the `right = turtle.turnRight`, etc...) changes. I realise aligning could be considered more readable, but after a few items it just gets annoying to maintain if one ever adds/removes commands.

UlisseMini:
Not to be rude but, the last time `go.lua` was edited was 2 years ago -- i don't think adding alignment will slow anyone down.

SquidDev:
In this file no, but it's just general code style across the ROM that table entries are not aligned. See, for instance, [`edit`](https://github.com/SquidDev-CC/CC-Tweaked/blob/master/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua#L113) or [`adventure`](https://github.com/SquidDev-CC/CC-Tweaked/blob/master/src/main/resources/assets/computercraft/lua/rom/programs/fun/adventure.lua#L24).

UlisseMini:
Yeah you're right, closing

PR #152, opened 2019-03-20
Add default path for programs and help

JakobDev:
I just added default paths. Programmers can place their programs and help files inside the path and it can be run from everywhere. I know, you can create a file in /startup to set the Path, but I thinks this is a little bit overkill. Default paths are the better way in my opinion.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/startup.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/startup.lua
@@ -1,6 +1,6 @@
 
 -- Setup paths
-local sPath = ".:/rom/programs"
+local sPath = ".:/rom/programs:/programs"
 if term.isColor() then
     sPath = sPath..":/rom/programs/advanced"
 end
@@ -22,7 +22,7 @@ if http then
     sPath = sPath..":/rom/programs/http"
 end
 shell.setPath( sPath )
-help.setPath( "/rom/help" )
+help.setPath( "/rom/help:/help" )
 
 -- Setup aliases
 shell.setAlias( "ls", "list" )


SquidDev:
It'd definitely be nice to have an actual standard for where to place user programs, but it's definitely something I'd want to get a lot of feedback from the community on first.

Might also be better to put the user-controllable path before the `/rom` one, as it allows people to override built-in programs.

lupus590:
@kepler155c startup files/directories don't exist by default either

SquidDev:
Having had another think about this, and chatting with some folks on Discord, I'm actually less sure about this:

The nice thing about the current system is that every program can be self-contained. For instance, `mbs` installs everything into one folder. This means that uninstalling it effectively boils down to removing the `.mbs` folder.

This is somewhat close to what package managers like `nix` do - every package, library, etc.. gets its own install directory and the interesting ones are added to your `PATH`. Given that you can have multiple startup programs now, it'd definitely be possible to use this sort of approach instead.

lupus590:
Would a way to add a program to the path without creating a startup file be better? Perhaps via a setting?

Say I downloaded and ran an installer for Hive (if I ever complete it) and it created `/hive` (and deployed Hive there) and edited the path setting to include `/hive/hive.lua` on the path.

Wilma456:
We can add a setting this too, but I think we need a specific folder. Many Programs are just one file, so they can be placed in this folder. It is good for newbies.

lupus590:
Setting with a default value which makes `/programs` work?

CrazedProgrammer:
It feels kind of arbitrary, considering no existing package manager puts executables there. I like where this is going (FHS but for CC), but I think it needs a bit more thought put into it.

PR #155, opened 2019-03-24
Some input handling improvements

SquidDev:
- Fire `key_up`/`mouse_up` events when closing the GUI.
 - Include active modifiers in the `key` event (`"key", code, repeat, { ctrl = false, alt = true, shift = false }`).

I'm really not sure about the format of the event, whether this is a good idea, etc... so thoughts welcome!

### DIFF
--- a/src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
+++ b/src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
@@ -31,23 +31,23 @@ public class WidgetTerminal extends Widget
 
     private final IComputerContainer m_computer;
 
-    private float m_terminateTimer;
-    private float m_rebootTimer;
-    private float m_shutdownTimer;
+    private float m_terminateTimer = 0.0f;
+    private float m_rebootTimer = 0.0f;
+    private float m_shutdownTimer = 0.0f;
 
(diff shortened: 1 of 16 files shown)


Lemmmy:
If you don't plan on adding any extra modifiers, then it might be better to use three separate booleans for the event, so that we're not throwing tables around (might be better for performance?):

```lua
local event, key, code, repeat, ctrl, alt, shift = os.pullEvent("key")
```

The table isn't too bad, though. Otherwise, this is pretty good.

SquidDev:
Truth be told, I'm not too worried about performance here - we chuck around tables for rednet, and those are often sent far more frequently than `key` events.

SquidDev:
I've added 7a7951ae686d3ca0614b761588fc908e5ef46f9f to master, but need to have another think about the modifiers change. There's lots of weird cases (AltGr, character events) which means you need more than a basic modifier table to implement shortcut keys accurately.

SquidDev:
OK, I'm going to close this for now. I don't think this approach will work as a way to implement modifier keybindings. I'd like to explore some alternative approaches (such as some `key.isDown` method) in the future.

JasonTheKitten:
I just wanted to mention that,  at least in CC 1.74, to detect keybindings with a table, key, and key_up events

PR #164, opened 2019-04-01
rednet.send should return if the message was sent

Luca0208:
According to the Wiki rednet.send should return wether the message was really sent or not: http://www.computercraft.info/wiki/Rednet.send

There's also already the variable "sent" in place to track if the message was sent, however it is unused currently.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/rednet.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/rednet.lua
@@ -81,13 +81,13 @@ function send( nRecipient, message, sProtocol )
         sProtocol = sProtocol,
     }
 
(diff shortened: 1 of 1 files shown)


SquidDev:
Huh, looks like [it was dropped in 1.6pr0](https://github.com/alekso56/ComputercraftLua/commit/f713033bea850754e3dd0fd453c38d9839798210). Should be simple enough to add back.

SquidDev:
I didn't even notice this was a PR. Sorry 🤦‍♂️ .

SquidDev:
I really need to fix that build....

Anyway, thanks for looking into this. Sorry about the kerfuffle - I really should know better than to look at issues when trying to cook.

PR #166, opened 2019-04-03
Update de_de.lang

Wilma456:


src/main/resources/assets/computercraft/lang/de_de.lang
```diff
@@ -1,3 +1,5 @@
+itemGroup.computercraft=CC: Tweaked
```
SquidDev: This is not necessary - the mod name should never be translated, and so should only exist in `en_us.lang`.

Wilma456: I think, it is better to put it in every language files, because some tools for search for missing entry's may find this and add it automatically or throw a warning for a missed entry.

PR #167, opened 2019-04-03
pastebin api

kepler155c:
Allow access to pastebin functionality via an API.


JasonTheKitten:
Feedback wanted? It is a good idea.

src/main/resources/assets/computercraft/lua/rom/apis/pastebin.lua
```diff
@@ -0,0 +1,146 @@
...
+      local code = paste:match( patterns[i] )
+      if code then return code end
+  end
+
+  return nil
+end
+
+function download(url)
+  if not http then
+    return false, "Pastebin requires http API"
```
SquidDev: So I'd be inclined to make this return `nil, <error message>` on failure, and just `<contents>` on success (and same with other functions). This is what most other Lua and CC functions do (see `load`, `http.*`, etc...)

src/main/resources/assets/computercraft/lua/rom/apis/pastebin.lua
```diff
@@ -0,0 +1,146 @@
...
+  end
+
+  if type( url ) ~= "string" then
```
SquidDev: Argument validation should occur before the `http` check.

src/main/resources/assets/computercraft/lua/rom/apis/pastebin.lua
```diff
@@ -0,0 +1,146 @@
...
+      return false, "Invalid pastebin code."
+  end
+
+  -- Add a cache buster so that spam protection is re-checked
+  local cacheBuster = ("%x"):format(math.random(0, 2^30))
+  local response, err = http.get(
+      "https://pastebin.com/raw/"..textutils.urlEncode( paste ).."?cb="..cacheBuster
+  )
+
+  if response then
```
SquidDev: Invert this `if`, and return the error message (so `if not response then return nil, err end`). Just means flow is a little more linear :).

JasonTheKitten:
Does CCT support help files? If so we should probably add one for the api

src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
```diff
@@ -18,59 +18,6 @@ if not http then
...
-    end
-
-    return nil
-end
-
-local function get(url)
-    local paste = extractId( url )
-    if not paste then
-        io.stderr:write( "Invalid pastebin code.\n" )
-        io.write( "The code is the ID at the end of the pastebin.com URL.\n" )
```
SquidDev: I assume this message is no longer displayed? It's a bit of a pain, as that additional information is no longer present.

src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
```diff
@@ -130,29 +55,19 @@ elseif sCommand == "get" then
...
-    if res then
-        local file = fs.open( sPath, "w" )
-        file.write( res )
-        file.close()
+    local success, msg = pastebin.get(sCode, sPath)
 
+    if success then
         print( "Downloaded as "..sFile )
+    else
+        print(msg)
```
SquidDev: Probalby be `printError` or `io.stdderr:write(msg .. "\n")`. Or at least keep the previous "Failed" as an error, and then include the detailed message.

Might be worth making the other messages follow the same format too.

src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
```diff
@@ -18,59 +18,6 @@ if not http then
...
-
-local function get(url)
-    local paste = extractId( url )
-    if not paste then
-        io.stderr:write( "Invalid pastebin code.\n" )
-        io.write( "The code is the ID at the end of the pastebin.com URL.\n" )
-        return
-    end
-
-    write( "Connecting to pastebin.com... " )
```
SquidDev: Likewise, I assume this message is no longer displayed?

src/main/resources/assets/computercraft/lua/rom/apis/pastebin.lua
```diff
@@ -0,0 +1,146 @@
...
+      -- If spam protection is activated, we get redirected to /paste with Content-Type: text/html
+      local headers = response.getResponseHeaders()
+      if not headers["Content-Type"] or not headers["Content-Type"]:find( "^text/plain" ) then
+          return false, "Pastebin blocked the download due to spam protection. Please complete the captcha in a web browser: https://pastebin.com/" .. textutils.urlEncode( paste )
```
SquidDev: This feels ... iffy. It's just a little odd a little weird to return such a detailed message from an API.

src/main/resources/assets/computercraft/lua/rom/apis/pastebin.lua
```diff
@@ -0,0 +1,146 @@
...
+  end
+
+  local file = fs.open( sPath, "w" )
+  file.write( msg )
+  file.close()
+
+  return true
+end
+
+function run(sCode, ...)
```
SquidDev: IMO this doesn't need to be part of the API as it's rather lacklustre for anything outside of basic usage. It doesn't distinguish between network, syntax and runtime errors, uses an ugly file name (the pastebin code, etc...).

src/main/resources/assets/computercraft/lua/rom/apis/pastebin.lua
```diff
@@ -0,0 +1,146 @@
...
+
+  if type( sCode ) ~= "string" then
+    error( "bad argument #1 (expected string, got " .. type( sCode ) .. ")", 2 )
+  end
+
+  local res, msg = download(sCode)
+  if not res then
+    return res, msg
+  end
+  local func, err = load(msg, sCode, "t", _ENV)
```
SquidDev: Probably should be `"=" .. sCode` TBH.

Wilma456:
I think, the API should be made for require() and not for the old os.loadAPI().

Otherwise I think it is a good idea.

kepler155c:
> I think, the API should be made for require() and not for the old os.loadAPI().
> 
> Otherwise I think it is a good idea.

I agree Wilma, but it goes against the conventions of the existing code. I really, really detest os.loadAPI.

SquidDev:
@kepler155c I really wouldn't worry about existing code conventions. `require` didn't exist before CC 1.8, and so no APIs were able to use it. Now that it's available, I'd rather make the most of it - rather that than polluting the global table even more.

kepler155c:
Changes requested have been applied. There needs to be some discussion around the location of the pastebin require. I believe the require path for this file should not be conditionally added as end-users should generally not have to check unknown preconditions to determine if a require can be invoked.

src/main/resources/assets/computercraft/lua/rom/modules/main/http/pastebin.lua
```diff
@@ -0,0 +1,150 @@
...
+function download(url)
+  if type( url ) ~= "string" then
+    error( "bad argument #1 (expected string, got " .. type( url ) .. ")", 2 )
+  end
+
+  if not http then
+    return false, "Pastebin requires http API"
+  end
+
+  local paste = extractId( url )
```
SquidDev: I think it'd probably be better to have this exposed as part of the API, `download`, `get`, etc... only accept the code.

src/main/resources/assets/computercraft/lua/rom/modules/main/http/pastebin.lua
```diff
@@ -0,0 +1,150 @@
...
+  )
+
+  if not response then
+    return response, err
+  end
+
+  -- If spam protection is activated, we get redirected to /paste with Content-Type: text/html
+  local headers = response.getResponseHeaders()
+  if not headers["Content-Type"] or not headers["Content-Type"]:find( "^text/plain" ) then
+    return false, "Pastebin blocked the download due to spam protection. Please complete the captcha in a web browser: https://pastebin.com/" .. textutils.urlEncode( paste )
```
SquidDev: Still would like to come up with a better way of dealing with this. Error messages from API functions shouldn't really be this informative.

src/main/resources/assets/computercraft/lua/rom/modules/main/http/pastebin.lua
```diff
@@ -0,0 +1,150 @@
...
+  end
+
+  local sResponse = response.readAll()
```
SquidDev: Just call it `contents` or something. I'd rather like to avoid too much Hungarian notation.

src/main/resources/assets/computercraft/lua/rom/modules/main/http/pastebin.lua
```diff
@@ -0,0 +1,150 @@
...
+  response.close()
+  return sResponse
+end
+
+local function put(sPath)
```
SquidDev: Maybe have a separate `upload` function which accepts a contents, (optional) file name, and (optional) file type. Then `put` delegates to that in the same way `get` delegates to `download`.

src/main/resources/assets/computercraft/lua/rom/modules/main/http/pastebin.lua
```diff
@@ -0,0 +1,150 @@
...
+  local res, msg = download(sCode)
+  if not res then
+    return res, msg
+  end
+
+  local file = fs.open( sPath, "w" )
+  file.write( res )
+  file.close()
+
+  return sPath
```
SquidDev: Ehhh, I don't think this needs to return anything beyond `true`.

src/main/resources/assets/computercraft/lua/rom/modules/main/http/pastebin.lua
```diff
@@ -0,0 +1,150 @@
...
+        return func, err
+    end
+    return pcall(func, ...)
+end
+
+return {
+  get = get,
+  put = put,
+  run = run,
+}
```
SquidDev: Line endings! Also, is this table indented with two spaces?

src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
```diff
@@ -121,40 +50,36 @@ elseif sCommand == "get" then
         return
     end
 
+    print( "Connecting to pastebin.com... " )
```
SquidDev: Well, this is categorically a lie. This should at least be after the "file already exists" check.

JasonTheKitten:
The last comment of which I had already mentioned. https://github.com/SquidDev-CC/CC-Tweaked/pull/167#issuecomment-479584235

src/main/resources/assets/computercraft/lua/rom/modules/main/http/pastebin.lua
```diff
@@ -1,150 +1,128 @@
...
-
-  for i = 1, #patterns do
-    local code = paste:match( patterns[i] )
-    if code then return code end
-  end
-
-  return nil
+--- Parse the pastebin code from the given code or URL
+local function parseCode(paste)
+	local patterns = {
```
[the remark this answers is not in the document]
kepler155c: 4 spaces on a tiny monitor... (but updated now)

src/main/resources/assets/computercraft/lua/rom/modules/main/http/pastebin.lua
```diff
@@ -0,0 +1,128 @@
+--- Parse the pastebin code from the given code or URL
+local function parseCode(paste)
+    local patterns = {
```
SquidDev: Now that I'm looking at this, might be worth hoisting this out of the function definition. Sorry.

src/main/resources/assets/computercraft/lua/rom/modules/main/http/pastebin.lua
```diff
@@ -0,0 +1,128 @@
...
+        "^https?://pastebin.com/([%a%d]+)$",
+        "^pastebin.com/([%a%d]+)$",
+        "^https?://pastebin.com/raw/([%a%d]+)$",
+        "^pastebin.com/raw/([%a%d]+)$",
+    }
+
+    for i = 1, #patterns do
+        local code = paste:match(patterns[i])
+        if code then
+        return code
```
SquidDev: Indentation is a little odd here.

src/main/resources/assets/computercraft/lua/rom/modules/main/http/pastebin.lua
```diff
@@ -0,0 +1,128 @@
...
+        return false, "Pastebin requires http API"
+    end
+
+    -- POST the contents to pastebin
+    local key = "0ec2eb25b6166c0c27a394ae118ad829"
+    local response = http.post(
+        "https://pastebin.com/api/api_post.php",
+        "api_option=paste&" ..
+        "api_dev_key=" .. key .. "&" ..
+        "api_paste_format=lua&" ..
```
SquidDev: I'd make `api_paste_name` optional (append it to the url only if present), and also add an optional format argument (for `api_paste_format`).

src/main/resources/assets/computercraft/lua/rom/modules/main/http/pastebin.lua
```diff
@@ -0,0 +1,128 @@
...
+        "api_paste_name=" .. textutils.urlEncode(name) .. "&" ..
+        "api_paste_code=" .. textutils.urlEncode(text)
+    )
+
+    if not response then
+        return false, "Failed."
```
SquidDev: We should be able to do `local response, err = http.post` and then return the error message here.

src/main/resources/assets/computercraft/lua/rom/modules/main/http/pastebin.lua
```diff
@@ -0,0 +1,128 @@
...
+    end
+
+    -- Determine file to upload
+    if not fs.exists(path) or fs.isDir(path) then
+        return false, "No such file"
+    end
+
+    -- Read in the file
+    local name = fs.getName(path)
+    local file = fs.open(path, "r")
```
SquidDev: IMO, it'd be nicer to do `local file, err = fs.open(path, "r"); if not file then return nil, err end` instead of having an explicit "exists and isDir" check.

src/main/resources/assets/computercraft/lua/rom/modules/main/http/pastebin.lua
```diff
@@ -0,0 +1,128 @@
...
+    if type(path) ~= "string" then
+        error("bad argument #2 (expected string, got " .. type(path) .. ")", 2)
+    end
+
+    local res, msg = download(code)
+    if not res then
+        return res, msg
+    end
+
+    local file = fs.open(path, "w")
```
SquidDev: Should possibly handle the case where we can't open the file here. So just `local file, err = fs.open(path, "w"); if not file then return nil, err end`. For instance, when writing to `rom` or a directory.

src/main/resources/assets/computercraft/lua/rom/modules/main/http/pastebin.lua
```diff
@@ -0,0 +1,128 @@
...
+    )
+
+    if not response then
+        return response, err
+    end
+
+    -- If spam protection is activated, we get redirected to /paste with Content-Type: text/html
+    local headers = response.getResponseHeaders()
+    if not headers["Content-Type"] or not headers["Content-Type"]:find("^text/plain") then
+        return false, "Pastebin blocked due to spam protection"
```
SquidDev: Pinging @Lemmmy here to get their thoughts on this. I don't really want to return a massively long error message, but at the same time we need to return informative info to the user within the `pastebin` program.

Lemmmy: Return a short error message, like `Got HTML response`, then in the `pastebin` program, change that to the longer, more friendly message `Pastebin blocked due to spam protection`.

PR #183, opened 2019-04-22
Thread safe inventory interactions

SquidDev:
This is an attempt to inventory actions from peripherals (and other APIs) thread-safe.

 - [x] Disk drives
 - [x] Printers
 - [x] Turtles

This largely involves shifting some things to execute on the main thread, though I'm also taking the opportunity to clean up some of the `TileEntity` code.


SquidDev:
Well, this uncovered a whole 'nother set of thread-safety issues, so yay. However, the actual inventory interaction code should be good now,

SquidDev:
On the bright side, there's lots of nice invariants which the Checker framework should be able to enforce (see #239).

PR #203, opened 2019-05-08
Add Package Manager

JakobDev:
I think, CCT should have a Package Manger preinstalled like OC has. It is very easier to install programs by their name as searching for the pastebin Code in the Internet.

This is the first Beta of CCTPM (CC-Tweaked Package Manager). It is in a early beta and some pieces of the code are looking bad but it works. I publish this beta the see what other people are thinking about this.

[Here is a example how the repos will looks like](https://gitlab.com/Wilma456/cctpm/tree/master)

Just try it yourself and say what you are thinking.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -982,6 +982,8 @@ settings.set( "edit.default_extension", "lua" )
 settings.set( "paint.default_extension", "nfp" )
 settings.set( "lua.autocomplete", true )
 settings.set( "list.show_hidden", false )
+settings.set( "cctpm.directory", "/var/cctpm" )
+settings.set( "cctpm.repolist", "repolist|https://gitlab.com/Wilma456/cctpm/raw/master/repolist.txt" )
 if term.isColour() then
     settings.set( "bios.use_multishell", true )
 end
--- a/src/main/resources/assets/computercraft/lua/rom/modules/cctpm.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/modules/cctpm.lua
@@ -1,0 +1,248 @@
+local cctpm = {}
+
+cctpm_cache = {}
+
+local function downloadFile( sURL, sPath )
+    local response, err = http.get( sURL )
+    if err then
+        return false, err
+    end
+    local handle = fs.open( sPath, "w" )
+    handle.write( response.readAll() )
+    response.close()
+    handle.close()
+    return true
+end
+
+local function addRepo( sRepo, sName, tInstalledList )
+    local tRepo = textutils.unserialise( sRepo )
+    if type( tRepo ) ~= "table" then
+        return false, "Can't Read Repo \""..sName..'"'
+    end
+    for a, b in pairs( tRepo) do 
+        cctpm_cache.packages[a] = b
+        if tInstalledList[a] then
+            cctpm_cache.packages[a]["installed"] = true
+        else
+            cctpm_cache.packages[a]["installed"] = false
+        end 
+    end
+    return true
+end
+
+function cctpm.readRepos()
+    cctpm_cache.packages = {}
+    local sDirectory = fs.combine( settings.get( "cctpm.directory", "/var/cctpm" ), "" )
+    local tInstalledList = {}
+    if fs.exists( sDirectory.."/installed.txt" ) then
+        for sLine in io.lines( sDirectory.."/installed.txt" ) do
+            local sPackageName = sLine:match("([^,]+);([^,]+)")
+            tInstalledList[sPackageName] = true
+        end
+    end
+    local sRepoList = settings.get( "cctpm.repolist", "" )
+        for sRepo in sRepoList:gmatch( sRepoList, "[^;]+") do
+            local sType, sMeta = sRepo:match("([^,]+)|([^,]+)")
+            if sType == "repo" then
+                sName, sURL = sMeta:match("([^,]+):([^,]+)")
+                local response, err = http.get( sURL )
+                if err then
+                    return false, err
+                end
+                local ok, err = addRepo( response.readAll() , sName, tInstalledList )
+                response.close()
+                if not ok then
+                    return false, err
+                end
+            elseif sType == "repolist" then
+                local ok, err = downloadFile( sMeta, "/tmp/repolist.txt" )
+                if not ok then
+                    return false, err
+                end
+                for sLine in io.lines( "/tmp/repolist.txt" ) do
+                    sName, sURL = sLine:match("([^,]+)|([^,]+)")
+                    local response, err = http.get( sURL )
+                    if err then
+                        return false, err
+                    end
+                    local ok, err = addRepo( response.readAll() , sName, tInstalledList )
+                    response.close()
+                    if not ok then
+                        return false, err
+                    end
+                end
+            end
+        end
+    fs.mkdir( sDirectory.."/repos" )
+    for k, v in ipairs( fs.list( sDirectory.."/repos" ) ) do
+        local sName = v:sub( 1, -5 )
+        local handle = fs.open( sDirectory.."/repos/"..v, "r" )
+        local sContent = handle.readAll()
+        handle.close()
+        local ok, err = addRepo( sContent, sName, tInstalledList )
+        if not ok then
+            return false, err
+        end
+    end
+    return true
+end
+
+local function findDependencies( sPackage, sParent )
+    if cctpm_cache.err then
+        return
+    end
+    if not cctpm_cache.packages[sPackage] then
+        cctpm_cache.err = '"'..sParent..'" depends to the the package "'..sPackage..'" which doesen\'t exists'
+        return
+    end
+    for k,v in ipairs(cctpm_cache.packages[sPackage]["dependencies"]) do
+        if not cctpm_cache.dependencies[v] then
+            cctpm_cache.dependencies[v] = true
+            findDependencies( v, sPackage )
+        end
+    end
+end
+
+function cctpm.getDependencies( sPackage )
+    if type( sPackage ) ~= "string" then
+        error( "bad argument #1 (expected string, got " .. type( sPackage ) .. ")", 2 )
+    end
+    if cctpm_cache.packages[sPackage] == nil then
+        return nil, "Can't find Package \""..sPackage..'"'
+    end
+    cctpm_cache.dependencies = {}
+    cctpm_cache.err = nil
+    findDependencies( sPackage )
+    if cctpm_cache.err then
+        return nil, cctpm_cache.err
+    else
+        local tDependencies = {}
+        for k, v in pairs( cctpm_cache.dependencies ) do
+            table.insert( tDependencies, k )
+        end
+        cctpm_cache.dependencies = nil
+        return tDependencies
+    end
+end
+
+function cctpm.getInstalledPackages()
+    local sDirectory = fs.combine( settings.get( "cctpm.directory", "/var/cctpm" ), "" )
+    if not fs.exists( sDirectory.."/installed.txt" ) then
+        return {}
+    end
+    local tInstalled = {}
+    for sPackage in io.lines( sDirectory.."/installed.txt" ) do
+        local sName = sPackage:match("([^,]+);([^,]+)")
+        table.insert( tInstalled, sName )
+    end
+    return tInstalled
+end
+
+function cctpm.install( sPackage, bForce )
+    if type( sPackage ) ~= "string" then
+        error( "bad argument #1 (expected string, got " .. type( sPackage ) .. ")", 2 )
+    end
+    if bForce ~= nil and type( bForce ) ~= "boolean" then
+        error( "bad argument #2 (expected boolean, got " .. type( bForce ) .. ")", 2 ) 
+    end
+    if cctpm_cache.packages[sPackage] == nil then
+        return false, "Can't find Package \""..sPackage..'"'
+    end
+    local tInstall, err = cctpm.getDependencies( sPackage )
+    if not tInstall then
+        return false, err
+    end
+    table.insert( tInstall, sPackage )
+    local sDirectory = fs.combine( settings.get( "cctpm.directory", "/var/cctpm" ), "" )
+    for a, sInstallPack in ipairs( tInstall ) do
+        if ( not cctpm_cache.packages[sInstallPack]["installed"] ) or ( bForce and sInstallPack == sPackage ) then
+            if not cctpm_cache.packages[sInstallPack]["files"] then
+                return false, '"'..sInstallPack..'" has no Filelist'
+            end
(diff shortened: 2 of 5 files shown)


zardyh:
Your example link is a 404.

Wilma456:
@zardyh Sorry, it had a l too much. I have it fixed now.

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -982,6 +982,8 @@ settings.set( "edit.default_extension", "lua" )
 settings.set( "paint.default_extension", "nfp" )
 settings.set( "lua.autocomplete", true )
 settings.set( "list.show_hidden", false )
+settings.set( "cctpm.directory", "/var/cctpm" )
+settings.set( "cctpm.repolist", "repolist|https://gitlab.com/Wilma456/cctpm/raw/master/repolist.txt" )
```
zardyh: Why does `repolist` occur in both the setting name and the value..?

src/main/resources/assets/computercraft/lua/rom/modules/cctpm.lua
```diff
@@ -0,0 +1,248 @@
+local cctpm = {}
+
+cctpm_cache = {}
```
zardyh: Why is this a global?

src/main/resources/assets/computercraft/lua/rom/modules/cctpm.lua
```diff
@@ -0,0 +1,248 @@
...
+    handle.write( response.readAll() )
+    response.close()
+    handle.close()
+    return true
+end
+
+local function addRepo( sRepo, sName, tInstalledList )
+    local tRepo = textutils.unserialise( sRepo )
+    if type( tRepo ) ~= "table" then
+        return false, "Can't Read Repo \""..sName..'"'
```
zardyh: Why is this sentence in title case?

src/main/resources/assets/computercraft/lua/rom/modules/cctpm.lua
```diff
@@ -0,0 +1,248 @@
...
+                if err then
+                    return false, err
+                end
+                local ok, err = addRepo( response.readAll() , sName, tInstalledList )
+                response.close()
+                if not ok then
+                    return false, err
+                end
+            elseif sType == "repolist" then
+                local ok, err = downloadFile( sMeta, "/tmp/repolist.txt" )
```
zardyh: Why are you not being consistent with your use of Dan's hungarian notation? (`sType`, `err`)

src/main/resources/assets/computercraft/lua/rom/modules/cctpm.lua
```diff
@@ -0,0 +1,248 @@
...
+    end
+    return true
+end
+
+local function findDependencies( sPackage, sParent )
+    if cctpm_cache.err then
+        return
+    end
+    if not cctpm_cache.packages[sPackage] then
+        cctpm_cache.err = '"'..sParent..'" depends to the the package "'..sPackage..'" which doesen\'t exists'
```
zardyh: You have no spaces around the `..` operator here...

src/main/resources/assets/computercraft/lua/rom/modules/cctpm.lua
```diff
@@ -0,0 +1,248 @@
...
+        if not cctpm_cache.dependencies[v] then
+            cctpm_cache.dependencies[v] = true
+            findDependencies( v, sPackage )
+        end
+    end
+end
+
+function cctpm.getDependencies( sPackage )
+    if type( sPackage ) ~= "string" then
+        error( "bad argument #1 (expected string, got " .. type( sPackage ) .. ")", 2 )
```
zardyh: .. but you have them here.

src/main/resources/assets/computercraft/lua/rom/modules/cctpm.lua
```diff
@@ -0,0 +1,248 @@
...
+            end
+            for k,v in pairs(cctpm_cache.packages[sInstallPack]["files"]) do
+                local ok, err = downloadFile( k, v )
+                if not ok then
+                    return false, err
+                end
+            end
+            if not cctpm_cache.packages[sInstallPack]["installed"] then
+                local file = fs.open( sDirectory.."/installed.txt", "a" )
+                file.writeLine( sInstallPack..";"..cctpm_cache.packages[sInstallPack]["version"] )
```
zardyh: No spaces here

src/main/resources/assets/computercraft/lua/rom/modules/cctpm.lua
```diff
@@ -0,0 +1,248 @@
...
+            end
+            cctpm_cache.packages[sInstallPack]["installed"] = true
+        end
+    end
+    return true
+end
+
+function cctpm.remove( sPackage, bForce )
+    if type( sPackage ) ~= "string" then
+        error( "bad argument #1 (expected string, got " .. type( sPackage ) .. ")", 2 )
```
zardyh: Spaces here

src/main/resources/assets/computercraft/lua/rom/programs/http/cctpm.lua
```diff
@@ -0,0 +1,85 @@
...
+    if not ok then
+        printError( err )
+    end
+elseif tArgs[1] == "upgrade" then
+    local tList = cctpm.getUpgradeable()
+    if #tList == 0 then
+        print( "All Packages are up to date!" )
+        return
+    end
+    print( "The following Packages can be upgradet:" )
```
zardyh: Upgraded.

src/main/resources/assets/computercraft/lua/bios.lua
Wilma456: The repolist inside the value is for the type. There are 2 different types: repolist, which are a list of repos (on this way we can add repos without publishing a new Version) and repo, which are the direct link to a repo.

Wilma456:
I had made some Changes to CCTPM.

Wilma456:
btw: We need a extra Repo which contains the list of Repositories. Maybe @SquidDev  can create one?

Wilma456:
@zardyh If you have any suggestions just let me know. This is just a beta. I can add and remove features.

SquidDev:
Right, many apologies for taking a while to get around to looking at this. I wanted some time to put my thoughts in order, and then totally failed to get round to commenting.

I think before any more work is done, we need to have a long think about how this will work. It'd be fantastic to have a package manager but, as always, it's something I want to get right the first time round. 

It's probably worth having a look at some of the existing Lua package managers (both for CC and not), seeing what they get right/wrong, and getting some feedback from the community. It'd be nice to have a rough sketch of what we want everything to look like before continuing with implementation.

Personally, I'd love to get something like [LuaRocks](https://luarocks.org/) working within CC - it's an existing system and so is well documented, and has a wealth of existing packages which (largely) work on CC. However, it may be a little over-the-top for our actual needs.

A couple of things which spring out at me:

 - **Conflicting files:** There's a bit of me which wants to go the nix route, and install every package into its own distinct folder, but that'd then means there needs to be some integration between `require` and the package manager.

 - **Package versioning:** Do we want it, and if so how do we handle it? The current implementation defines a version field, but it doesn't really mean anything - you can't say "I need at least version X", nor can you get access to older versions - it's always downloading master after all.

 - **Metadata:** Currently the format stores the name and description. Probably would be worth also providing fields for a license, and links to source code (+issue tracker?). It may also be worth storing the metadata for installed files locally - otherwise you run the the remote metadata not reflecting what's installed on your computer.

 - **Building:** Some projects may require some pre-processing before being installed (say, using Howl). Do we want to/need to handle this, or should it be the responsibility of the author to provide pre-built files.

I don't know - I'm incapable of doing anything without overcomplicating it. I'll try to send this round and attempt to get people to provide thoughts.

hugeblank:
I think that a topic of discussion should also be if we want to integrate with a specific service, like github itself. 
This would: 
- Remove the need for a weird repository list. We could just use usernames/organizations instead.
- Reduce complexity a bit, seeing as files aren't scattered all over the place, but rather sorted, and can easily be tracked down by the github API. 
- Partially answer the package versioning question, allowing players to simply define the version as a tag in their repositories. 

Here's my stab at giving a pseudo-explanation of my idea:
`manager install LDDestroier/enchat /programs/chat` would install LDD's enchat to `/programs/chat`, much like the typical repository grabbing programs do, but would also keep the commit hash tucked away somewhere. When an update is called, the manager checks the repository's latest commit hash, comparing it to the version it currently has. If the hash has changed, update.

If I had the will to I guess what I envision for CC as a package manager is more of an interface with github than something standalone. This is completely counter to what this PR is shooting for, but I believe that integration with an existing service will reduce the complexity, and increase usability in the end.

SquidDev:
@hugeblank I think one thing I really don't want to do is tie everything to GitHub (or any other hosting provider). Obviously their API makes it very easy to work with, but I also think it's unfair on people who may want to use other sites (such as self-hosted git, GitLab, etc...).

lupus590:
This will probably want to tie into the path extending issue too https://github.com/SquidDev-CC/CC-Tweaked/pull/152

dmarcuse:
I'm of the opinion that CC should focus on providing the necessary facilities to make a package manager feasible (ideally supporting an environment similar enough to PUC Lua that luarocks could be used) rather than including it's own package manager. By including the package manager in CC itself, it's no longer possible to update it independently of updating the mod - which may not be feasible (e.g. for servers with many players). I also feel that forcing users to use a specific package manager isn't a good idea (which is also a reason that I wish there was a more straightforward way of replacing CraftOS).

SquidDev:
> ideally supporting an environment similar enough to PUC Lua that luarocks could be used

This is never going to be 100% possible, as LuaRocks shells out to `git` and `tar` (and uses `luasocket` + `lfs`) - you'd need to at least stub those. But yes, it'd be nice to have a standard compliant `os` library (fairly sure that's the only thing lacking now).

dmarcuse:
I was also thinking something along the lines of a setting for package.path, since it's currently reset to a hardcoded value with each program run.

Wilma456:
I think making a LuaRoocks Package Manager is a bad idea. There are many Packages on the LuaRock Website which doesn't work with CC. And most CC Programs dosen't work in a normal Lua environment. I think we should separated this.

The Package Format is inspired by OPPM from OpenComputers. The Repo can be loaded with textutils.unserialise(), so we don't need a extra parser.

I think we shouldn't create a extra folder for each program. Without this packages are able to contain plugins or other extra stuff. Conflicting files can be theoretically a problem, but real life packages like DEB have this problem too and it work without problems.

The only reason the Version exits, is that the package manager know what packages are need to update and he don't need to update all packages. The most package managers do it this way and doesn't have the possibility to install older Versions.

I think, this like the license or the URL to the Repo can be put in the description.

I think, a package manger doen't need to build programs e.g. with Howl. The most programs are just written in Lua. I think it's better if you gave the package manger the build. But we can add a source argument like apt has, which downloads the source.

lupus590:
What about compatibility with older versions of cc? Eventually there will be a program which no longer works on an old version of cc which someone will be using and they might update without realising the issue and now can't downgrade to the prior working version.

Wilma456:
Yo mean, I should add a "min CC Version" to the packages?

lupus590:
Perhaps, but then what if a third party want to make a compatibility patch for it? Could users force install and suffer the consequences? The compatibility patch could then list it as a dependency and do a force install of a version that it can patch.

Wojbie:
I will admit that adding Proper package manager would be nice thing to add to CC.

But it would also open can of worms of having official package source that needs to be hosted/managed/upkeept. Right now CC:T is a one man project (Squids) and i really don't see this community having dedicated members/moderators and admins that would be needed to upkeep a proper safe package manager.. And if we had static one then what's the point?

Having Package manager API itself that works properly with `require` ect would be lovely and great addition to the mob but i would suggest leaving it disabled as default (without any 'main' package repository) and then just let community make/upkeep one? Then it would be a player choice to connect to one (or make one) without adding extra upkeep (other then upkeeping package manager itself in source code) onto this project.

As for what would be needed in package manager itself? Well i think Squid mentioned most needed parts that would be needed.
Plus with offloading the repository onto community/server hosts would make idea of compatibility void cause repositories would be hosted per specific version of mod.

JakobDev:
@lupus590 I think that would be to overkill and I it can only emulate the LUA Part, not the Java Part.

@Wojbie I had planned that every User gets his won Repo which he can Manage like Packman or OPPM does. I think it should be enabled by default, so every User can just run cctpm install <program> direct after placing the Computer.

Wojbie:
@JakobDev and what would you have cctpm install do by default? Where would it download packages from?

JakobDev:
@Wojbie cctpm supportes a repo list. [Here is a example](https://gitlab.com/JakobDev/cctpm/blob/master/repolist.txt). @SquidDev can maybe create a repo which contains only a repolist and every user is able to add his own Repo with a PR.

Wojbie:
That sounds like security nightmare to be honest.

JakobDev:
@Wojbie It's the same security nightmare like pastebin, wget or any other download from the internet.

dmarcuse:
No, this is undoubtedly worse. Here's two quick examples:

- You're using `textutils.unserialize` on arbitrary code that is outside the control of the user. While this is sandboxed, it is still vulnerable to issues like `while true do end`.
  - As a followup, `pastebin run` makes it explicit that external code is being run, while this package manager does not. The other usages of `pastebin` and `wget` don't run any external code.
- Package name collisions aren't detected or handled. If repo A specifies package "mypackage", repo B can also specify "mypackage" and replace it with a malicious version, and this implementation of the package manager won't warn the user or attempt to resolve this at all.

SquidDev:
OK, had a long think about this and talked to some people on discord, and going to close this. I'm not opposed to bundling a package manager with CC, but the wider opinion is a lot more ... mixed, and it's really not a battle I want to fight right now.

I think if people want to continue work on this, there's some things which are worth considering:

 - Should the package manager live in ROM? I'd argue yes, but there's an argument for only providing an install command and nothing else.

 - There's generally a bit of a difference between system package managers (`apt`, `pacman`) and language ones (`npm`, `luarocks`). System package managers generally only allow you to install the latest version, are somewhat more lax on versioning, etc... Language package managers are _more_ oriented towards development, and so allow installing multiple versions of a library, version bounds, etc...

   Both are obviously useful in different situations, and it's worth working out what benefits CC users more.

 - Package distribution is messy. I for one don't want to tie anything to GitHub, but want to create a centralised package repository even less.

 - Relatedly, namespaces are probably a good idea - so people can't mask other people's packages.

 - Should transitive dependencies also be added to the `bin` and `require` path, or require an explicit `ccpm install` command for this to occur?

I'm not expecting any of these to be answered here (hell, I don't think any of these have an absolute answer), it's just some things which people have brought up and are worth addressing should this be revisited.

PR #207, opened 2019-05-14
Add a function for type-checking arguments

zardyh:
The basic interface is thus:
```lua
  expect (index, actual_value, types ...)
```

If the argument is optional, just add 'nil' to the possible types (it
won't be printed).

If `type(actual_value)` isn't one of `types`, an error like this:
```
  bad argument #index (expected types₁ or types₂..., got type(actual_value))
```
is thrown.


SquidDev:
Well, that was a wild time.

SquidDev:
OK, I'm largely in favour of this. I'm still not a fan of polluting the global table with functions like this, but given the nature of `os.loadAPI`, I don't think there's really another way.

zardyh:
Perhaps we could put it in the os API as a small counter to namespace pollution?

dmarcuse:
How does this impact performance, specifically around the `window` API? The individually small impact of handling varargs in `expect` calls could add up in a tight rendering loop.

Wilma456:
@dmarcuse The window API should be no problem. It already had checks. But the performance could be a problem.

zardyh:
There's an easy way to fix the performance problem: remove the checks.

Wojbie:
That would be something to consider.. window api is called multiple times each time something is written on screen. Having it do and re-do the checks is somewhat redundant in some cases.

dmarcuse:
@Wilma456 The window API did have checks, but this PR makes those checks more complicated and potentially more performance intensive, and these small changes can quickly add up.

zardyh:
The overhead from `expect` is minimal in the general case, especially as the arguments are generally in the order of likelihood so the number of iterations is small, since it's basically just 3 calls to `local`ised Java functions.

SquidDev:
OK, just pushed a small commit which removes the type checks from `expect`. It's meant to be an internal function (for the time being) so doesn't need the same level of guards. The second is definitely uglier, but 

With this, I've done a little bit of benchmarking, and `expect` ends up taking about 2.8 times as long than an inlined `type` check (959μs vs 387μs).

In most complex functions, I'd say this is an acceptable cost, as the function's body far outweighs the `expect` overhead. However, for very small functions (such as `window.setCursorPos`), obviously this will introduce some non-trivial overhead.

One improvement we can make would be to introduce a non-variadic function (say, `expect1`) - this takes only 1.4x more than an inline check. However, having two functions which do the same thing is a little ugly - I'd appreciate people's thoughts on whether it's worth it or not!

One other thing to discuss is how to make it clear that this is not intended to be used externally. I'm kinda tempted to wrap the function in earmuffs, and so require it to be accessed as `_G["*expect*]` <sub>(or even, `_G["*expect* (this is internal, please don't use"]`)</sub>.

dmarcuse:
On the topic of performance, I think that the cases where performance matters (i.e. `window`) are infrequent enough that just using inline checks would be better than introducing non-variadic but otherwise identical functions.

SquidDev:
OK, just pushed a couple of changes to this - it's in a state I'm happy with, but would like some thoughts from others first:

 - `expect` is now stored as `_G["~expect"]` in the global table. This makes it invisible to auto-complete, though obviously will still show up if you dump the whole table.

   Please _do not_ use this function in user code, as I cannot guarantee it will remain there in the future.

 - I've somewhat inlined the type checks within the `window` API, so performance should remain the same.

 - We will attempt to gather the function name when producing an error message, in a manner similar to PUC Lua. This means `os.sleep("hello")` will error with `bad argument #1 to 'sleep' (expected number, got string)`. IMO this makes things that wee bit nicer, though comments are welcome!

zardyh:
I'm ok with this.

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -1,5 +1,48 @@
...
+        error( ("bad argument #%d to '%s' (expected %s, got %s)"):format(index, name, type_names, t), 3 )
+    else
+        error( ("bad argument #%d (expected %s, got %s)"):format(index, type_names, t), 3 )
+    end
+end
+
+-- We expose expect in the global table as APIs need to access it, but surround
+-- it with earmuffs. expect is an internal function, and should not be used by
+-- users.
+_G["~expect"] = expect
```
lupus590: I wouldn't describe one ~ as earmuffs. I understand the point but I think the description needs improving.

Perhaps unspeakable name? This is a term borrowed from C# so it kinda works.

SquidDev: Ahhhh, it was originally earmuffs and then I changed it without updating the comment. Good catch!

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -78,7 +114,7 @@ if _VERSION == "Lua 5.1" then
     else
         loadstring = function(string, chunkname) return nativeloadstring(string, prefix( chunkname ))
 
-        -- Inject a stub for the old bit library
+            -- Inject a stub for the old bit library
```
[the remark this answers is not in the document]
SquidDev: _Technically_ the indentation is correct, it's just that the comment is in the wrong place - looks like it managed to end up within `loadstring` by mistake.

dmarcuse:
:tada:

PR #210, opened 2019-05-15
Make delete delete many files

zardyh:
Actually, many *globs*. It additionally prints the glob if no files
matched it, since that's clearer.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/delete.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/delete.lua
@@ -1,16 +1,17 @@
+local args = table.pack(...)
 
-local tArgs = { ... }
-if #tArgs < 1 then
-    print( "Usage: rm <path>" )
+if args.n < 1 then
+    print("Usage: rm <paths>")
     return
 end
 
-local sPath = shell.resolve( tArgs[1] )
-local tFiles = fs.find( sPath )
-if #tFiles > 0 then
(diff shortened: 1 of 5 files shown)


src/main/resources/assets/computercraft/lua/rom/programs/delete.lua
```diff
@@ -1,16 +1,17 @@
...
-    for n,sFile in ipairs( tFiles ) do
-        fs.delete( sFile )
+for i = 1, args.n do
+    local files = fs.find(shell.resolve(args[i]))
+    if #files > 0 then
+        for n, file in ipairs(files) do
+            fs.delete(file)
+        end
+    else
+        printError(args[i] .. ": No matching files")
```
SquidDev: Maybe `No matching files for '%s'`? Not sure, but probably should wrap it in quotes either way.

zardyh: well, your call

zardyh: can I get a formatting printError though?

zardyh:
@SquidDev Should we stop at the first not-matching file?

kepler155c:
doesn't seem to support ```rm a.lua b.lua```

zardyh:
can't reproduce

SquidDev:
Thank you!

PR #213, opened 2019-05-17
Add "Goto Line" to edit

JakobDev:
I just added a new Menuitem which allows jumping to a specific line. This is useful for large files where the user has to scroll long to reach a line.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
@@ -60,6 +60,7 @@ if peripheral.find( "printer" ) then
     table.insert( tMenuItems, "Print" )
 end
 table.insert( tMenuItems, "Exit" )
+table.insert( tMenuItems, "Goto" )
 
 local sStatus = "Press Ctrl to access menu"
 if string.len( sStatus ) > w - 5 then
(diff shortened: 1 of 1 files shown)


SquidDev:
So this was something which was discussed a while back, and Dan mentioned that he wasn't a fan of adding new items to the menu (https://github.com/dan200/ComputerCraft/pull/317#discussion_r122574059).

I wonder if a nicer solution would be to not show it on the menu, but still allow using the <kbd>C g</kbd> shortcut. It would be even nicer if we could get it so <kbd>C-g</kbd> (holding control) works, but that might require a bit more of a rework to edit.

Lemmmy:
val's bedit program just takes an extra arg

JakobDev:
@SquidDev Dan is not the maintainer of this Repo. If we add just a keyboard shortcut, many people will not discover this function.

@Lemmmy I think it's better when the user can jump any time while editing and not only at the start.

SquidDev:
@JakobDev While true, I do (generally) consider Dan WoG, especially when it comes to being beginner-friendly. Dan's definitely got more expertise in that area than me.

Given that space is pretty constrained already on the menu (especially on smaller screens like turtles and pocket computers), I think it's best if the functionality is there first - we can work on discoverability later.

JakobDev:
@SquidDev I don't think, that a Goto Menupoint is not userfriendly. It's very basic and easy to understand.

nift4:
What about a MOTD for the Keyboard shortcut?

Issue #208, opened 2019-05-14
Holding Computer In One Hand Causes Render Funk


EntranceJew:
## Useful information to include:
 - Minecraft 1.14.1
 - CC: Tweaked version 1.84.4
 - Detailed reproduction steps: put a computer in one hand, and another item in a different hand; alternatively, hold a computer in both hands

![https://i.imgur.com/92RsbGS.png](https://i.imgur.com/92RsbGS.png)

at some point handheld computers get real messed up, the one on the left has got some extreme margins

the one on the right is similar around the edges but not quite so bad

PR #215, opened 2019-05-25
Rework rendering of in-hand pocket computers

SquidDev:
We've had in-hand pocket computer rendering for a while, but it's always been a little funky:
 - The terminal did not fit well within the item's texture, so we had a rather large border (see dan200/ComputerCraft#497 for example).
 - The "correctness" of this was very tied to Minecraft's item rendering code. This changed a little in 1.13, causing problems like #208.

We're now rendering something very similar to the pocket computer GUI, though also handling coloured pocket computers, and adding in the modem light.

I'd appreciate some thoughts on this before merging. IMO it looks better, but comments are welcome!

## Examples
Pocket computer in one hand

Pocket computers in both hands

Yes, there's still a small number of artefacts - I believe this is an issue with the underlying texture, which I'll probably resolve in a separate change.


lupus590:
Looking into the texture, it could be that the texture coords could be off not the texture.

lupus590:
looks like that was it, I erased the overlapping bit of the texture and it looked better. I'll try to find the code for you

PR #218, opened 2019-05-27
Add webrun

JakobDev:
Webrun does the same as pastebin run but for any URL and not only for pastebin.


SquidDev:
I'm a little iffy about this, as per the discussion on dan200/ComputerCraft#215.

If this does end up being added, I do wonder if it'd be nicer to bundle this as part of `wget`, rather than having two near-identical programs.

JakobDev:
What argument for wget do you suggest?

zardyh:
"Insanity is doing the same thing over & over again & expecting different results." - literally albert einstein

Not even joking you already tried this once

lupus590:
-r or --run makes sense to me

JakobDev:
@zardyh I had made the old PR for the original CC. This Repo is maintained by SquidDev.

@lupus590 -r is for recursive in the original wget, but --run would be an option. The only thing is, that the original wget is just for downloading. What if we the CC wget give more features in the future like recursive? That would be a problem with run.

Lemmmy:
CC wget already doesn't behave like real wget.
CC wget: `wget [URL] [filename]`
Real wget: `wget -O [filename] [URL]`

Thus, using `-r` for run would be ok.

JakobDev:
I seems that nobody has noticed the commit yesterday, so I wrote this comment to trigger a notification.

JakobDev:
The conflict with #225 has now been resolved!

JakobDev:
I just ping @SquidDev because I think he diden't see the new changes.

JakobDev:
@SquidDev Arguments works perfect in the current version:
wget
Just download the program from GitHub and try it yourself. 

-r was a suggestion from @lupus590 . I think we should wait for opinion of other people what argument we sould use.

I'm sorry, but I don't know how to write a test for gradlew.

SquidDev:
> Arguments works perfect in the current version:

Ahh, sorry - not sure how I missed that 🤦‍♂. That said, I think the comment about the file already existing check is still valid, so it's probably worth making that change anyway.

> I'm sorry, but I don't know how to write a test for gradlew.

There's a couple of examples within [this folder](https://github.com/SquidDev-CC/CC-Tweaked/tree/master/src/test/resources/test-rom/spec/programs) - `delete_spec.lua` probably being the most complete. You can either run the tests using `./gradlew check`, or copying the entire `test-rom` folder onto a computer and running `mcfly`. _Hopefully_ it's relatively clear what it all does, but obviously do ask if you're confused!

JakobDev:
@SquidDev Sorry, but I don't understand the delete_spec.lua file. I just see the function expect is many times called.

JakobDev:
@SquidDev Changes are now done!

src/main/resources/assets/computercraft/lua/rom/programs/http/wget.lua
```diff
@@ -1,7 +1,7 @@
 
 local function printUsage()
     print( "Usage:" )
-    print( "wget [-r] <url> [filename]" )
+    print( "wget [run] <url> [filename]" )
```
lupus590: Usage here makes it look like I can do `wget run example.com test.lua` and `wget example.com`.

I know that the second will infer the file name but the first looks like it could be a bad input. The usage might need to examples: `wget run <url>` and `wget <URL> [filename]`

PR #220, opened 2019-05-29
Add a custom renderer for monitor block highlights

### LINKED ISSUE #219: Custom bounding box for monitors
Shamelessly borrowing from MightyPirates/OpenComputers#3072, it would be rather nice if we could hide the bounding box for monitors when hovering over them. This would hopefully make working with multi-block screens a little more visually appealing, as you haven't got an ugly rectangle over it.

One possible improvement we could make to the above suggestion would be to render the whole bounding box aside from edges adjacent to other monitors. Not sure how this well this'll look though, so something to play with.

SquidDev:
This only renders the bounding box on non-screen edges of the monitor, meaning you have an uninterrupted view of the screen when hovering hover.

Personally I like this change, and feel it's rather natural, but would appreciate other's thoughts.

Closes #219

#### A side of a monitor 
The side of a monitor with a bounding box only covering one side

#### The front of the monitor
Front of the monitor, with no visible bounding box

#### The front of the monitor (sneaking)
Front of the monitor while sneaking, meaning a bounding box is rendered.

### DIFF
--- a/src/main/java/dan200/computercraft/client/render/CableHighlightRenderer.java
+++ b/src/main/java/dan200/computercraft/client/render/CableHighlightRenderer.java
@@ -29,13 +29,13 @@ import net.minecraftforge.fml.relauncher.Side;
 import org.lwjgl.opengl.GL11;
 
 @Mod.EventBusSubscriber( modid = ComputerCraft.MOD_ID, value = Side.CLIENT )
-public final class RenderOverlayCable
+public final class CableHighlightRenderer
 {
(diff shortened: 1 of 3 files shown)


lupus590:
The sneaking to show it again is a nice feature.

Looks good to me.

hugeblank:
\> adds a desired behavior
\> does not remove initial behavior
\> maintains initial behavior where desired behavior is unnecessary

You appeal to all sides, what more feedback can one give? :P

PR #221, opened 2019-05-29
Add curl as alias for wget

JakobDev:
curl is also a well known downloader.


SquidDev:
`curl` and `wget` have similar purposes, but are very different programs and operate very differently. I don't think it's fair or accurate to alias the two.

JakobDev:
@SquidDev Thats right, but the must people just run wget <url> or curl <url> to download a file.

Lemmmy:
curl isn't a downloader, `curl https://example.com` doesn't download the file, it outputs it to stdout. To save it, you'd have to use the `-O` option, or redirect stdout to a file, e.g. `curl -O https://example.com` or `curl https://example.com > example.com`

lupus590:
Could curl be a stub program which just prints "you probably want wget [URL] [filename]"

PR #225, opened 2019-06-01
Add more MOTD messages.

JakobDev:


src/main/resources/assets/computercraft/lua/rom/motd.txt
```diff
@@ -3,3 +3,13 @@ View the documentation at https://wiki.computercraft.cc
 Visit the forum at https://forums.computercraft.cc
 You can disable these messages by running "set motd.enable false"
 You can create directories with "mkdir".
+Want to see hidden files? Run "set list.show_hidden true".
+Run "list","ls" or "dir to see all files in a directory.
+You can delete files and directories with "delete" or "rm".
+Use "pastebin put" to upload a program to pastebin.
+Use "pastebin get" to download a program from pastebin.
+Use "pastebin run" to run a program from pastebin without downloading it.
```
SquidDev: Maybe "without saving it first."? Running a file from the internet without downloading it would be quite impressive!

src/main/resources/assets/computercraft/lua/rom/motd.txt
```diff
@@ -3,3 +3,13 @@ View the documentation at https://wiki.computercraft.cc
 Visit the forum at https://forums.computercraft.cc
 You can disable these messages by running "set motd.enable false"
 You can create directories with "mkdir".
+Want to see hidden files? Run "set list.show_hidden true".
+Run "list","ls" or "dir to see all files in a directory.
```
SquidDev: Maybe " "list" (or "ls" for short) to see all files in a directory." Listing all aliases may be a little too much.

src/main/resources/assets/computercraft/lua/rom/motd.txt
```diff
@@ -3,3 +3,13 @@ View the documentation at https://wiki.computercraft.cc
...
+You can delete files and directories with "delete" or "rm".
+Use "pastebin put" to upload a program to pastebin.
+Use "pastebin get" to download a program from pastebin.
+Use "pastebin run" to run a program from pastebin without downloading it.
+If you using a pocket computer, you can play tetris with the "falling" program.
+You can use the "paint" program to draw pictures.
+Use the "edit" program to create and edit textfiles.
```
SquidDev: Maybe "create and edit your programs."? "Textfiles" is a little awkward, and while it can technically edit any plaintext file, it's very much oriented towards Lua!

src/main/resources/assets/computercraft/lua/rom/motd.txt
```diff
@@ -3,3 +3,13 @@ View the documentation at https://wiki.computercraft.cc
...
 You can disable these messages by running "set motd.enable false"
 You can create directories with "mkdir".
+Want to see hidden files? Run "set list.show_hidden true".
+Run "list","ls" or "dir to see all files in a directory.
+You can delete files and directories with "delete" or "rm".
+Use "pastebin put" to upload a program to pastebin.
+Use "pastebin get" to download a program from pastebin.
+Use "pastebin run" to run a program from pastebin without downloading it.
+If you using a pocket computer, you can play tetris with the "falling" program.
+You can use the "paint" program to draw pictures.
```
SquidDev: Can we remove this and the tetris one for now? Ideally you'd only show these when the program is actually available (which they won't be on normal computers, etc...).

JakobDev: If we only show this, when it's available, we need to create a extra motd for pocket and for advanced computers.

SquidDev: Yeah, that's why I was just suggesting removing it for now - short of making motd files dynamic, there's not really a nice way of restricting which ones we show.

JakobDev: I just think, it won't hurt, if people who are using a normal computer see what they can do with a advanced computer.

Lemmmy: I think the tetris one is too long anyway with the extra sentence regarding the pocket computer, so it should be removed until/if there is support for per-device motd

JakobDev:
btw:
Should we now enable MOTD by default?

PR #228, opened 2019-06-03
Add Check to delete.lua

JakobDev:
delete.lua checks now, if it has the permission to delete a file.

### DIFF
--- a/src/main/java/dan200/computercraft/api/filesystem/FileOperationException.java
+++ b/src/main/java/dan200/computercraft/api/filesystem/FileOperationException.java
@@ -1,0 +1,41 @@
+/*
+ * This file is part of ComputerCraft - http://www.computercraft.info
+ * Copyright Daniel Ratcliffe, 2011-2019. Do not distribute without permission.
+ * Send enquiries to dratcliffe@gmail.com
+ */
+
+package dan200.computercraft.api.filesystem;
+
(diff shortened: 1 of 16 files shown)


JasonTheKitten:
I like this, only thing I would say is that I believe that when displaying paths most APIs and programs will ommit rom from the pathname

JakobDev:
@JasonTheKitten It's better to use the full path. I see no reason, why we should remove the /rom part.

@lupus590 An empty string means /, which can't be deleted.

JasonTheKitten:
Ok, I had deleted my comment, yet you still replied...

SquidDev:
I wonder if it'd be worth doing something like this instead:

```lua
local ok, err = pcall(fs.delete, file)
if not ok then
  printError((err:gsub("^pcall: ", "")))
end
```

This way we don't need the special casing of `""`, and handle any of the weird errors that mounts may throw themselves.

I've just added the ability to [verify ROM programs produce a specific output too](https://github.com/SquidDev-CC/CC-Tweaked/blob/master/src/test/resources/test-rom/spec/programs/delete_spec.lua#L43-L46), so it might be worth writing a test for this too.

JakobDev:
@SquidDev It use now pcall, but I don't know, how to write a test.

JakobDev:
Closing because some problems with git. It's better to create a new branch.

PR #236, opened 2019-06-06
Add Check to delete.lua (Update)

JakobDev:
Updated Version of #228 

The gradlew check has now been added.


SquidDev:
Looking good now 👍. Thank you for polishing that off.

PR #237, opened 2019-06-06
Make type.lua accept multiple arguments

JakobDev:


SquidDev:
I'm inclined to say no here, sorry. It just adds complexity to a program which really doesn't need it.

PR #242, opened 2019-06-09
Dynamically load path variable on startup

dirthsj:
The startup program currently builds a static PATH variable during boot.  Loading it dynamically allows extensions of the default programs by addon mods.

For example, I can mount a resource directory to `rom/programs/ccgit`.  If I do so, this dynamic loading will add that to the path, which will allow users to much more easily execute scripts provided by the mod.


SquidDev:
Hrrm, this definitely feels cleaner - just a couple of things which might be worth changing:

 - Maybe remove the `:` prefix from each element, and add it in within the loop (or build up a table, and then `table.concat` it, but not sure if that'll be cleaner).
 - It might be worth inverting the conditions, so a path is added iff `tConditions[path] ~= false`.
 - I don't believe this'll pick up nested paths (such as `fun/advanced`) - I see you've got `tPathStrings`, but it's not used anywhere!

That said, I'm a little confused why you're putting stuff into a sub-directory. It _should_ be possible to mount directly into `rom/programs`.

lupus590:
It might be a good idea to keep the subdirectory thing to reduce collisions, so mod added programs end up a bit like treasure disks are?

dirthsj:
I tried mounting directly to `rom/programs` first.  Doing that fails.  I also agree with lupus590 on this one, it should help distinguish which programs were added for what purpose (as the current system does).

I can certainly invert the conditions, that's an easy fix and actually makes this a bit nicer to read.  I also need to fix the `tPathStrings` issue.  I was actually debating whether or not I should just move `rom/fun/advanced` to `rom/advancedfun` or some such, but decided to go the path of least changes.

SquidDev:
Thinking about this a little more, I wonder if we can merge the two loops. Entirely untested, and probably in need of a cleanup, but I hope this gets the general idea across:

```lua
local tPath = { "/rom/programs" }
local nPathIndex = 1
while nPathIndex <= #tPath do
  local children = fs.list(tPath[nPathIndex])
  local interesting = false
  for _, v in ipairs(children) do
    local child = "/" .. fs.combine(tPath[nPathIndex], v)
    if fs.isDir(child) then
      if tConditions[child] ~= false then tPath[#tPath + 1] = child end
    elseif v:sub(-4) == ".lua" or not v:find(".", 1, true) then
      -- Only add this file to the path if it has executable files within it.
      interesting = true
    end
  end

  -- If this should be added to the path, increment the index. Otherwise
  -- remove it.
  if interesting then
    nPathIndex = nPathIndex + 1
  else
    table.remove(tPath, nPathIndex)
  end
end

local sPath = table.concat(tPath, ":")
```


Hrmr, maybe that's not actually any more elegant then :/. Though I do rather like the checking of `.lua` - means we don't need that manual for the redirection levels.

dirthsj:
Ehh, I don't like modifying tables I'm looping through.  Adding to it was already pushing it IMO.

You *could* do it in one loop but the logic gets messy.  The current way is pretty easy to follow - essentially "list everything in rom/programs recursively" followed by "add things to the path".  I will at least add the lua check, as it's a much cleaner way to handle stuff.

SquidDev:
Hrmr, was about to merge this, and then had all sorts of second thoughts - sorry!

It should currently be possible to add arbitrary folders to the path by putting a startup file in the `autorun` directory. Arguably that's slightly more powerful than this change, as it allows you to only add your directory in specific situations.

I don't know if that's something you've looked into @dirthsj, or if it's just not appropriate for your use-case.

dirthsj:
> It should currently be possible to add arbitrary folders to the path by putting a startup file in the autorun directory. Arguably that's slightly more powerful than this change, as it allows you to only add your directory in specific situations.

The issue with this is that a mod can't put a startup file in the autorun directory (unless something like `autorun/ccgit/startup.lua` would work, haven't checked).

SquidDev:
Hrmr, that seems like a deficiency in the API. 1.14 should work around this, by mounting all mod's `assets/computercraft/lua/rom` folders, which makes it very easy for other people to add to the path. Might be worth looking into how feasible that is for 1.12.

SquidDev:
@dirthsj So the above commit should implement what I've just discussed - can you try that out and see if it's sufficient for your needs? If not, happy to expand the actual API in some way (or take another look at this PR).

PR #253, opened 2019-06-14
Add more tests

JakobDev:


src/test/resources/test-rom/spec/programs/alias_spec.lua
```diff
@@ -0,0 +1,33 @@
...
+    it("displays its usage when given to many arguments", function()
+        
+        expect(capture(stub, "alias a b c"))
+            :matches { ok = true, output = "Usage: alias <alias> <program>\n", error = "" }
+    end)
+
+    it("list alias", function()
+        
+        expect(capture(stub, "alias"))
+            :matches { ok = true, output = "\n\n\n", error = "" }
```
SquidDev: The fact that `alias` isn't emitting anything makes me suspect that `capture` isn't handling anything correctly. That's definitely not the expected behaviour.

src/test/resources/test-rom/spec/programs/alias_spec.lua
```diff
@@ -0,0 +1,33 @@
...
+    it("set alias", function()
+        shell.run("alias test Hello")
+
+        local tAlias = shell.aliases()
+
+        expect(tAlias.test):eq("Hello")
+    end)
+
+    it("clear alias", function()
+        shell.setAlias("test","hello")
```
SquidDev: If we're doing this properly, you'd stub out bits of the shell API and verify that they were called with a specific method/arguments. This'll need some extensions to the test library though.

src/test/resources/test-rom/spec/programs/command/commands_spec.lua
```diff
@@ -0,0 +1,22 @@
...
+
+    it("list commands", function()
+        stub(_G,"commands",{
+            list = function()
+                return {"computercraft"}
+            end
+        })
+        
+        expect(capture(stub, "/rom/programs/command/commands.lua"))
+            :matches { ok = true, output = "Available commands:\n\n", error = "" }
```
SquidDev: Same with `aliases` - this seems like an issue with the `capture` method.

src/test/resources/test-rom/spec/programs/monitor_spec.lua
```diff
@@ -0,0 +1,11 @@
+local capture = require "test_helpers".capture_program
+
+describe("The monitor program", function()
+
+    it("displays its usage when given no argument", function()
+        multishell = nil
```
SquidDev: Just curious as to the reason for this?

JakobDev: Just a copy and paste mistake. I will remove it.

src/test/resources/test-rom/spec/programs/alias_spec.lua
JakobDev: capture has problems with the textutils functions. hello.lua will crash, if you run it with capture.

src/test/resources/test-rom/spec/programs/clear_spec.lua
```diff
@@ -0,0 +1,18 @@
...
+            local x,y = term.getCursorPos()
+            if x == 1 and y == 1 then
+                return true
+            end 
+        end
+
+        term.setCursorPos(5,5)
+        shell.run("clear")
+
+        expect(checkCursor()):eq(true)
```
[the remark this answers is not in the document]
JakobDev: I don't know that this works. Thanks.

JakobDev:
@SquidDev btw: I can't get the check for go.lua work. Maybe you can take a look:
```
local capture = require "test_helpers".capture_program

describe("The go program", function()
    local function setup_turtle(fuel_level)
        stub(_G, "turtle", {
            getFuelLevel = function()
                return fuel_level
            end,
            forward = function()
               print("forward")
            end,
            back = function()
                print("back")
            end,
            up = function()
                print("up")
            end,
            down = function()
                print("down")
            end,
            turnLeft = function()
                print("turnLeft")
            end,
            turnRight = function()
                print("turnLeft")
            end
        })
    end

    it("displays the error without the turtle api", function()
        stub(_G,"turtle",nil)
        
        expect(capture(stub, "/rom/programs/turtle/go.lua"))
            :matches { ok = true, output = "", error = "Requires a Turtle\n" }
    end)


    it("displays its usage when given no argument", function()
        setup_turtle(0)
        
        expect(capture(stub, "/rom/programs/turtle/go.lua"))
            :matches { ok = true, output = "Usage: go <direction> <distance>\n", error = "" }
    end)

    it("out of fuel", function()
       setup_turtle(0)
        
       expect(capture(stub, "/rom/programs/turtle/go.lua forward 1"))
           :matches { ok = true, output = "Out of fuel\n", error = "" }
    end)

    it("wrong direction", function()
       setup_turtle(0)
        
       expect(capture(stub, "/rom/programs/turtle/go.lua nothing 1"))
           :matches { ok = true, output = "o such direction: nothing\nTry: forward, back, up, down\n", error = "" }
    end)
    
    it("run the program", function()
       setup_turtle(1)
        
       local function check_go(direction,result)
            expect(capture(stub, "/rom/programs/turtle/go.lua "..direction.." 1"))
                :matches { ok = true, output = result.."\n", error = "" }
       end
       
       check_go("fd","forward")
       check_go("forward","forward")
       check_go("forwards","forward")
       check_go("bk","back")
       check_go("back","back")
       check_go("up","up")
       check_go("dn","down")
       check_go("down","down")
       check_go("lt","left")
       check_go("left","left")
       check_go("rt","right")
       check_go("right","right")
    end)

end)
```

JakobDev:
@SquidDev The capture function now works better!

JakobDev:
@SquidDev Can you please merge?

SquidDev:
@JakobDev Sorry to the delay in looking at this - I've been rather exhausted due to IRL stuff lately, and so have had to prioritise bugs over reviewing PRs. Sorry, I really shouldn't be leaving it this long, especially given the time you've spent on this.

I've made a couple of improvements to the stubbing system, similar to what I mentioned above - this hopefully makes tests a wee bit cleaner, and means we can separate what we're testing into more separate units.

Can you have a read through, just to make sure you're happy with the changes I've made? If so, I'll go ahead and merge.

One thing to note, I've removed the `apis` program test for the time being - it's just a bit of an ugly program, and I'm not sure it's something which can be checked in an elegant way, so I'm happy to pretend it doesn't exist for the time being.

Issue #232, opened 2019-06-05
Refactor shell completion helpers into a module


SquidDev:
[`rom/startup.lua`](https://github.com/SquidDev-CC/CC-Tweaked/blob/master/src/main/resources/assets/computercraft/lua/rom/startup.lua) has a whole host of completion helpers (`completeMultipleChoice`, etc...). However, these are not usable to people outside the program, which means anyone else who adds completions to programs has to copy and paste them.

It may be worth moving these into a helper module, which can be `require`d by anyone.

A couple of things worth discussing first:

 - I'd be inclined to put this under `shell.completion` or `cc.shell.completion` - thoughts? Basically, I don't want to fill up the top-level module namespace with ComputerCraft specific things, and it'd be good to establish a rule to follow for future modules.

 - A lot of helpers are of the form `if nIndex == 1 then return a(text) elseif nIndex  == 2 then return b(text)`. It'd be nice if we could replace this with a function which just accepts a list of "argument completers"s, so you can do `completing(a, b)`.

Wojbie:
Having proper requireable module for dealing with completion would make lot of programs a bit smaller. And it would make it a bit more simple for users to add it to their programs while teaching them to `require` things.

- As it would be first module that would be added to the system i would say `cc.shell.completion` having 'cc' prefix would make a good separator for build in official modules that exist in `/rom/modules/main/` and ones made by other players.

- As for your proposed `completing(a , b)` function [looking from this line onward in startup for examples ](https://github.com/SquidDev-CC/CC-Tweaked/blob/d8e1c73d265b83d5fc6744c715cd439d4e6d91c8/src/main/resources/assets/computercraft/lua/rom/startup.lua#L57) i only see only few places  where it could be used without making interim functions for each index that just tail call prebuild function with all parameters it needs `if nIndex == 1 then return fs.complete( sText, shell.dir(), true, false ) end` would end up as `completing(function(sText) return fs.complete( sText, shell.dir(), true, false ) end)` for example.

SquidDev:
> as for your proposed `completing(a , b)` function looking from this line onward in startup for examples i only see only few places where it could be used without making interim functions

One thing I was thinking would be to allow writing `completeMultipleChoice` (and potentially other functions) in a more curried style - so you could have `completing(completeMultipleChoice(tLabelOptions), completePeripheralName)` or something. But yes - I don't know if it'd be worth it or not - I think something to try out, and see if it ends up being any cleaner.

[...]

PR #267, opened 2019-07-08
Expose expect as a module

SquidDev:
This moves expect from the bios into a new `craftos.expect` module (using naming as discussed in #232), removing the internal `_G["~expect"]` definition. Aparently people were using this irrespective of the "don't use this" comment, so we need to find another solution.

While this does introduce some ugliness (having to load the module in weird ways for programs, duplicating the expect function in memory), it does allow people to use the function in a supported way, and removes the global ugliness.


dmarcuse:
If people are already using it despite the whole `~expect` thing, it might as well be official I guess.

lupus590:
Don't we have a modules folder in ROM? Couldn't it go in there?

Or am I misremembering?

SquidDev:
@lupus590 That's exactly where it is.

lupus590:
Once again I prove that I don't read things properly

hugeblank:
> Aparently people were using this irrespective of the "don't use this" comment, so we need to find another solution.
Well [I](https://github.com/hugeblank/syssm/blob/master/startup.lua#L23) [feel](https://github.com/hugeblank/qs-cc/blob/master/src/icup.lua#L16) [targeted](https://github.com/hugeblank/qs-cc/blob/master/src/skymodem.lua#L54)... 

In my defense, you can't blame me for skipping the "do not use this" comment for what it does. It simplifies having to scream at the user! Less screaming and more useful code.

I'm glad that this has moved somewhere more accessible.

SquidDev:
> In my defense, you can't blame me

Well, I'm not sure who else to blame!

Issue #270, opened 2019-07-12
Making window more useful


SquidDev:
I think Kepler and I discussed this when #156 was opened, but I never spun it out into a separate issue.

The `window` API is incredibly useful, but I (and I suspect other people) often end up rolling their own implementation because they need a specific bit of functionality. It'd be nice if we could expand the `window` API to include some of the more common use-cases.

Some ideas of things it might be potentially nice to have:
 - [x] Ability to read back the current frame buffer - possibly something as simple as `Window.getLine(line:number): string, string, string`.

 - [x] Extend `Window.reposition(...)` to also take a different redirect buffer - allow changing the parent after creating the window.

 - [ ] Some way to determine when the buffer has changed/is dirty? I'm really not sure how to do this in an efficient way - both a polling (set a flag) or callback (call a function) mechanism have their problems.

 - More? I don't know what sort of other common use-cases there are.

A couple of things to stress here: Firstly, I don't want to sacrifice on performance any more than is necessary. On advanced computers, every program is running through at least one layer of window instances, so we need to keep this as fast as possible.

I also don't want to handle _every_ use case, only ones which are well suited to the current design, and people often find themselves needing. Things like mbs's scrolling buffer are really not suitable.

lupus590:
For the has changed/is dirty thing is there any issue with providing both a callback option and a polling option? Aside from the duplication.

LiquidObsidian:
I think `window.getLine` can be as simple as this:
`window.getLine(line:number, includeColors:bool): string`
If you pass `includeColors` it will return the line with `§<color>` for each color change, and obviously prepending the first character's color.

lupus590:
Possible useful feature, being able to change the colour of a pixel without changing the character in the pixel.

SquidDev:
> For the has changed/is dirty thing is there any issue with providing both a callback option and a polling option? Aside from the duplication.

I'm a _little_ worried about performance, though I doubt it'll be too terrible TBH.

LiquidObsidian:
> This just feels a really ugly API, as it's not really in sync with the rest of how CC's terminals work.

I just took a look at Terminal.java and you're right. Your way makes a lot more sense now lol.

SquidDev:
`getLine` has just been added 🎉, though I'm not entirely happy with the naming. Don't know if people have any alternative suggestions?

absolument:
Since we use the "write" function, the getLine function could be called "read". In the same way as a file handler.

[...]

PR #272, opened 2019-07-13
Add term.getLine and window.getLine

LiquidObsidian:
From issue #270 

Added implementation for `getLine` to `term` and `window`, and added them both to the test spec.

**Usage:**
`term.getLine(y: number): text: string, textColor: string, textBgColor: string`

Example


LiquidObsidian:
I see it failed some of the style checks on travis, I'll leave that up to you.

SquidDev:
I'm really sorry, but can you revert the changes to `TermAPI.java` for the time being?

I kinda want to keep this just on the window API (for now at least) - by putting it on the term API, we force every redirect to implement it and in a specific way. I don't really want to force other redirects to have this, especially as it's _technically_ a breaking change.

The Window API is very much a frame buffer. Other redirects, and (IMO) `term.native` however _shouldn't_ be treated as such.

SquidDev:
Well, I did not mean to push to master, but there we go.

PR #276, opened 2019-07-17
Fixed turtle cable duping

### LINKED ISSUE #273: Turtles can duplicate wired modems and cables
 - Minecraft version: 1.12.2
 - CC: Tweaked version: 1.81.1
 - Detailed reproduction steps:
get a turtle to place a wired modem (the small one not the full block one)
get the turtle to place a modem cable into the same block space
get the turtle to dig
note that the turtle gets two cables and a modem and that there is still a modem in the world
the turtle can dig again to get a second modem

powerboat9:
#273 
When harvesting a block, Block.removedByPlayer() is supposed to be able to cancel Block.harvestBlock() by returning false. Although BlockCable relies on that behavior to act like a multiblock, turtles don't follow it.

http://greyminecraftcoder.blogspot.com/2015/01/summary-of-logic-flow-when-mining-blocks.html


SquidDev:
Thank you for the fix - that's a very good catch! (albeit somewhat embarrassing).

Issue #54, opened 2018-07-26
Global coroutine manager


Lemmmy:
This issue is to discuss a potential future implementation of a global coroutine manager. This would be a system similar to `rednet.run` that allows a user to define add a global coroutine to manage events at the top level. It could also behave somewhat like a process manager.

The main concern with this issue is:- how far should this go? How much do we want to overhaul CraftOS?

BombBloke:
Do we not already have such a manager in the form of the parallel API, which is used to start `rednet.run`? Seems easiest to me to simply expand that - say each coroutine started by a `waitForAny`/`waitForAll` call got a custom parallel table, with additional functions for influencing that particular call. These would be rigged to allow for the probing of coroutine IDs, the addition of new ones, the ability to kill some without ending an "Any" call, the ability to immediately end an "All" call, and so on.

This would only allow a given coroutine to add a new one "at the top level" if already running as a direct child of the top level. That seems preferable to me, though - the result would basically be a lower-level version of the multishell API, and I can't see it needing any changes to CraftOS itself, nor any other existing code.

The only time I've wanted this sort of functionality was when I was building an fs wrapper that relied upon rednet messaging. But I imagine OS developers would love it, as it'd allow for a less hackish TLCO than is currently required.

SquidDev:
I'm closing this for now, as I personally don't wish to implement this.

I'm perfectly happy to talk about potential ways of doing this if someone's prepared to have a stab, similarly will happily consider merging if someone submits a PR to the CC repo. It's just not something that I see CC:T needing to solve directly.

hugeblank:
Apologies for bumping a closed issue, but I have just created something that does this. Maybe it's worth reopening this issue, discussing how I could improve the program, and prepare it for implementation? 
It's called [syssm](https://github.com/hugeblank/syssm). The relevant portions are the chunk of code meant to be placed in the startup.lua file, and the shell program to allow for analysis of all services.

SquidDev:
I've thought about this sort of thing a couple of times, and I feel the "optimal" design would just be closest to what BombBloke already suggested (well, my interpretation of it).

`parallel.waitForAll` and `parallel.waitForAny` return some sort of "coroutine manager" object, with `:add(function)`, `:remove(not sure)` (and maybe `:stop()`) functions. Then the top-level manager can just be exposed on the `os` API.

hugeblank:
I'll have a crack at this, expect a PR in the next few days

[...]

PR #280, opened 2019-07-23
Global Coroutine Manager

hugeblank:
A stab at #54. 

Discussion points:
- Is this sufficient?
- Removal functions are a bit of a mess, I'm considering making a single function within the bundle handle for removal, though that would require retaining all the functions and ugh.
- Is the exposure of the coroutines and boolean that stops parallel execution acceptable?
- How should the top level parallel API be exposed? 
  - Should the removal of the rednet.run and shell coroutines be allowed? (probably not)

To-Do:
- Top level exposure
- Test cases

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
@@ -102,7 +102,7 @@ end
 
 
 
-function createBundle( ... )
+function createGroup( ... )
     local inst = setmetatable(
         { 
             _run = true, 
--- a/src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
@@ -13,48 +13,34 @@ local function create( level, ... )
     return tCos
 end
 
-local external = {
-    __index = {
-        add = function(bundle, ...)
-            local set = create(1, ...)
-            local rfuncs = {}
-            for i = 1, #set do
-                bundle[#bundle+1] = set[i]
-                rfuncs[#rfuncs+1] = function()
-                    local del = #bundle
-                    return table.remove(bundle, del) and true
-                end
-            end
-            return table.unpack(rfuncs)
-        end,
-        stop = function(bundle)
-            bundle.data.run = false
-        end
-    }
-}
-
-
-local function setup( bundle, ... )
-    local routines = {}
-    local data
-    if type(bundle) == "table" then
-        routines = bundle
-        data = bundle.data 
-    else
-        routines = create(1, bundle, ...)
-    end
-    return routines, data
-end
-
-local function runUntilLimit( _routines, _bundleData, _limit )
-    local count = #_routines
-    local living = count
+local function runUntilLimit( _routines, _data, _limit )
+    local run, count = true, #_routines
 
     local tFilters = {}
     local eventData = { n = 0 }
+
+    local function checkDead(r, n)
+        local living = count
+        if r and coroutine.status( r ) == "dead" then
+            _routines[n] = nil
+        end
+        for i = 1, count do
+            if _routines[i] == nil then
+                living = living-1
+            end
+        end
+        if living <= _limit then
+            return n
+        end
+    end
+
     while true do
+        local buffer = {}
+        for i = 1, count do
+            buffer[#buffer+1] = _routines[i]
+        end
         for n=1,count do
-            local r = _routines[n]
+            local r = buffer[n]
             if r then
                 if tFilters[r] == nil or tFilters[r] == eventData[1] or eventData[1] == "terminate" then
                     local ok, param = coroutine.resume( r, table.unpack( eventData, 1, eventData.n ) )
@@ -63,45 +49,82 @@ local function runUntilLimit( _routines, _bundleData, _limit )
                     else
                         tFilters[r] = param
                     end
-                    if coroutine.status( r ) == "dead" then
-                        _routines[n] = nil
-                        living = living - 1
-                        if living <= _limit or (_bundleData and not _bundleData.run) then
-                            return n
+                    if _data then
+                        local last_count = count
+                        run, count = _data()
+                        if _limit > 0 then
+                            _limit = _limit+(count-last_count)
                         end
                     end
+                    local c = checkDead(r, n)
+                    if c then return c end
                 end
             end
         end
-        for n=1,count do
-            local r = _routines[n]
-            if r and coroutine.status( r ) == "dead" then
-                _routines[n] = nil
-                living = living - 1
-                if living <= _limit or (_bundleData and not _bundleData.run) then
-                    return n
-                end
-            end
+        for n = 1, count do
+            local c = checkDead(_routines[n], n)
+            if c then return c end
+        end
+        if not run then
+            return
         end
         eventData = table.pack( os.pullEventRaw() )
     end
 end
 
+local group = {}
+
+function group:add(...)
+    local set = create(1, ...)
+    local rfuncs = {}
+    for i = 1, #set do
+        self._coroutines[self._total+1] = set[i]
+        self._total = self._total+1
+        local del = self._total
+        rfuncs[#rfuncs+1] = function()
+            self._coroutines[del] = nil
+        end
+    end
+    return table.unpack(rfuncs)
+end
+
+function group:stop()
+    self._run = false
+end
+
+function group:waitForAll()
+    runUntilLimit( self._coroutines, self._data, 0 )
+end
+
+function group:waitForAny()
+    return runUntilLimit( self._coroutines, self._data, #self._coroutines-1 )
+end
+
+
+
 function createBundle( ... )
-    local bundle = {}
-    bundle.data = {
-        run = true
-    }
-    bundle = setmetatable(bundle, external)
-    return bundle, bundle:add(...)
+    local inst = setmetatable(
+        { 
+            _run = true, 
+            _coroutines = {},
+            _total = 0
+        }, 
+        { 
+            __index = group 
+        }
+    )
+    inst._data = function()
+        return inst._run, inst._total
+    end
+    return inst, inst:add(...)
 end
 
 function waitForAny( ... )
-    local routines, data = setup( ... )
-    return runUntilLimit( routines, data, #routines - 1 )
+    local routines = create(0, ...)
+    return runUntilLimit( routines, nil, #routines - 1 )
 end
 
 function waitForAll( ... )
-    local routines, data = setup( ... )
-    runUntilLimit( routines, data, 0 )
+    local routines = create(0, ...)
+    runUntilLimit( routines, nil, 0 )
 end
--- a/src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
@@ -1,11 +1,10 @@
-
-local function create( ... )
+local function create( level, ... )
     local tFns = table.pack(...)
     local tCos = {}
     for i = 1, tFns.n, 1 do
         local fn = tFns[i]
         if type( fn ) ~= "function" then
-            error( "bad argument #" .. i .. " (expected function, got " .. type( fn ) .. ")", 3 )
+            error( "bad argument #" .. i .. " (expected function, got " .. type( fn ) .. ")", level+3 )
         end
 
         tCos[i] = coroutine.create(fn)
@@ -14,7 +13,40 @@ local function create( ... )
     return tCos
 end
 
-local function runUntilLimit( _routines, _limit )
+local external = {
+    __index = {
+        add = function(bundle, ...)
+            local set = create(1, ...)
+            local rfuncs = {}
+            for i = 1, #set do
+                bundle[#bundle+1] = set[i]
+                rfuncs[#rfuncs+1] = function()
+                    local del = #bundle
+                    return table.remove(bundle, del) and true
+                end
+            end
+            return table.unpack(rfuncs)
+        end,
+        stop = function(bundle)
+            bundle.data.run = false
+        end
+    }
+}
+
+
+local function setup( bundle, ... )
+    local routines = {}
+    local data
+    if type(bundle) == "table" then
+        routines = bundle
+        data = bundle.data 
+    else
+        routines = create(1, bundle, ...)
+    end
+    return routines, data
+end
+
+local function runUntilLimit( _routines, _bundleData, _limit )
     local count = #_routines
     local living = count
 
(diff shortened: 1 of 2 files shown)


src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
```diff
@@ -55,12 +87,21 @@ local function runUntilLimit( _routines, _limit )
...
+    return runUntilLimit( routines, data, #routines - 1 )
 end
 
 function waitForAll( ... )
-    local routines = create( ... )
-    runUntilLimit( routines, 0 )
-end
+    local routines, data = setup( ... )
+    runUntilLimit( routines, data, 0 )
+end
```
Tyegan: Trailing newline

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
```diff
@@ -55,12 +87,21 @@ local function runUntilLimit( _routines, _limit )
     end
 end
 
+function createBundle( ... )
+    local bundle = {}
+    bundle.data = {
+        run = true
+    }
+    bundle = setmetatable(bundle, external)
```
Tyegan: `local bundle = setmetatable({ data = { run = true } }, external)`

Tyegan: Or if you think this will be extended in the future, at least 
```lua
local bundle = {
  data = { run = true }
}
```

hugeblank: Amended this locally, thank you!

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
```diff
@@ -14,7 +13,40 @@ local function create( ... )
...
-local function runUntilLimit( _routines, _limit )
+local external = {
+    __index = {
+        add = function(bundle, ...)
+            local set = create(1, ...)
+            local rfuncs = {}
+            for i = 1, #set do
+                bundle[#bundle+1] = set[i]
+                rfuncs[#rfuncs+1] = function()
+                    local del = #bundle
```
Tyegan: I don't believe this will behave how you expect, should probably be moved up to an upvalue

Tyegan: Although on the other hand that would invalidate other removal functions, perhaps use the function value as the identifier?

hugeblank: This is one of the discussion points I have in the PR description.

I could use the function, but I'd have to store that as well as the coroutines, in order to find the value to remove later on. I'm not sure if that's what is desired.

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
```diff
@@ -55,12 +87,21 @@ local function runUntilLimit( _routines, _limit )
     end
 end
 
+function createBundle( ... )
+    local bundle = {}
```
SquidDev: Honestly, probably better to do something like this:

```lua
local bundle = setmetatable({
  run = true,
  coroutines = {},
}, blah)
```

Having `bundle` be a list, something to store data and something with methods is a odd.

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
```diff
@@ -14,7 +13,40 @@ local function create( ... )
...
+            return table.unpack(rfuncs)
+        end,
+        stop = function(bundle)
+            bundle.data.run = false
+        end
+    }
+}
+
+
+local function setup( bundle, ... )
```
SquidDev: It feels a little odd to have `setup` either take a bundle or list of functions. I'd be inclined to have the `wait*` functions as methods on the bundle (instead of doing `parallel.waitForAny(bundle)`, have `bundle:waitForAny()`), and then just keep the original `waitFor*` methods as-is.

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
SquidDev: One thing worth considering is what happens if you remove a coroutine while the runner is executing - there's a risk that a coroutine will be skipped. The simple solution would be to copy the coroutine table each time we modify it (or at least remove from it), so the list currently being iterated doesn't change, but will be updated next event.

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
```diff
@@ -34,7 +66,7 @@ local function runUntilLimit( _routines, _limit )
                     if coroutine.status( r ) == "dead" then
                         _routines[n] = nil
                         living = living - 1
-                        if living <= _limit then
+                        if living <= _limit or (_bundleData and not _bundleData.run) then
```
SquidDev: If we add new coroutines to the list, `parallel.waitForAll` will exit when the original number has finished, rather when all of them have. It's probably worth having a "any/all" flag, and then changing the behaviour based on that instead.

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
hugeblank: ooh good idea, that'll be the next step I take.

hugeblank:
@SquidDev I've made amendments to all of the requests here, the last thing I want to do is rename the current "bundles". You proposed TaskManager/TaskRunner, however the function itself doesn't do any running or managing per se. It really is just a _collection of coroutines_. That alliteration led me to come up with another possible name: "corollection". It's short, it describes what exactly the function provides, the only drawback is it's not technically an English word. I'll leave it here for discussion though, it's a fairly strong candidate imo.

hugeblank:
I'm also getting some feedback saying that it being in the `parallel` namespace is enough to provide context that it's a group/bundle/collection of coroutines. I think I'll change the function name to `newGroup` for now, as that's the shortest.

SquidDev:
> You proposed TaskManager/TaskRunner, however the function itself doesn't do any running or managing per se. It really is just a collection of coroutines. 

I _guess_. I'd argue you're going to then going to run that collection, so the difference is moot, but I'll reserve judgement 'til I see the latest iteration of the API :).

> corollection

No. This is one of those clever names which only really makes sense when you know what it's referring to. It's not very descriptive on its own.

> newGroup

Yeah, this is _fine_. Though we can always bikeshed it later if we need to :).

hugeblank:
Darn it! I forgot to change the name to `newGroup`. Expect that in the next commit, along with a crack at implementing the actual purpose of this PR, global coroutine managment.

hugeblank:
Question regarding the implementation in `bios.lua`. When we're talking about just the shell and rednet.run, `waitForAny` is sufficient. But the user isn't stopped from putting in something that could error or end abruptly, so maybe I should use `waitForAll`, and then just test the status of the first two coroutines using the table that should technically be private but isn't. Thoughts?

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
```diff
@@ -31,36 +49,82 @@ local function runUntilLimit( _routines, _limit )
                     else
                         tFilters[r] = param
                     end
-                    if coroutine.status( r ) == "dead" then
-                        _routines[n] = nil
-                        living = living - 1
-                        if living <= _limit then
-                            return n
+                    if _data then
```
SquidDev: Instead of doing this whole `_data` argument and resizing the count based on that, I do wonder if it'd be better to do the "any/all" flag as discussed elsewhere. That way it's pretty much just: when all: "return when a coroutine dies", when any: "return when the live list is empty".

Should drop the whole need for the `_data` parameter, which is a tad ugly.

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
```diff
@@ -14,15 +13,34 @@ local function create( ... )
...
+                living = living-1
+            end
+        end
+        if living <= _limit then
+            return n
+        end
+    end
+
     while true do
+        local buffer = {}
```
SquidDev: I assume you're copying the copying the coroutines here to guard against modifications? Perhaps it would be cleaner to only copy it when the coroutine list is changed - do it within the `:add`/`remove` functions.

It'll mean you'll need to pass in `self` instead of coroutines/data, but that's fine.

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
```diff
@@ -31,36 +49,82 @@ local function runUntilLimit( _routines, _limit )
...
-            local r = _routines[n]
-            if r and coroutine.status( r ) == "dead" then
-                _routines[n] = nil
-                living = living - 1
-                if living <= _limit then
-                    return n
-                end
-            end
+        for n = 1, count do
+            local c = checkDead(_routines[n], n)
```
SquidDev: It's probably also worth dropping dead coroutines from the coroutine list.

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
hugeblank: I guess it makes sense to do that now, since the functionality to stop running outside of any/all has been implemented. Will do.

hugeblank:
Now that I'm thinking about it some, I could return a number instead of a removal function in `:add`. When that number gets called in `:remove`, it gets added to a table, and then any index that gets called after that gets subtracted by 1, any index before will remain unchanged. It'll add an extra table, but allow me to do all of what was requested.

SquidDev:
@hugeblank I'm a little confused about what you're proposing? I may be mis-understanding, but it feels more complex than what is currently there.

hugeblank:
The dilemma I'm having with removing coroutines is that each function added is given a respective removal function, due to the user not knowing the index at which the coroutine the want to remove is within the `_coroutines` table. This is why I don't use `table.remove`, since doing that would shift the index over by one. Setting the index to nil does nothing since the runner checks for the existence of a coroutine before it executes it.

So my answer to that was:
1. In `:add` instead of returning removal functions just return the index the coroutine is being added.
2. Create `:remove` that takes in that index and, removes the coroutines using `table.remove`. That index also gets added to a table that `:remove` scans, subtracting the index by 1 until it lands on the correct index. here's what I just put together:
```lua
function group:remove(id)
    local i = 1
    while _removed[i] <= id do
        id = id-1
        i = i+1
    end
    table.remove(self._coroutines, id)
    _removed[#_removed+1] = id
end
```
3. Change the `runUntillimit` function to take in one of these groups, and work on the changes requested. 

On 3, the `runUntilLimit` function is essentially a group function now, meaning that I could put it in the group 'class'. Is this something that should be done?

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
hugeblank: Do you mean adding a buffer to the group object that `:add/:remove` use, and then when the group is executed, update the `_coroutines` table? I'm confused.

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
lupus590: Or use the null object pattern? Have a fake coroutine which doesn't go dead it just yields in a loop. If all of the coroutines in the list are this fake one then all of the original ones are dead. Saves having to resize the list.

SquidDev:
@hugeblank Just wanted to check the status of this? Do you need any feedback from me on things, or have you just been a little busy?

hugeblank:
I have been busy with stuff other than programming lately so that's been part of it. I also had a question regarding your `:add/remove` buffer idea: https://github.com/SquidDev-CC/CC-Tweaked/pull/280#discussion_r310830285.

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
SquidDev: So instead of having your remove function be:
```lua
rfuncs[#rfuncs+1] = function()
  self._coroutines[del] = nil
end
```

have something like:

```lua
local current = set[i]
rfuncs[#rfuncs+1] = function()
  local new_routines, n = {}, 0
  for i = 1, self._total do
    local co = self._coroutines[i]
    if co ~= current then
      new_routines[n + 1], n = co, n + 1
    else
      self._coroutines[i] = nil
    end
  end
  self._coroutines = new_routines
  self._total = n
end
```

Effectively, removing an item does two things:
 - Delete it from the current set, but don't remove it - we don't want to shrink the list.
 - Replace `_coroutines`/`_total` with a new collection with the item removed.

This means here you just need to this at each iteration - no need to copy the whole `_routines` list:

```lua
local buffer, count = self._coroutines, self._total
```

SquidDev: Of course, you'll need to pass in the `self` object instead of `_routines`/`_limit`/`_data` - I don't think this is an issue if you do the any/all flag as mentioned elsewhere.

hugeblank:
Oh hey also not mentioned  in the commit, added `os.stopTask` that takes in the ID given by runTask. `stopTask` also protects the main coroutines from getting stopped (for now). I also forgot to remove an unnecesary thread from the bios. I'll wait until more changes are requested to commit with a removal.

src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
hugeblank: Right so I've changed how removal works. Since it's now a function of the group 'class' the `add` function returns an ID that is then used later on to remove the thread. Take a look at the latest commit and give me a shout with what should be done.

SquidDev:
Right, sorry for the delay in responding. That looks much better, thank you!

With regards to removing elements, sorry for not being clearer - I have a pretty good mental model of how it should probably work, but have evidently done a poor job of explaining it!

I've put together a quick POC commit at 0fadeba5d3145a34aff3823873fa9d8a28a5ea10, which hopefully illustrates what I was trying to explain. I've tried to explain things in the commit message, but just shout if there's anything I can clarify.

Ideally we'd avoid integer IDs entirely, as they're all too prone to changing as we add/remove elements, and the additional bookkeeping required to keep track of them just doesn't seem worth while.

SquidDev:
Hrmr, odd. I meant to push that change to a new branch, but clearly screwed up. Sorry!

hugeblank:
All's good! It fit what was requested so I had no reason not to use it.

SquidDev:
Oh wow, this has been open a while. Sorry, I'm not a very good maintainer.

I'm afraid I'm going to close this for now. The more I think about the whole idea of #54, the more  uncomfortable I am with it.

The nice thing about CC is that (generally) if you close a program that's it done. I definitely get why such a feature would be useful, but I don't think CraftOS is really set up in a way that any complex background process makes sense.

hugeblank:
all good! I haven't found the drive to work on it either, and anyways this kind of thing can be somewhat replicated by a library.

PR #302, opened 2019-09-30
Fixed turtle property category

Wendelstein7:
Likely was a leftover from copy pasting and a tired programmer.

In for example the Forge settings GUI, the HTTP category description is overwritten with the description regarding to turtles. This should fix that.


SquidDev:
Sorry, just reopening in order to trigger CI. Finally fixed the configuration for it all!

PR #324, opened 2019-12-08
feat: support PATCH

jaredallard:
**What this PR does**: Theoretically brings in `PATCH` support for HTTP. Need to verify, however.

### DIFF
--- a/.github/workflows/main-ci.yml
+++ b/.github/workflows/main-ci.yml
@@ -30,4 +30,4 @@ jobs:
         test -d bin || mkdir bin
         test -f bin/illuaminate || wget -q -Obin/illuaminate https://squiddev.cc/illuaminate/bin/illuaminate
         chmod +x bin/illuaminate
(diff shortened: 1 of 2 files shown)


SquidDev:
Well, this is the new CI's first PR, so no surprise it's kinda broken. I'll look into 
it. 

I wonder if it'd be worth including `TRACE` there too? Not sure how I missed these two when doing the initial change.

jaredallard:
@SquidDev it happens! It might be worth just allowing any method to be passed, a few services have their own custom HTTP methods. Unsure if the Java HTTP client will work with that though.

SquidDev:
Sadly not - Netty [has a predefined list](https://github.com/netty/netty/blob/4.1/codec-http/src/main/java/io/netty/handler/codec/http/HttpMethod.java#L92) of HTTP request methods, and I don't believe we can use anything else.

jaredallard:
OK, I've confirmed this PR is working (for my use-case anyways!)

PR #332, opened 2019-12-29
Make builded Jar downloadable

JakobDev:
The Jar that is builded with GitHub Actions can now be downloaded.


SquidDev:
I wonder if it's worth just uploading the `build/libs` folder instead? This way we keep the Minecraft/version within the mod jar? I don't know if that'll place it in a nested folder or not though.

JakobDev:
I didn't know that a full directory is possible. It's now using the directory.

SquidDev:
Many thanks, and happy new year!

PR #333, opened 2019-12-29
More MOTDs

osmarks:
I've added some extra MOTDs containing useful information new players may not be aware of.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/motd.txt
+++ b/src/main/resources/assets/computercraft/lua/rom/motd.txt
@@ -28,17 +28,16 @@ Use "monitor" to run a program on a attached monitor.
 View all attached peripherals with "peripherals".
 Use "time" to see the in-game time.
 You can set the label of a computer with "label set".
-A computer needs a label to keep its files if it gets destroyed.
+A computer needs a label to keep its files if it is destroyed.
 You can disable auto completion in the shell with "set shell.autocomplete false".
 You can disable auto completion in edit with "set edit.autocomplete false".
 Feeling creative? Use a printer to print a book!
 Files beginning with a "." character are hidden from "list" by default.
-Running "set" lists all currently set settings.
+Running "set" lists the current values of all settings.
 Some programs are only available on advanced computers, turtles, pocket computers, or command computers.
-The "equip" and "unequip" programs let you add or remove peripherals from a turtle or pocket computer without crafting.
+The "equip" and "unequip" programs let you add or remove supported upgrades from a turtle or pocket computer without crafting.
 You can change the color of a disk by crafting it with dye.
 Right-clicking a turtle with a dye changes its color.
-Distances from ender/wireless modems are relative to the full-block modems or computers that they are placed on.
 You can print on a printed page again to get multiple colors.
-Holding the Control and T keys on your keyboard while using a computer terminates the running program.
+Holding the Control and T keys terminates the running program.
 Holding Control and S or R shuts down or reboots the computer you are using.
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/computercraft/lua/rom/motd.txt
```diff
@@ -28,7 +28,17 @@ Use "monitor" to run a program on a attached monitor.
 View all attached peripherals with "peripherals".
 Use "time" to see the in-game time.
 You can set the label of a computer with "label set".
-A computer needs a label to keep its files if it got destroyed.
+A computer needs a label to keep its files if it gets destroyed.
 You can disable auto completion in the shell with "set shell.autocomplete false".
 You can disable auto completion in edit with "set edit.autocomplete false".
 Feeling creative? Use a printer to print a book!
+Files beginning with a "." character are hidden from "list" by default.
```
SquidDev: I wonder if it's worth merging this into the one about `list.show_hidden`. Or if that's too unwieldy.

src/main/resources/assets/computercraft/lua/rom/motd.txt
```diff
@@ -28,7 +28,17 @@ Use "monitor" to run a program on a attached monitor.
...
+Running "set" lists all currently set settings.
```
SquidDev: "all available settings" maybe? It's a little unwieldy a sentence otherwise.

src/main/resources/assets/computercraft/lua/rom/motd.txt
```diff
@@ -28,7 +28,17 @@ Use "monitor" to run a program on a attached monitor.
...
+Some programs are only available on advanced computers, turtles, pocket computers, or command computers.
+The "equip" and "unequip" programs let you add or remove peripherals from a turtle or pocket computer without crafting.
```
SquidDev: "upgrades" rather than "peripherals"

src/main/resources/assets/computercraft/lua/rom/motd.txt
```diff
@@ -28,7 +28,17 @@ Use "monitor" to run a program on a attached monitor.
...
+You can change the color of a disk by crafting it with dye.
+Right-clicking a turtle with a dye changes its color.
+Distances from ender/wireless modems are relative to the full-block modems or computers that they are placed on.
```
SquidDev: Remove this one - it's better suited for the wiki/help files about GPS.

src/main/resources/assets/computercraft/lua/rom/motd.txt
```diff
@@ -28,7 +28,17 @@ Use "monitor" to run a program on a attached monitor.
...
+You can print on a printed page again to get multiple colors.
+Holding the Control and T keys on your keyboard while using a computer terminates the running program.
```
SquidDev: I think "while using the computer" is implicit here, no?

src/main/resources/assets/computercraft/lua/rom/motd.txt
```diff
@@ -28,7 +28,17 @@ Use "monitor" to run a program on a attached monitor.
 View all attached peripherals with "peripherals".
 Use "time" to see the in-game time.
 You can set the label of a computer with "label set".
-A computer needs a label to keep its files if it got destroyed.
+A computer needs a label to keep its files if it gets destroyed.
```
SquidDev: "is" perhaps?

src/main/resources/assets/computercraft/lua/rom/motd.txt
osmarks: It's not necessarily all available settings, as some user programs may use the `settings` API too, and won't be listed in that unless you explicitly set them.

src/main/resources/assets/computercraft/lua/rom/motd.txt
osmarks: It turns out that the [new wiki](https://wiki.computercraft.cc/GPS_Hosts) actually has nothing on GPS hosts as far as I can tell (it looks like search is broken?) so I may have to write some basic documentation for that.

PR #334, opened 2019-12-29
Use a Wadler style pretty printer in the Lua REPL

### LINKED ISSUE #195: Nested strings are misformatted on \n
 - Minecraft version 1.12.2
 - CC: Tweaked version cc-tweaked-1.12.2-1.82.3
 - Detailed reproduction steps: 
image
When printing a string in the LUA console, the '\n' are displayed as actual new lines. The print method behave the same.

When said string is in an array, the backslash character remains in Lua console (see above screenshot). The print method only shows the table reference in that case, so we can't compare.

### LINKED ISSUE #234: Improved pretty-printing within the REPL
The current pretty-printer within the REPL is a little naive: it tries to use `textutils.serialise`, and if that errors just uses `tostring`. While that's sufficient for basic things, it rather stinks when you're trying to dump an object with methods, or recursive table.

It might be worth writing a custom pretty-printer for the Lua REPL, which handles functions, recursive tables, and coroutines (most likely just by `tostring`ing them).

A couple of neat things we could try out:
 - [Colours](https://raw.githubusercontent.com/SquidDev-CC/mbs/master/img/00-lua-serialise.png) <sub>Oh no, more mbs</sub>!
 - Some level of "layout awareness". So short tables (such as `{ 1, 2, 3 }` are displayed on one line, rather than 5.
 - Length limits - if we've got a massive table or string, there's no point trying to print the whole thing, as it'll just take a long time and not fit on the screen! It would be nice if we could restrict how much we print in one go.
    That said, this is kinda complex to do well - mbs tries to be smart, and still leaves a lot to be desired. A naive implementation (only do the first 20 keys, 1000 characters, etc...) may be sufficient.

SquidDev:
- Add a `cc.pretty` module, which provides a [Wadler style pretty printer][app]. I'm not entirely sure of the complexity of this yet, due to the lack of laziness, so we'll need to play around/reason about this a little.

 - The `cc.pretty.pretty` function converts an arbitrary object into a pretty-printed document. This can then be rendered with ".write".

 - Convert the Lua REPL to use the pretty printer rather than `textutils.serialise`.

Closes #234, fixes #195.

This provides the following benefits:
 - Small tables are printed on one line instead
 - Coloured REPL output!
 - Table keys are sorted.

## Still to do
 - [x] Expose a "plain text" version of `cc.pretty.document.write` which returns a (non-formatted) string instead.
 - [x] Think about how/what functionality we want to expose in the public API. I'm personally a big fan of the internals of this, but it may not be suitable for CC itself.
   - Based on this, we might want to have some operator syntax (such as `/` for separating via a line, so you can have `text("foo") / text("bar")`).

 - [x] Handle multiline and wrapped strings correctly. Right now they'll just go off the edge of the screen.
 - [x] Improve documentation a lot. It's currently very abstract and doesn't really go into the underlying details.
 - [x] Tests!

## Screenshots
image

[app]: http://homepages.inf.ed.ac.uk/wadler/papers/prettier/prettier.pdf


SquidDev:
If people are interested, you can now [try out this branch on copy-cat](https://copy-cat.squiddev.cc/gh/pull/334/). It'd be lovely to get some feedback from others on it!

Wendelstein7:
Either this REPL or Copy-Cat are missing scrolling features like MBS provides! But perhaps that's a matter so complex it would fit for its own PR. Other than that, it looks great 👍 and I think it has done a great job at not being too intense of a change to disturb the vanilla CC feel.

SquidDev:
It's probably going to be a wee while until the next CC release, so if people have any comments or suggestions regarding this do shout. There's still time for API changes!

PR #341, opened 2020-01-06
Anonymized GPS

osmarks:
I've made GPS generate a random reply channel for each ping, so that things can't be tracked by ID if they send GPS requests a lot, and to avoid hitting the channel limit if for some reason a computer has a very high ID.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/gps.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/gps.lua
@@ -144,7 +144,7 @@ function locate( _nTimeout, _bDebug )
 
     -- Close the channel, if we opened one
     if bCloseChannel then
-        modem.close( os.getComputerID() )
+        modem.close( channel )
     end
 
     -- Return the response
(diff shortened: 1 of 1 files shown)


SquidDev:
One problem I see with this is that there is now no guarantee that the received message is for this computer. It's possible that two computers will open the same channel, and thus receive conflicting information.

I feel a better solution would be to send _and_ receive on `CHANNEL_GPS`, and then tag each message with a random "query id".

osmarks:
I don't *think* conflicting information is actually a problem, as a correctly-functioning GPS server should send the same position information in response to all requests, and the distances will be different per-receiver anyway.

SquidDev:
That's true. Perhaps we don't need the ID at all then, and can just operate on `CHANNEL_GPS`?

osmarks:
Possibly, though the current channel-based way does slightly reduce the amount of work GPS hosts and clients have to do.

SquidDev:
Eh, I don't think the overhead is too big of a concern here. I'm happy to reuse a channel if it avoids us tramping over the main rednet channel space.

osmarks:
I'm not sure about putting it on the existing channel, as it may require a change to the existing GPS host programs, I'll check.

SquidDev:
And merged, thank you very much!

PR #348, opened 2020-01-16
allow parallel to accept one table arg

lupus590:
this table is then used as the list of function to run in parallel instead of having each function passed as separate arguments

the old interface of separate arguments is still valid

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/parallel.lua
@@ -1,6 +1,6 @@
 
-local function create( ... )
-    local tFns = table.pack(...)
+local function create( listOrFirst, ... )
+    local tFns = type( listOrFirst ) == "table" and listOrFirst or table.pack(listOrFirst, ...)
     local tCos = {}
     for i = 1, tFns.n, 1 do
         local fn = tFns[i]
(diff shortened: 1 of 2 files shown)


lupus590:
Tiem to figure out this unit test stuff

JasonTheKitten:
Looks good, by why is it necessary? It shouldn't affect speed to much, and users can already just use unpack.

lupus590:
it seems wasteful to have the user unpack when parallel is just going to repack it

although now that I think of it tables are passed by reference which means that the user could modify the table while parallel is using it

lupus590:
I don't think the dead build is me.

fatboychummy:
> I don't think the dead build is because of me.

The tests you wrote fail.

lupus590:
Test output is nothing like what I was expecting it to look like

Also more copy and paste errors, that's what I get when doing this so late

SquidDev:
I suspect the problem here is that `create` expects an `n` field (so `{ n = 2, f, g }`), and so you get errors when tables are missing them.

Yeah, I'm afraid that the test output isn't very accessible outside of an IDE. I might be able to mount the test rom as part of the copy-cat CI code, which would mean the tests could be run and viewed online.

lupus590:
Going to have another go at this

PR #350, opened 2020-01-22
fix models

magiczocker10:


SquidDev:
Thanks for the contribution!

Can you post a before and after screenshot of the models? I'm not entirely sure what was broken about them which needed fixing.

magiczocker10:
Some textures were wrong rotated in comparison to the original before the rewrite to json some years ago.

magiczocker10:
Minecraft 1.1 were Turtles are introduced:
MC 1 1

Minecraft 1.14.4 were some textures are wrong rotated:
MC 1 14 4

SquidDev:
Thank you very much!

PR #353, opened 2020-01-25
Create en_ud.json (upside down english)

magiczocker10:


SquidDev:
While the upside-down English (and similar ones, like pirate speak) are fun, they're not something I really want to add to CC:T. It's another maintenance burden which doesn't really add any benefit to users.

Sorry!

JasonTheKitten:
It adds the benifit of consistency to the users gameplay

PR #358, opened 2020-02-03
Update monitor.lua to add mouse_up events

fatboychummy:
Some programs as I've come to find out will act differently upon a long click, compared to a short click.  Others have buttons which act explicitly on the `mouse_up` event.

This change will send `mouse_up` events `0.1` seconds after the `mouse_click` event, allowing the programs which do the above to operate normally.

It will also be more "realistic" in the sense that a player hitting the monitor is essentially a "tap" where the mouse is down then back up in rapid succession.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/monitor.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/monitor.lua
@@ -39,49 +39,25 @@ local function resume( ... )
     return param
 end
 
-local timers = { n = 0 }
+local timers = {}
 
--- add a timer to the timers list with id 'id', xpos 'x', ypos 'y'
-local function addTimer( id, x, y )
-    timers.n = timers.n + 1
-    timers[timers.n] = { n = 3, id, x, y }
-end
-
--- remove a timer from the timers list with id 'id'
-local function removeTimer( id )
-    for i = 1, timers.n do
-        if timers[i][1] == id then
-            local x, y = table.unpack( timers[i], 2, timers[i].n )
-            table.remove( timers, i )
-            timers.n = timers.n - 1
-            return true, x, y
-        end
-    end
-    return false
-end
-
-local ok, param = pcall( parallel.waitForAny,
-function()
-    while true do
-        local _, id = os.pullEventRaw( "timer" )
-        -- when a timer is set off, check if it exists and remove it if so
-        -- also generate "mouse_up" event
-        local ok, x, y = removeTimer( id )
-            if ok then
-            os.queueEvent( "mouse_up", 1, x, y )
-        end
-    end
-end,
-function()
+local ok, param = pcall( function()
     local sFilter = resume()
     while coroutine.status( co ) ~= "dead" do
         local tEvent = table.pack( os.pullEventRaw() )
+        if tEvent[1] == "timer" and timers[tEvent[2]] then
+            local x, y = table.unpack(timers[tEvent[2]], 1, 2)
+            timers[tEvent[2]] = nil
+            tEvent = { n = 4, "mouse_up", 1, x, y }
+            -- hijack event, insert mouse_up event. Child process shouldn't know
+            -- timer event exists.
+        end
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/computercraft/lua/rom/programs/monitor.lua
```diff
@@ -39,7 +39,40 @@ local function resume( ... )
     return param
 end
 
-local ok, param = pcall( function()
+local timers = { n = 0 }
+
+-- add a timer to the timers list with id 'id', xpos 'x', ypos 'y'
+local function addTimer( id, x, y )
+    timers.n = timers.n + 1
```
SquidDev: Instead of making this a queue, why not just a map of ids -> positions. This way insertion and removal will be O(1) rather than O(n). Should probably mean the functions can be inlined too.

src/main/resources/assets/computercraft/lua/rom/programs/monitor.lua
```diff
@@ -39,7 +39,40 @@ local function resume( ... )
...
+            local x, y = table.unpack( timers[i], 2, timers[i].n )
+            table.remove( timers, i )
+            timers.n = timers.n - 1
+            return true, x, y
+        end
+    end
+    return false
+end
+
+local ok, param = pcall( parallel.waitForAny,
```
Wojbie: Why are you using parallel instead on integrating this logic into main event loop?

src/main/resources/assets/computercraft/lua/rom/programs/monitor.lua
```diff
@@ -39,7 +39,40 @@ local function resume( ... )
...
+function()
+    while true do
+        local _, id = os.pullEventRaw( "timer" )
+        -- when a timer is set off, check if it exists and remove it if so
+        -- also generate "mouse_up" event
+        local ok, x, y = removeTimer( id )
+            if ok then
```
plt-abigail: Why's the indentation here all screwy?

fatboychummy: attempting to match indentation that squid has here (4 spaces), must have made a mistake while adding more spaces.

fatboychummy:
Updated it, would we want comments removed for final version (if it gets added?)?

src/main/resources/assets/computercraft/lua/rom/programs/monitor.lua
```diff
@@ -39,15 +39,25 @@ local function resume( ... )
     return param
 end
 
+local timers = {}
+
 local ok, param = pcall( function()
     local sFilter = resume()
     while coroutine.status( co ) ~= "dead" do
         local tEvent = table.pack( os.pullEventRaw() )
+        if tEvent[1] == "timer" and timers[tEvent[2]] then
```
SquidDev: Instead of hijacking the event, I wonder if it'd be better just to do what we do elsewhere and fire another event. So something like:

```lua
if coroutine.status( co ) ~= "dead" and (sFilter == nil or sFilter == "mouse_up") then
  if tEvent[1] == "timer" and timers[tEvent[2]] then
    sFilter = resume ("mouse_up", -- etc...
```

While we're technically sending an event the receiving coroutine will never use, it is a bit more consistent with the other handlers.

fatboychummy: Then keep the old timer event as well I assume?

SquidDev: Yep.

SquidDev: Oh, can you also move it to below the initial resume, so it's with the other events? Sorry!

SquidDev: You'll also need to include the event filtering code, like the other cases.

fatboychummy: I believe that's done now.  Anything I missed?

SquidDev:
Thank you very much!

Issue #262, opened 2019-07-02
fs.getDriveCapacity and fs.stat


xAnavrins:
While coding a drive utility to show disk usage of the internal storage or a disk drive, I came up with a few issue;

It is very tricky to figure out the maximum capacity of a computer or disk storage using only `fs.getFreeSpace` and `fs.getSize` on all files recursively, due to files that are smaller than 500 bytes being counted as 500 bytes used by getFreeSpace, but not by getSize.
`fs.getDriveCapacity('hdd')` would return a drive's max capacity to make that process easier.

`fs.stat(path)` would be a way to check and return a file's size as well as if it is a directory and if it is read-only, without the overhead of calling those three functions for every files.

SquidDev:
> It is very tricky to figure out the maximum capacity of a computer or disk storage

It's not _that_ hard truth be told - the logic is pretty much just `math.max(500, fs.getSize(path))`, for every file and directory. That said, I can definitely see why such a function would be useful.

I think it's worth figuring out what `fs.getDriveCapacity("rom")`<sup>1</sup> should return. Both 0 and the sum of all files on the drive _technically_ make sense.

I've been thinking about an `fs.stat`<sup>2</sup> for a while - it might also be nice to expose file modification/update times in there too, though would need to think about that first.

<sub>1. Though I'd be inclined to have `fs.getCapacity(path)` or something. Just so it's consistent with existing methods.</sub>
<sub>2. Probably `fs.attributes`, to be inline with `lfs`. `stat` is a kinda terrible name.

lupus590:
Definitely agree with both footnotes @squiddev

As for capacity of ROM, 0 as the user can't add more in it, atleast not from inside CC.

xAnavrins:
> It's not _that_ hard truth be told - the logic is pretty much just `math.max(500, fs.getSize(path))`, for every file and directory. That said, I can definitely see why such a function would be useful.

This is what I've been doing, although there are still some weird cases, of which I can't figure out, where this method gives inaccurate results nonetheless.

PR #365, opened 2020-02-07
Add fs.getCapacity and fs.attributes

SquidDev:
See #262 for a bit of background and rationale.

 - `fs.getCapacity` just returns the capacity of the current drive, if    available. This will be nil on rom mounts.
 - `fs.attributes` returns a lfs like table of various file attributes.    Currently, this contains:
    - `access`, `modification`, `created`: When this file was last accessed, modified and created. Time is measured in milliseconds since the epoch, same as `os.epoch("utc")` and what is accepted by `os.date`.
    - `size`: Same as `fs.getSize`
    - `isDir`: same as `fs.isDir`


Wojbie:
I think that `access` would cause a little bit of difference between linux and windows systems. Windows tend to have issues with `lastAccessTime()` from what i seen done before.

SquidDev:
Yeah, that's a good point @Wojbie. I did have some suspicious when I added it, but it seemed to work from my very brief testing.

PR #367, opened 2020-02-08
Allow lua REPL to warn about using local variables

lupus590:
`local varname = value` results in `varname` being inaccessible in the next REPL cycle
sometimes this is unintended and is the result of the user having good habits when writing programs


plt-abigail:
`>  local x = 1`
just defeated your pattern

lupus590:
> 
> 
> `> local x = 1`
> just defeated your pattern

works on ccemux, I didn't try in game

lupus590:
I'm not surprised, this is my first go at lua patterns. I'm committing the fix now

plt-abigail:
Detecting this with patterns is straight-up impossible.. Also, you put in a `5`, not a `%`. Also, my example wasn't the `>`, that's the repl prompt; It was the spaces.

lupus590:
extra white space is valid lua though

or am I wrong on that?

plt-abigail:
Yes, and the pattern you checked in at first didn't match that.
Do this one next: `print(); local x = 1`

lupus590:
I still don't get what you were trying to chek with the `> local x = 1` one

lupus590:
`print(); local x = 1` doesn't match, which I expected and I would say is desired behaviour as I would assume that the person knows what they are doing as that looks a little odd

plt-abigail:
`local x` `local x, y` `local x, y = 1`

MrGoose13:
Coming in a bit late but I don't think this issue is about catching every use of `local`, it's about catching a common mistake newbies make.

I'd recommend a simple pattern, no need to go overboard: `"^%s*local%s+[a-zA-Z_]"`

I made this mistake as a newbie and I'm assuming @lupus590 did, too. A little hint would have been helpful for sure.

lupus590:
I'm now thinking that this might be better on [SquidDev-CC/mbs](https://github.com/SquidDev-CC/mbs).

SquidDev:
Hrmr, it occurs to me actually that we should possibly accent the message. Perhaps make it yellow? Otherwise it ends up rather noisy:

This message being rather hard to read.

It's even worse on pocket computers. Curse your low resolution CC :D:.

SquidDev:
image

It's still a little gross, especially on pocket computers, but I'm honestly not sure what to do about that. I've merged this for now, and will update if something occurs to me.

exerro:
The pattern could just search for `local` anywhere in the input. Of course, this would match it in strings too, but as it's just a generic warning that explains a situation I don't think that's too bad. It's also pretty simple to check for things in strings if you wanted to disable that.

For the message, what do you think of this:
> To access a variable in multiple inputs, remove the `local` keyword.

It's a bit shorter and I think it gets the same point across.

screenshot of warning in an emulator at pocket size

lupus590:
> The pattern could just search for `local` anywhere in the input.

I specifically didn't want to do this in case someone was defining a function which had its own variables which needed to be local.

I do like that shorter message though.

Lemmmy:
If this were MBS, you could just piggyback off it's syntax highlighter

exerro:
Other than effort, is there any reason for the built-in `lua` program not to be super fancy? Using a Lua-side parser would let it give better error messages, handle multiple lines of input, and maybe even remove locals in the input automatically.

SquidDev:
As you say, effort (and licensing). The BSRocks REPL bundled a Lua parser, which allowed us to do multi-line input, and mbs has syntax highlighting. It's all possible, it's just a matter of someone finding time to do it.

PR #368, opened 2020-02-08
Removed rightAlt from edit.lua

Ale32bit:
Some keyboard layouts require rightAlt to type certain character, i.e. [, ], { and } for fr_CH

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
@@ -681,7 +681,7 @@ while bRunning do
 
             end
 
-        elseif param == keys.leftCtrl or param == keys.rightCtrl or param == keys.rightAlt then
+        elseif param == keys.leftCtrl or param == keys.rightCtrl then
             -- Menu toggle
             bMenu = not bMenu
             if bMenu then


SquidDev:
- [Added in 1.74pr16][blame]
 - [Changlog][changelog] says "Fixed the “edit” program being hard to use on certain European keyboards."
 - [Discussion of original bug report](http://www.computercraft.info/forums2/index.php?/topic/22289-)

[changelog]: http://www.computercraft.info/2015/06/27/computercraft-1-74/
[blame]: https://github.com/alekso56/ComputercraftLua/commit/16f3898ccddbab3f9a989c4f5696403635af9553

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -681,7 +681,7 @@ while bRunning do
 
             end
 
-        elseif param == keys.leftCtrl or param == keys.rightCtrl or param == keys.rightAlt then
+        elseif param == keys.leftCtrl or param == keys.rightCtrl then
```
lupus590: `elseif param == keys.leftCtrl or param == keys.rightCtrl or ( not settings.get("edit.ignore_right_alt") andparam == keys.rightAlt) then`

Ale32bit: I agree if `edit.ignore_right_alt` is true by default.

lupus590: I don't think that would be best for majority user experience.

SquidDev:
Cross-posting from Discord some potential useful comments from Discord.

> can we just stick to left ctrl
> it's better for everyone

> I wonder if a more sane thing would be to make any (non-command) character event close the menu. Or if that's equally confusing.

Honestly, I have no clue what the correct solution is, because I'm fairly sure this change was originally made to avoid a very similar bug.

lupus590:
could we make the menu key a setting?

SquidDev:
Thanks past me for closing this issue and leaving future me to comment why.

As mentioned at the top, this change was added to fix a similar bug. I don't really want to remove it, as that will come back, and frankly configuration options are just dodging the problem.

I feel a cleaner solution is probably to pass through unhandled `char` events back to the main editor. So <kbd>s</kbd> will save, but <kbd>]</kbd> will type a "]".

It might also be nice to handle "actual" control keys, so <kbd>Ctrl-S</kbd> will save a file, without having to release <kbd>Ctrl</kbd> first. But that's largely orthogonal to this.

PR #372, opened 2020-02-12
Make list.lua use fs.attributes

JakobDev:
This make list.lua display data from the new function fs.attributes() added in #365 when it is runed with -l. The numbers are currently cutted because they are too large to be displayed on the screen. We need a function that turns the numbers into a human readable format.


lupus590:
sounds like we need a textutils.humanise function

JakobDev:
i had found the already existing function os.date().

lupus590:
Could we change the date format to yyyy/mm/dd as it is more international.

JakobDev:
You mean 2020/02/13? For yyyy is not enough space. It's better to use yy.

src/main/resources/assets/computercraft/lua/rom/programs/list.lua
```diff
@@ -2,8 +2,27 @@
 local tArgs = { ... }
 
 -- Get all the files in the directory
-local sDir = shell.dir()
-if tArgs[1] ~= nil then
+local sDir
+if #tArgs == 0 then
+    sDir = shell.dir()
+elseif string.sub(tArgs[1],1,1) == "-" then
```
SquidDev, line 8: This argument parsing code doesn't handle several cases you might expect it to do, such as `ls -l -a foo`, or `ls foo -a`.

I wonder if it's worth adding something like cloud-catcher's [arg-parse] to CC, which would mean we can have something which is a little more general and well behaved.

[arg-parse]: https://github.com/SquidDev-CC/cloud-catcher/blob/master/src/host/argparse.lua

SquidDev: Probably worth thinking about how the shell auto-completion should handle arguments too. Currently `ls -l f` won't auto-complete into `ls -l foo`.

src/main/resources/assets/computercraft/lua/rom/programs/list.lua
```diff
@@ -31,8 +50,33 @@ end
...
+        table.insert( tSingeEntry, os.date( sDate, tAttr["created"] ) )
+        table.insert( tSingeEntry, os.date( sDate, tAttr["modification"] ) )
+        table.insert( tPrint, tSingeEntry )
+    end
+    return tPrint
+end
+
+if bShowAttr then
+    tPrint = {{"Name","Type","Size","Created","Modified"}}
+    tPrint = writeFileTable( tPrint, tDirs )
```
SquidDev, line 73: I wouldn't bother to return `tPrint` in `writeFileTable`. `writeFileTable( tPrint, tDirs )` ends up being cleaner than `tPrint = writeFileTable( tPrint, tDirs )`.

src/main/resources/assets/computercraft/lua/rom/programs/list.lua
```diff
@@ -31,8 +50,33 @@ end
...
+        table.insert( tSingeEntry, tostring( tAttr["size"] ) )
+        table.insert( tSingeEntry, os.date( sDate, tAttr["created"] ) )
+        table.insert( tSingeEntry, os.date( sDate, tAttr["modification"] ) )
+        table.insert( tPrint, tSingeEntry )
+    end
+    return tPrint
+end
+
+if bShowAttr then
+    tPrint = {{"Name","Type","Size","Created","Modified"}}
```
SquidDev, line 72: I'd leave off the "Created" column - as you've mentioned elsewhere, there's limited space, and IMO we don't need to include both. This'd mean we can use `%Y-%m-%d` instead of `%y`.

Ideally we could also leave off "Type" and just colour "Name", but I guess `tabulate` doesn't allow that.

src/main/resources/assets/computercraft/lua/rom/programs/list.lua
```diff
@@ -31,8 +50,33 @@ end
...
+    tPrint = writeFileTable( tPrint, tDirs )
+    tPrint = writeFileTable( tPrint, tFiles )
+    textutils.tabulate( table.unpack( tPrint ) )
```
SquidDev, line 75: Should we use `pagedTabulate` here?

SquidDev (changes requested):
Sorry for merging master into it - there were a couple of problems that CI really should have picked up on and hadn't for whatever reason. I think the latest changes to the build script have fixed that.

Most of the problems it's identified are petty formatting stuff, but there's a couple of places where you're declaring globals - would you be able to fix them?

JakobDev:
I had now added a setting which controls the columns. Created is now disabled by default.

I had thinked the same of colouring dirs green, but textutilstabulate doesn't allow this.

I know about the problems of my argparsing function, but I think, we should keep this, unless we have a function for that.

PR #378, opened 2020-03-03
add _requireFromDisk function

lupus590:
`require` but we remove the package from `package.preload` first

it should make API/module testing easier for some people


lupus590:
might be better to use https://github.com/SquidDev-CC/mbs/pull/24

SquidDev:
As mentioned in SquidDev-CC/mbs#23/SquidDev-CC/mbs#24, this isn't something I want to implement.

PR #381, opened 2020-03-14
Create ko_kr.lang

mindy15963:
## A quick checklist


SquidDev:
And merged, sorry for the delay!

PR #392, opened 2020-04-07
Configureable turtle speed

lupus590:
A turtle's ability to move and interact with the world are two of its greatest strengths, so what would help people who are writing programs which do a lot of moving more than allowing them to make the turtle move faster?

Basically someone wanted to be able to test their mining program and found that the turtle moving slowly was tedious.

Draft pull request as I need to figure out how to get the config menu to be a bit nicer (localisation strings probably)

### DIFF
--- a/src/main/java/dan200/computercraft/shared/turtle/core/TurtleBrain.java
+++ b/src/main/java/dan200/computercraft/shared/turtle/core/TurtleBrain.java
@@ -44,7 +44,7 @@ import java.util.concurrent.TimeUnit;
 
 public class TurtleBrain implements ITurtleAccess
 {
-    private static final int ANIM_DURATION = 8;
+    private static final int ANIM_DURATION = ComputerCraft.turtleAnimationDuration;
 
     private TileTurtle m_owner;
     private ComputerProxy m_proxy;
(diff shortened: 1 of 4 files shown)


lupus590:
Ideally, the config option would be something like `turtle speed multiplier` instead of `turtle animation duration`. Maybe I will try that once I get this better and the speed multiplier might need more rework on the turtle movement stuff.

src/main/java/dan200/computercraft/shared/turtle/core/TurtleBrain.java
```diff
@@ -44,7 +44,7 @@
 
 public class TurtleBrain implements ITurtleAccess
 {
-    private static final int ANIM_DURATION = 8;
+    private static final int ANIM_DURATION = ComputerCraft.turtleAnimationDuration;
```
dmarcuse, line 47: If `ANIM_DURATION` is replaced everywhere else by `ComputerCraft.turtleAnimationDuration`, why not simply remove this field?

lupus590: This is why it's a draft, I new I wasn't finished.

lupus590:
Closing as I think #393 is better.

PR #393, opened 2020-04-07
Turtle Speed Multiplier Config Option

lupus590:
Alternative to https://github.com/SquidDev-CC/CC-Tweaked/pull/392


SquidDev:
This is something I tried [after this discussion][tools], and anything beyond a small speedup ended up looking really poor. While I appreciate this may be useful for testing, all in all it's not really something I want to allow people to change for now. Sorry.

[tools]: https://github.com/dan200/ComputerCraft/issues/274#issuecomment-302953395

lupus590:
Perhaps a note that the default should be used unless the user is testing a turtle program? Or maybe a boolean value `turtle_test_animation_speed`?

PR #395, opened 2020-04-09
Add Danish translation

ChristianLW:
Add Danish translations for items and blocks, as well as the connect/disconnect message and misc tooltips. Commands and config have not been translated.


SquidDev:
Fantastic, thank you very much!

PR #399, opened 2020-04-12
Add argparse module

JakobDev:
This module contains 2 functions:

simpleParser:
This function just parse the arguments without checking anything.

advancedParser:
This is object orientated. It has a help function and checks the arguments for existing. I'm planing to add features like needed arguments and type checking later.

This is currently early beta. It misses features and checks. This is just to hear opinions if you like this or not.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/modules/main/.ignoreme
+++ b/src/main/resources/assets/computercraft/lua/rom/modules/main/.ignoreme
@@ -1,4 +1,0 @@
---[[
-Alright then, don't ignore me. This file is to ensure the existence of the "modules/main" folder.
-You can use this folder to add modules who can be loaded with require() to your Resourcepack.
-]]
--- a/src/main/resources/assets/computercraft/lua/rom/modules/main/argparse.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/modules/main/argparse.lua
@@ -1,0 +1,85 @@
+local function simpleParser( tArgs )
+    local parameters = {}
+    local arguments = {}
+    local options = {}
+    local currentArgument = nil
+    for k, v in ipairs( tArgs ) do
+        if currentArgument ~= nil then
+            arguments[currentArgument] = v
+            currentArgument = nil
+        elseif v:find("--",1,true) == 1 then
+            v = v:sub(3,-1)
+            currentArgument = v
+        elseif v:find("-") == 1 then
+            v = v:sub(2,-1)
+            for c in v:gmatch"." do
+                options[c] = true
+            end
+        else
+            table.insert( parameters, v )
+        end
+    end
+    return parameters, options, arguments
+end
+
+local function advancedParser()
+    parserTable = setmetatable({options={},arguments={}},{})
+    function parserTable.addOption(parserTable, char, description)
+        parserTable.options[char] = {description=description}
+    end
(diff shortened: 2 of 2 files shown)


SquidDev:
Right, sorry for the delay - took me a while to get my thoughts in order.

I guess first thing to discuss is what forms of things we need to accept:

 - **Flags/switches:** Take no arguments, and just enable an option. So, for instance `-l` or `-a` in the `ls` command (and, of course, `-la`). We should also accept long-form flags though, such `--all`.

 - **Options:** Like flags, but take an additional value. Again, we should accept long and short forms for this, so `--color=always`, `--color always`, `-calways` and `-c always`.

 - **(Positional) Arguments:** Basically anything not starting with a `-`. So the filename `foo` in `ls` foo`. If we see a `--` on the command line, anything after it should be considered an argument.

There's definitely more we _could_ support, such as options which don't require a value, but we can probably ignore that for now.

From that, I guess there's a couple of points on the implementation:
 - We should allow giving multiple names for options/flags. So instead of `:addOption("a")`, we should probably accept `:addOption({"a", "all"})`.
 - I'm not sure how feasible `simpleParser` is, just due to all the potential ambiguities (such as `-calways`, and `--color always`).

   However, I think we could expose a simpler API than the full blown `:parseArgs` - perhaps something of the style of C's `getopt`. Anyway, something which can probably be deferred 'til later :).

As far as the API goes, I would appreciate some thoughts from other people. I'm personally a massive fan of [optparse-applicative] and [cmdliner], though I don't think either are really suitable for Lua :). Might be worth having a look to see how how existing [Lua] and [Python] parsers handle this - I dunno, I'm probably overcomplicating things.

Sorry, this was a bit of a braindump, so not sure how useful any of this was. I don't think there's anything you need to do immediately - would be good to get other people's thoughts first.

[optparse-applicative]: https://hackage.haskell.org/package/optparse-applicative-0.15.1.0/docs/Options-Applicative.html
[cmdliner]: https://erratique.ch/software/cmdliner/doc/Cmdliner.Arg.html
[lua]: https://github.com/luarocks/argparse
[python]: https://docs.python.org/3/library/argparse.html

JakobDev:
Are there any other opinion? Should I just update my code with the suggestions from @SquidDev or do we need another concept?

neumond:
This will require additional checking between short & long options on user side, e.g. `-h` and `--help`, `-v` and `--verbose`. As well as type conversions/validation. I would rather put positional arguments as a member of result table instead separate table. I also often use subcommands with different parsers. Probably I'm little bit opinionated after python's argparse. It's also great to be able to add good description for program in help output.

What I would like to see

```lua
p = parser('List files')  -- program description is optional
p.addFlag('--all').alias('-a')
p.addArgument('--width').alias('-w').converter(function(w) return tonumber(w) end)
p.addArgument('dir')
p.addArgument('moredirs').variadic()

r = p.parse('-aw', '3', '/rom', '/rom/help', '/disk')
-- {all=true, width=3, dir='/rom', moredirs={'/rom/help', '/disk'}}
```

SquidDev:
I think I'm going to close this. While it would be nice to have a standard argument parsing library, I'm not comfortable right now committing to an API, and there's not really been enough feedback from people for me to matke a final call on what it should look like.

Sorry :/.

PR #400, opened 2020-04-13
Add Gist client to ComputerCraft as a Pastebin alternative

MCJack123:
Gist is a service provided by GitHub that allows uploading small sets of text files, similar to Pastebin.

Advantages of Gist over Pastebin:
* Hosted by GitHub, meaning more reliable service
* Multiple files can be uploaded at once
* Descriptions can tell people what your files do
* Markdown & automatic syntax hightlighting online
* Files can be edited & deleted after uploading
* Tied to your GitHub account

Disadvantages of Gist:
* Requires you to log in to upload files (users are assisted in making a token for login)
* Longer IDs mean more typing (or pasting)

Since Pastebin has been having some downtime recently, and since Gist has more features than Pastebin, it may be good to have Gist available for users alongside or instead of Pastebin. I've been distributing `gist.lua` independently for a while, but I just got around to polishing it up and adding some more features, and I think it would be a great addition to the standard CraftOS tools.

The version of `gist.lua` I have included is the same as the one I ship with my emulator, CraftOS-PC. It has a few parts that are left over, but if necessary I can strip those out here (however, that may make any possible updates slightly harder to port to CC:T). It has six commands: `put`, `edit`, `delete`, `get`, `run`, and `info`. It simulates most of the behavior of `pastebin`, but also allows downloading and uploading multiple files, as well as some other new functions that are Gist-specific.

One drawback that I have had to work around is that Gist does not allow anonymous upload, but the script points the user to an online article that shows steps on creating a personal access token. (Should the program list the steps itself?) This can then be pasted into ComputerCraft, and the token will be stored in `gist.id` in the `settings` API, allowing future uploads on the same computer without having to generate another token. (Is storing the user's personal token in plaintext too insecure?)

This commit includes the `gist` program, a help document, and a test script which has passed all tests. (I will need to add some more tests since I forgot to add a few for the new commands.) Shell autocomplete has also been added, though it is somewhat incomplete due to the complex variation of the command's syntax.

You may want to [take a look at the help document](https://github.com/MCJack123/CC-Tweaked/blob/9c75003299b730649fec9ad27c611c119e942363/src/main/resources/assets/computercraft/lua/rom/help/gist.txt) for information about the features of `gist.lua` and how to use it.


src/main/resources/assets/computercraft/lua/rom/programs/http/gist.lua
```diff
@@ -0,0 +1,669 @@
...
+    local jsondata = ([[{
+  "description": %s,
+  "public": true,
+  "files": {
+%s
+  }
+}]]):format(data.description and '"' .. data.description .. '"' or "null", jsonfiles)
+    write("Connecting to api.github.com... ")
+    local handle
+    if http.patch ~= nil then handle = http.patch("https://api.github.com/gists/" .. id, jsondata, headers)
```
SquidDev, line 618: This should be the following right:

```lua
http.post {
  url = "https://api.github.com/gists/" .. id,
  body = jsondata,
  headers = headers,
  method = "PATCH"
}
```

Neither `http.patch` nor the extended `http.request` syntax are supported in CC:T. The latter is from CCTweaks, not sure where the former is from.

src/main/resources/assets/computercraft/lua/rom/programs/http/gist.lua
```diff
@@ -0,0 +1,669 @@
...
+-- THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+-- IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+-- FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+-- AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+-- LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+-- OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+-- SOFTWARE.
+--
+
+json = { _version = "0.1.2" }
```
SquidDev, line 33: CC ships its own JSON encoder - see `textutils.serializeJSON`. It's probably worth adding a JSON decoder there too, so other people can make use of it. Really I'd rather we rolled our own, or got permission to include someone's, so we don't have to worry about licensing.

Ideally we'd support NBT style JSON too, but I can live if we don't do that for now.

src/main/resources/assets/computercraft/lua/rom/programs/http/gist.lua
```diff
@@ -0,0 +1,669 @@
...
+                if event == "http_success" and param1 == _url then
+                    handle = param2
+                    break
+                elseif event == "http_failure" and param1 == _url then
+                    print("Failed: ", param2, param3)
+                    return 3
+                end
+            end
+        else print("Failed: " .. err) return 3 end
+    else print("Failed: This version of ComputerCraft doesn't support the PATCH method. Update to CC: Tweaked or a compatible emulator (CraftOS-PC, CCEmuX) to use 'edit'.") return 3 end
```
SquidDev, line 663: Update to CC:T 🤔.

src/main/resources/assets/computercraft/lua/rom/programs/http/gist.lua
```diff
@@ -0,0 +1,669 @@
...
+local helpstr = "Usages:\ngist put <filenames...> [-- description...]\ngist edit <id> <filenames...> [-- description]\ngist delete <id>\ngist get <id> <filename>\ngist run <id> [arguments...]\ngist info <id>"
+
+if #args < 2 then
+    print(helpstr)
+    return 1
+end
+
+if not http then
+    printError("Gist requires http API")
+    if _G.config ~= nil then printError("Set http_enable to true in the CraftOS-PC configuration")
```
SquidDev, line 495: Hrm, might just be better to say "ComputerCraft's configuration" - the filename changes between versions. I'll update pastebin and wget to be the same.

src/main/resources/assets/computercraft/lua/rom/programs/http/gist.lua
MCJack123: `http.patch` is added in CraftOS-PC. I thought the `http.request` syntax was what CC:T used, but I'll swap that out for the correct syntax.

(I'll probably also need to add the table syntax for http.post into CraftOS-PC... but that's irrelevant here.)

src/main/resources/assets/computercraft/lua/rom/programs/http/gist.lua
MCJack123: The JSON library's license was something I was worried about. Most of the JSON operations required involve extensive decoding, so just using `textutils.serializeJSON` won't cut it. I'll look for a public domain decoder, but if you have plans on adding a custom decoder I'll use that instead.

SquidDev:
Oh, congrats on getting issue 400 too!

lupus590:
Would it be worth adding hastebin and other clients or simular services?

SquidDev:
@lupus590 It's worth noting that hastebin has a 30 day expiry, which means it's not fantastic for storing programs - only temporarily uploading them.
I don't really think there's any benefit to supporting multiple paste sites if they all have the same features - it's honestly just more scope for confusion.

MCJack123:
I've made some changes to the code, if you haven't seen. I'm thinking that the `gist` command would be used more as a distribution strategy, rather than as a quick way to upload files. In that case, no token generation or account is necessary, so the ease of use would be equal to `pastebin`'s. I don't know how commonly `pastebin` is used to upload files, though, so I may be underestimating the demand for a quick file upload solution from CC.

SquidDev:
I have seen the changes, thank you. I'm afraid I haven't had a chance to have another proper read through, but it is looking much better!

As far as functionality goes, I think the "selling point", as it were, has to be with the uploading and editing functionality. It's not as nice, but you can download gists with `wget` afterall.

As far as `pastebin put` usage goes, I don't know. Going through https://pastebin.com/archive/lua reveals a fair number of CC programs, so I guess it's fairly common, though probably more so amongst the casual users of the mod.

SquidDev:
>  I think the "selling point", as it were, has to be with the uploading and editing functionality.

Well, and dealing with multiple files  - I wonder if it'd be worth allowing people to upload/dovnload entire directories too?

lupus590:
One thing I would ask is to make this an API with the program maybe being a light wrapper around that API. All too often people do `shell.run("pastebin get code filename")` to get their APIs

MCJack123:
I've updated the program with a few new features that were suggested here, including folder support and the ability to load the program as an API. I ended up putting the API in `rom/programs/http` inside the program itself instead of making a new API file in `rom/apis`, but if we decide it's better to have the API globally available it can easily be moved into its own API file.

The API has the following functions exported:
```
gist.get(id)
gist.getAll(id)
gist.run(id, ...)
gist.info(id)
gist.put(files[, description[, id[, interactive]]])
gist.delete(id[, interactive])
```

These should provide all of the functionality available in the actual `gist` program.

src/main/resources/assets/computercraft/lua/rom/programs/http/gist.lua
```diff
@@ -0,0 +1,335 @@
...
+        setTextColor(colors.blue)
+        write("gist")
+        setTextColor(colors.yellow)
+        print("' checkbox on step 7 (under 'Select scopes'). Once done, paste it here.")
+        setTextColor(colors.lime)
+        write("PAK: ")
+        setTextColor(colors.white)
+        local pak = read()
+        if pak == nil or pak == "" then error("Invalid PAK, please try again.") end
+        settings.set("gist.id", pak)
```
SquidDev, line 43: Do we want to save the settings afterwards?

src/main/resources/assets/computercraft/lua/rom/programs/http/gist.lua
```diff
@@ -0,0 +1,335 @@
...
+-- * Otherwise, if there's more than one file but only one *.lua file, retrieves the Lua file
+-- * Otherwise, retrieves the first Lua file alphabetically (with a warning)
+-- * Otherwise, fails
+
+--- Retrieves one file from a Gist using the specified ID.
+-- @tparam string id The Gist ID to download from. See above comments for more details.
+-- @tparam[opt] any showMessages Set to any true value to show messages while connecting.
+-- @treturn string|nil The contents of the specified Gist file, or nil on error.
+-- @treturn string|nil The name of the file that was chosen to be downloaded, or nil on error.
+function get(id, showMessages)
```
SquidDev, line 63: Instead of a `showMessages` boolean, I wonder if a `progress: function(msg:string)` argument would be better. So we'd just have:


```lua
expect(2, progress, "function", "nil")
progress = progress or do_nothing
-- …
progress("Connecting to api.github.com")
```

src/main/resources/assets/computercraft/lua/rom/programs/http/gist.lua
```diff
@@ -0,0 +1,335 @@
...
+        local d = handle.readAll()
+        handle.close()
+        return d
+    end
+end
+
+local function setTextColor(c) if term.isColor() then term.setTextColor(c) elseif c == colors.white or c == colors.yellow then term.setTextColor(colors.white) else term.setTextColor(colors.lightGray) end end
+
+local function requestAuth(headers, interactive)
+    if settings.get("gist.id") ~= nil then headers.Authorization = "token " .. settings.get("gist.id") elseif interactive then
```
SquidDev, line 27: I wonder if it's worth returning a boolean about whether auth could be successfully fetched? So one can do `if not requestAuth(headers) then return nil end`.

src/main/resources/assets/computercraft/lua/rom/programs/http/gist.lua
```diff
@@ -0,0 +1,335 @@
...
+-- @treturn string|nil The name of the file that was chosen to be downloaded, or nil on error.
+function get(id, showMessages)
+    expect(1, id, "string")
+    local file
+    if id:find("/") ~= nil then id, file = id:match("^([0-9A-Fa-f:]+)/(.+)$") end
+    if id == nil or not id:match("^[0-9A-Fa-f][0-9A-Fa-f:]+[0-9A-Fa-f]$") then error("bad argument #1 to 'get' (invalid ID)", 2) end
+    if id:find(":") ~= nil then id = id:gsub(":", "/") end
+    if showMessages then write("Connecting to api.github.com... ") end
+    local handle = http.get("https://api.github.com/gists/" .. id)
+    if handle == nil then if showMessages then print("Failed.") end return nil end
```
SquidDev, line 71: Just a couple of things:

 - It's probably nicer to return `nil, <failure reason>` rather than displaying a message here. That's more in line with what's done elsewhere.
 - Can we attempt to extract a reason from the response message and/or handle? For instance, "Cannot connect", or the JSON `message` field.

src/test/resources/test-rom/spec/programs/http/gist_spec.lua
```diff
@@ -0,0 +1,260 @@
+local capture = require "test_helpers".capture_program
+
+local gistout = [[{
+  "url": "https://api.github.com/gists/aa5a315d61ae9438b18d",
```
SquidDev, line 4: Maybe shift this into the `test-rom/data` folder, and just read it in instead?

src/main/resources/assets/computercraft/lua/rom/programs/http/gist.lua
MCJack123: This function is internal to the API, but I added a return value to improve code readability anyway.

src/main/resources/assets/computercraft/lua/rom/programs/http/gist.lua
MCJack123: I wasn't sure where to save it at first, but I updated it to save to `/.settings`.

src/main/resources/assets/computercraft/lua/rom/programs/http/gist.lua
MCJack123: I've added this. `showMessages` was originally intended to be used only by the `gist` program, but I realize this may be useful for other programs. I didn't do this for `put` and `delete` since they do need a boolean value to know whether to prompt for an ID.

src/main/resources/assets/computercraft/lua/rom/programs/http/gist.lua
MCJack123: I've fixed the functions so they now return `nil` + an error message. They also print `Failed.` through `progress` if specified. I also made it so the message is added if available.

MCJack123:
I've moved all of the Gist handling code into the `cc.http.gist` module. The `gist.lua` program now uses this module for all of its functions, and I added a test script for the module. I've also applied the changes you requested.

SquidDev:
Sorry for the delay in getting to this, I spent last night and this morning writing up some comments, but realised I'm increasingly unsure if I'm going to merge this. 

It's a good program to have out there - Howl has done gist uploading for a while, and it'd definitely be nice to have a standardised API out there. However, I definitely think it's something best done outside of CC:T.

I realise that you've put a lot of effort into this PR, so I'm really sorry for giving you a run about the Wrekin.

Issue #363, opened 2020-02-07
Update Cobalt to 0.5.1


SquidDev:
With the help of Abigail, I've been making some updates to Cobalt to bring it more in line with PUC Lua, and incorporate some versions of Lua 5.2/5.3.

Obviously it'd be nice to update CC:T to use these features. However, there's a couple of things which will need testing to ensure they do not break existing programs too badly.

[You can see the list of changes since the last CC:T update](https://github.com/SquidDev/Cobalt/compare/v0.5.0...master), but a brief summary of things which may make a noticeable difference:

 - The `__len` metamethod may now be used by tables.

 - Strings now accept hexadecimal, unicode and the zap (`\z`) escape code. Please make sure you're only using valid escape codes - this can be done by chucking your code into [illuaminate](https://squiddev.cc/illuaminate/), or running it through Lua 5.3.

 - Java functions are no longer tail called. This means that `return error("bad argument")` will no longer lay blame at the calling function. This should be replaced with `error("bad argument", 2)`, or use the [`cc.expect`](https://tweaked.cc/module/cc.expect.html) library.

#### Some other niceties which won't break anything, but worth pointing out.
 - Add Lua 5.3's `utf8` library.
 - `math.log` also accepts a "base" argument (e.g. `math.log(123, 2)` is log<sub>2</sub>(123)).
 - Parser error messages now include the token where the error occurred.
 - `table.unpack` now uses `__len` and `__index` metamethods.

Lemmmy:
Slight issue with the errors in tail calls, the one I noticed was in the peripheral API. Example:

```lua
local mon = peripheral.find("monitor")
mon.setTextColour(nil)
```

![](https://i.lemmmy.pw/9lmn.png)

Claims the error is in `peripheral.lua`.

PR #403, opened 2020-04-14
Add Json API

JakobDev:
JSON is used by almost every API in the web. Because of that, a built-in API for parsing JSON was often requested. Minecraft has already the JSON library gson. So I created a API which uses gson to parse JSON text and return it as table. I'm planning to add a function which creates JSON Text from a lua table.

### DIFF
--- a/src/main/java/dan200/computercraft/core/computer/ComputerExecutor.java
+++ b/src/main/java/dan200/computercraft/core/computer/ComputerExecutor.java
@@ -174,6 +174,7 @@ final class ComputerExecutor
         apis.add( new FSAPI( environment ) );
         apis.add( new PeripheralAPI( environment ) );
         apis.add( new OSAPI( environment ) );
+        apis.add( new JSONAPI() );
         if( ComputerCraft.http_enable ) apis.add( new HTTPAPI( environment ) );
 
         // Load in the externally registered APIs.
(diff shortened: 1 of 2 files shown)


src/main/java/dan200/computercraft/core/computer/ComputerExecutor.java
```diff
@@ -174,6 +174,7 @@
         apis.add( new FSAPI( environment ) );
         apis.add( new PeripheralAPI( environment ) );
         apis.add( new OSAPI( environment ) );
+        apis.add( new JSONAPI() );
```
lupus590, line 177: Is it possible to have this use require instead of APIs? I can see this conflicting with some older programs which test for and expect a different JSON API.

SquidDev: Not for Java ones. Honestly, my inclination has always been to define this in Lua - it's much more of a pain, but I like to keep the Java side of things as minimal as possible.

JakobDev: I think thats one of the rare cases that could be done better on Java side.  As I already said, we have already a lib for that in java. In Lua we have to write everything from scratch.Most of the JSON APIs for CC out there have bugs and did not work with every file. With gson, we can be sure that  everything works and we only need a few lines of code.

SquidDev: As far as I'm aware, the main with existing JSON APIs is that they don't handle all string escapes correctly, especially UTF-8 escapes. Once #363 is closed, that should be much easier to get right.

I'll have a fiddle over the weekend and see - I have a big deadline coming up this Friday, so a little preoccupied with that :).

SquidDev:
I added a Lua implementation #407, so going to close this. Sorry!

PR #407, opened 2020-04-18
Add textutils.unserialiseJSON

SquidDev:
This is relatively unoptimised and untested right now, but is a good starting point.

 - [x] Add `textutils.json_null`. This will be serialized into a literal   `null`. When deserialising, and `parse_null` is `true`, this will be returned instead of a `nil`.

 - [x] Add `textutils.unserializeJSON` (and `textutils.unserializeJSON`). This is a standard compliant JSON parser (hopefully).

 - [x] Passing in `nbt_style` to `textutils.unserializeJSON` will allow it to handle stringified NBT (no quotes around object keys, numeric suffices). We don't currently support byte/long/int arrays - something to add in a future commit.

 - [x] Add some tests for this. Some general ideas:
   - Round tripping random objects.
   - Some JSON test suites, [for instance this one of edge cases](https://github.com/nst/JSONTestSuite/tree/master/test_parsing).


JakobDev:
I#m not a fan of putting this in the textutils api. I think, it's better to have this in a own modul.

SquidDev:
@JakobDev I agree that it's somewhat ugly, but given that's where similar functions are (`textutils.serializeJSON`, `textutils.deserialize`) I think this is the best place for it.

JakobDev:
@JakobDev Mybae we could put the json code in a other API and make textutils.unserialiseJSON just call the function from the other API, like os.sleep() just call sleep(). unserialiseJSON has a lot of code and this would help to keep the textutils code clean.

SquidDev:
I'd rather avoid adding new APIs, because global polution is ewww. So the ideal solution would be to have a `cc.json` module, and have `textutils` load that, like we do for `cc.expect`.

However, then we have weird behaviour where users could have multiple references to `empty_json_array` (via textutils and a required module), and that _is_ going to be confusing.

Issue #351, opened 2020-01-24
Settings API Improvements


Lemmmy:
I know you're interested in rewriting the settings API, but here's a few things I noticed and features that may be useful to include in the future. I'm happy to write a PR for all of these.

## Settings Events
Events should be fired when settings are changed, saved, loaded, etc. The most useful event would be `settings_changed` that returns a table of settings keys that changed (optionally with old and new values, but this is a luxury). More specific events for each type of action could be included too, including `settings_save`, `settings_load` and `settings_clear`.

This would be useful for programs to automatically update their own variables or refresh their user interfaces when a setting is changed, e.g. from the shell command. A more specific example would be my street sign program, which is managed via the settings API - it would save the user having to restart the computer by catching a settings change event and redrawing the sign immediately.

## Settings Save
Settings tend to be saved in `/.settings` (for example, the BIOS loads [user settings](https://github.com/SquidDev-CC/CC-Tweaked/blob/0de5969ec135dfa3b7d96bbdc201964f07142da0/src/main/resources/assets/computercraft/lua/bios.lua#L989-L991) from there). However, the [`settings.load`](https://github.com/SquidDev-CC/CC-Tweaked/blob/0de5969ec135dfa3b7d96bbdc201964f07142da0/src/main/resources/assets/computercraft/lua/rom/apis/settings.lua#L57-L80) and [`settings.save`](https://github.com/SquidDev-CC/CC-Tweaked/blob/0de5969ec135dfa3b7d96bbdc201964f07142da0/src/main/resources/assets/computercraft/lua/rom/apis/settings.lua#L82-L93) functions require a path to be specified. While this functionality is definitely useful, I think it would support a more common use case to save and load to the `/.settings` file by default, and having the path be an optional argument. I am not sure if this was a deliberate design choice by dan200, or an oversight.

SquidDev:
> I know you're interested in rewriting the settings API

Just to clarify with this, I want to find a nice way to disambiguate between settings which have been defined (such as those set in the bios), and those which have actually been changed (i.e. set in `.settings`). Ideally `.settings` would only store _changed_ settings (rather than the entire state), and unset would reset to the default value (rather than clearing entirely).

PR #408, opened 2020-04-18
Some redesigning of the settings API

SquidDev:
This is based on some of the comments in #351, and some other things I've wanted to do. There's a couple of things going on in this change:

 - Add the ability to describe a specific setting. `settings.define("my_setting", { ... })` allows you to provide a default value and description about a specific setting. This has several benefits:

   - `/.settings` now only stores options which have been changed, as the bios registers default values, rather than stating their value exactly.
   - `settings.clear` and `settings.unset` reset something to its default value, rather than clearing it entirely.
   - The `set` program can now print whether an option has been changed, and its description:

     The fancy new `set` program

 - `settings.load`/`settings.save` now default to `/.settings`
 - `settings.set` and any other modifying function will fire a `settings_changed` event, with the arguments `name:string, new_value, old_value`.

---

I'd like to get some feedback from others before merging with this - there's definitely some design decisions I'm not happy with. As always, [the PR can be tried online using cloud-catcher](https://copy-cat.squiddev.cc/gh/branch/feature/new-settings/).

 - ~~`settings.add` is a bit of a confusing name. We probably also need a way to remove a setting's metadata, and really don't want to go with `settings.remove`.~~

 - Firing an event whenever a setting changes feels somewhat inefficient, especially when doing bulk actions such as loading from `/.settings`. On one hand, this is exactly what the event system is for, but given that 99% of programs won't care about this, I wonder if there's another way. Ideas?

   The original issue (#351) suggested batching all changes into one single event. It works, but it also feels inelegant in its own way. Ughr, I don't know :).


Lemmmy:
I wonder if `settings.define` and `settings.undefine` would be better names. This makes it clear that you need to explicitly define a setting for the computer to know about it's description.

I agree that batching things into a table doesn't feel very CC. Two alternative (but perhaps not any better) options come to mind, as food for thought:

- Register callback handlers instead of events (*very* not CC), e.g. `settings.onChanged(function(setting, oldValue, newValue) ... end)`
- A setting, off by default, for emitting settings events - `settings.notifyChanges` or `settings.emitChangeEvents` perhaps.

JakobDev:
We could also add a type check for settings

SquidDev:
Yeah, I did wonder about allowing some basic `type = "string"` field, but wasn't sure as some settings have dynamic types (such as `string` or `nil`). But even something simple would be a good start for now.

JakobDev:
You can use a table `type={"string","nil"}`. it would also be nice, if we have custom types e.g. color.

lupus590:
Tables like the one @JakobDev described is how my arg validator works, if you want to take bits of that then here's a [link](https://github.com/lupus590/CC-TreeFarm/blob/master/treeFarm/libs/utils/argValidationUtils.lua) to it.

SquidDev:
I'm not entirely with some of the error messages, especially the ones that the `set` program spits out, but this'll do for now.

PR #410, opened 2020-04-19
Remove / and \ at the start of shell.setDir()

JakobDev:
If you change the directory with the cd command, it will remove / or \ at the start of a path. With this PR, the prompt of the shell will look the same, if you use shell.setDir().

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
@@ -194,8 +194,7 @@ function shell.setDir( _sDir )
     if not fs.isDir( _sDir ) then
         error( "Not a directory", 2 )
     end
-    _sDir = fs.combine(_sDir, "")
-    sDir = _sDir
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/computercraft/lua/rom/programs/shell.lua
```diff
@@ -194,6 +194,10 @@ function shell.setDir( _sDir )
     if not fs.isDir( _sDir ) then
         error( "Not a directory", 2 )
     end
+    local sStartChar = string.sub( _sDir, 1, 1 )
```
SquidDev, line 197: Can we use `_sDir:sub(1, 1)` here too? Just so we're consistent with what's below.

SquidDev:
Hrmr, I wonder if it'd be better just to do `fs.combine(sDir, "")` - this way we normalise file paths, and eliminate redundancies such as `rom/../`.

Lemmmy:
Might have to make a test suite

SquidDev:
And merged. Thank you very much!

Issue #356, opened 2020-01-30
Compare performance of monitors in Minecraft 1.12 and 1.15.


SquidDev:
Due to rendering changes in Minecraft 1.15, monitors were rewritten to use VBOs rather than display lists. It would be good to compare performance between the two on a variety of GPUs and see if there are any regressions.

There's two situations we should compare here:
 - Static monitors - those which render a static image.
 - Constantly changing monitors. This doesn't need to be massive changes - a single character is sufficient to cause a re-draw.

SquidDev:
![A terribly FPS graph on 1.12.2](https://user-images.githubusercontent.com/4346137/74032606-48096800-49ac-11ea-9801-c4da9833e77b.png)

![A reasonable FPS graph on 1.15.1](https://user-images.githubusercontent.com/4346137/74032624-4e97df80-49ac-11ea-9c5d-5eb4bc424f24.png)


I was kinda looking forward to doing some profiling to see what could be improved, but I think this is pretty clear-cut.

PR #412, opened 2020-04-20
Backport 1.15's terminal rendering code

SquidDev:
This is a backport of 1.15's terminal rendering code with some further improvements. This duplicates a fair bit of code, and is much more efficient.

I expect the work done in #409 will supersede this, but that's unlikely to make its way into the next release so it's worth getting this in for now.

 - Refactor a lot of common terminal code into `FixedWithFontRenderer`. This shouldn't change any behaviour, but makes a lot of our terminal renderers (printed pages, terminals, monitors) a lot cleaner.

 - Terminal rendering is done using a single mode/vertex format. Rather than drawing an untextured quad for the background colours, we use an entirely white piece of the terminal font. This allows us to batch draws together more elegantly.

 - Some minor optimisations:
   - Skip rendering `"\0"` and `" "` characters. These characters occur pretty often, especially on blank monitors and, as the font is empty here, it is safe to skip them.
   - Batch together adjacent background cells of the same colour. Again, most terminals will have large runs of the same colour, so this is a worthwhile optimisation.
   - ~~Don't render monitors which are facing away from the player, and thus are invisible.~~

   These optimisations do mean that terminal performance is no longer consistent as "noisy" terminals will have worse performance. This is annoying, but still worthwhile.

 - Switch monitor rendering over to use VBOs. 

   We also add a config option to switch between rendering backends. By default we'll choose the best one compatible with your GPU, but there is a config option to switch between VBOS (reasonable performance) and display lists (bad).

### Numbers
Using the same setup as #356. I can't recall the exact monitor dimensions, but the image is 960x360. Here's a before and after:

Before

After

That big peak is due to the VBO uploading. It's still pretty bad, but is noticeably better than the current 1.12 and 1.15 code.

When this is more ready to merge, I'd also like to do some comparisons with smaller, but more frequently updating monitors. Graphics is not my forte, so I've no idea how it'll compare there.

If there's any rendering gurus out there, who have any additional comments, please do shout!


SquidDev:
@Lignum Don't know if you have any comments on the above? I don't _think_ there's anything unreasonable, but would appreciate a pair of eyes from someone who knows what they're doing if you've got time.

Lemmmy:
Minecraft doesn't support anything less than OpenGL 2.1. VBOs are core in that version, so there's no reason at all to even implement a direct or display list renderer

SquidDev:
@Lemmmy I believe that restriction was only introduced in 1.15? That's what [the changelog](https://minecraft.gamepedia.com/Java_Edition_1.15) seems to suggest at least.

src/main/java/dan200/computercraft/client/gui/FixedWidthFontRenderer.java
```diff
@@ -5,195 +5,318 @@
...
+     */
+    public static final VertexFormat POSITION_COLOR_TEX = new VertexFormat();
 
-    public static FixedWidthFontRenderer instance()
+    static
     {
-        if( instance != null ) return instance;
-        return instance = new FixedWidthFontRenderer();
+        POSITION_COLOR_TEX.addElement( DefaultVertexFormats.POSITION_3F );
+        POSITION_COLOR_TEX.addElement( DefaultVertexFormats.COLOR_4UB );
```
Lignum, line 38: Do we need the 4th colour component?

SquidDev: No, but `BufferBuilder` does! I want to stick with using this, as it's kinda unavoidable on 1.15 anyway, so might as well be consistent.

Lignum:
At this point, shouldn't we just rename `FixedWidthFontRenderer` to `TerminalRenderer`?

SquidDev:
_Technically_ it's also used for printed pages. Though yes, the bulk of the code is for terminals.

SquidDev:
Reminder for myself tomorrow: 
 - There's slight z-fighting at the edges of monitors. Not the end of the world, but worth seeing if it's fixable. 
 - Skip rendering monitors which are facing away from the player. So just bog standard backface culling, but earlier in the process.

Lignum:
That didn't go quite so well.

SquidDev:
Hah, that's what I get for writing code 6:30 in the morning. I'm fairly sure it's even more broken in 3rd person, as it uses the entity look vector rather than the actual render rotation.

I've got a fixed version locally, just need to test it and see what breaks.

SquidDev:
Heh, it's still broken. I've reverted it for now - I'll have another look when I've got some more time. It's an optimisation I imagine is worthwhile in something like SC's spawn, as you've got lots of monitors facing in other directions.

Either way, finally remembered to turn off vsync and FPS cap, and got some very happy screenshots:



There's definitely some more performance which could be eked out of monitors, but at this point I'm fairly comfortable that most of the remaining FPS issues aren't CC's fault.

SquidDev:
Yeah, I'm going to regret that. Hey ho :).

PR #418, opened 2020-04-21
Fix settings.save() sPath param.

Wojbie:


SquidDev:
Ughr, knew I should have written test cases.

PR #419, opened 2020-04-21
Add mouse support for the menu in edit and paint

### LINKED ISSUE #416: Mouse support for paint and edit menus
Currently, in order to access the menu bar of the `edit` and `paint` programs, you must press <kbd>Ctrl</kbd>. This is perfectly fine, but it's rather curious that you cannot use the mouse to either open or interact with the menu.

I think there's three bits of functionality we need here:
 - Clicking the menu bar will open it.
 - Clicking out of the menu bar will close it.
 - Clicking a button will close the menu bar, and act upon that action.

This is so obvious, I'm kinda mad I didn't do it years ago.

JakobDev:
Closes #416

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
@@ -61,7 +61,12 @@ if peripheral.find("printer") then
 end
 table.insert(tMenuItems, "Exit")
 
-local sStatus = "Press Ctrl or click here to access menu"
+local sStatus
+if term.isColour() then
+    sStatus = "Press Ctrl or click here to access menu"
+else
+    sStatus = "Press Ctrl to access menu"
+end
 if #sStatus > w - 5 then
     sStatus = "Press Ctrl for menu"
 end
@@ -730,9 +735,9 @@ while bRunning do
             if param == 1 then
                 -- Left click
                 if cy < h then
-                    local newY = math.min( math.max( scrollY + cy, 1 ), #tLines )
-                    local newX = math.min( math.max( scrollX + cx, 1 ), #tLines[newY] + 1 )
-                    setCursor( newX, newY )
+                    local newY = math.min(math.max(scrollY + cy, 1), #tLines)
+                    local newX = math.min(math.max(scrollX + cx, 1), #tLines[newY] + 1)
+                    setCursor(newX, newY)
                 else
                     bMenu = true
                     redrawMenu()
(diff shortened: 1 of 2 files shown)


lupus590:
I case you have not noticed, you might want to fix the linter errors.

SquidDev:
I haven't done any thorough testing yet, but fiddling around on mobile, this looks really good. Thank you!

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -61,7 +61,7 @@ if peripheral.find("printer") then
 end
 table.insert(tMenuItems, "Exit")
 
-local sStatus = "Press Ctrl to access menu"
+local sStatus = "Press Ctrl or click here to access menu"
```
SquidDev, line 64: On normal computers, this should probably still show "Press Ctrl to access menu" - we don't have mouse support there!

src/main/resources/assets/computercraft/lua/rom/programs/fun/advanced/paint.lua
```diff
@@ -261,27 +261,58 @@ local function accessMenu()
...
             if selection == k then
                 term.setTextColour(colours.yellow)
                 local ox = term.getCursorPos()
                 term.write("[" .. string.rep(" ", #v) .. "]")
                 term.setCursorPos(ox + 1, h)
+                nStartPos = ox + 1
                 term.setTextColour(colours.white)
                 term.write(v)
-                term.setCursorPos(term.getCursorPos() + 1, h)
+                nEndPos = term.getCursorPos()
```
SquidDev, line 304: Is this missing a `+ 1`? We currently have this, which is a little broken:

Slightly broken menu

That said, do we need to keep track of the positions? You don't appear to within edit, which does end up being a little shorter/cleaner.

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
Lemmmy: Double check how the message looks on advanced pocket computers, might be worth trying to shorten it

E.g. changing 'access' to 'open'

SquidDev: Pocket computers have a shorter message anyway:

https://github.com/SquidDev-CC/CC-Tweaked/blob/da419b24e701229eee3afa06315969c5d2a4f972/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua#L65-L67

Lemmmy: Perfect

PR #420, opened 2020-04-21
Update CraftOS Version to 1.9

JakobDev:
I think CCT should use Version 1.9 to spot the difference between the old CC with 1.8 and the newer CCT.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -98,7 +98,7 @@ end
 
 -- Install lua parts of the os api
 function os.version()
-    return "CraftOS 1.8"
+    return "CraftOS 1.9"
 end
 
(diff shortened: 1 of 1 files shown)


lupus590:
If the goal is to differentiate CC from CC:T why not call it `CraftOS: Tweaked`?

SquidDev:
I wonder if the best thing to do here would be just to use the current CC:T version number. So currently it'd be `CraftOS 1.86.2`, or just `CraftOS 1.86`. This is _kinda_ what CC did, though obviously not given its wacky versioning scheme.

So I guess something like `"CraftOS " .. (_HOST:match("ComputerCraft (%d+\.%d+)") or "unknown")`.

Lemmmy:
API changes are generally only introduced in new CC:T versions, so I think that's the best idea moving forward

PR #427, opened 2020-04-26
Fixed settings.load merge check

MCJack123:
`settings.load` had a line that was using a deleted variable to index a table, resulting in it always returning `nil`, and thus the check never went through. I'm surprised this got by the linter. I've fixed it to use `k` instead of `name`.


SquidDev:
Ouch, woops. Thanks for the fix.

Yeah, the linter doesn't currently check for unknown globals because CC:T has so many additional ones, and I haven't got round to writing a system for handling that yet. It's on the todo list!

SquidDev:
Sorry - ended up manually applying your patch as part of updating the linter's configuration. Turns out there were a few other of these, but hopefully this shouldn't happen again now.

PR #431, opened 2020-04-28
Improve syntax highlighting

Ocawesome101:
More different colors, inspired by vim's default Lua highlighting, and configurable too. Color settings should probably be checked, something I didn't bother to implement, or they could be removed.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
@@ -29,18 +29,22 @@ local tLines = {}
 local bRunning = true
 
 -- Colours
-local highlightColour, keywordColour, commentColour, textColour, bgColour, stringColour
+local highlightColour, booleanColour, keywordColour, functionColour, commentColour, textColour, bgColour, stringColour
 if term.isColour() then
-    bgColour = colours.black
-    textColour = colours.white
-    highlightColour = colours.yellow
-    keywordColour = colours.yellow
-    commentColour = colours.green
-    stringColour = colours.red
+    bgColour = settings.get("edit.backgroundColor") or colours.black
+    textColour = settings.get("edit.textColor") or colours.white
+    highlightColour = settings.get("edit.highlightColor") or colours.yellow
+    keywordColour = settings.get("edit.keywordColor") or colours.yellow
+    booleanColour = settings.get("edit.booleanColor") or colours.magenta
+    functionColour = settings.get("edit.functionColor") or colours.lightBlue
+    commentColour = settings.get("edit.commentColor") or colours.green
+    stringColour = settings.get("edit.stringColor") or colours.red
 else
     bgColour = colours.black
     textColour = colours.white
     highlightColour = colours.white
+    booleanColour = colours.magenta
+    functionColour = colours.white
     keywordColour = colours.white
     commentColour = colours.white
     stringColour = colours.white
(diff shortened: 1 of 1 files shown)


src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -164,10 +165,16 @@ local function writeHighlighted(sLine)
             tryWrite(sLine, "^\'\'", stringColour) or
             tryWrite(sLine, "^\'.-[^\\]\'", stringColour) or
             tryWrite(sLine, "^%[%[.-%]%]", stringColour) or
-            tryWrite(sLine, "^[%w_]+", function(match)
+            tryWrite(sLine, "^[%w_%.]+", function(match)
                 if tKeywords[match] then
```
Wojbie, line 169: You removed `tKeywords` table so this if will never trigger.

Ocawesome101:
Oh I did? Oops! Didn’t mean to do that

Lemmmy:
Does GitHub have a feature to disable web edits :facepalm:

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -29,18 +29,22 @@ local tLines = {}
...
-local highlightColour, keywordColour, commentColour, textColour, bgColour, stringColour
+local highlightColour, booleanColour, keywordColour, functionColour, commentColour, textColour, bgColour, stringColour
 if term.isColour() then
-    bgColour = colours.black
-    textColour = colours.white
-    highlightColour = colours.yellow
-    keywordColour = colours.yellow
-    commentColour = colours.green
-    stringColour = colours.red
+    bgColour = settings.get("edit.backgroundColor") or colours.black
```
SquidDev, line 34: - It's probably defining these settings at the bottom of `bios.lua` too.
 - Maybe make them strings instead? So one can define it as "red" instead.

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -29,18 +29,22 @@ local tLines = {}
...
+    textColour = settings.get("edit.textColor") or colours.white
+    highlightColour = settings.get("edit.highlightColor") or colours.yellow
+    keywordColour = settings.get("edit.keywordColor") or colours.yellow
+    booleanColour = settings.get("edit.booleanColor") or colours.magenta
```
SquidDev, line 38: Maybe "literal colour", so we can include `nil` too?

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -115,30 +119,27 @@ local function save(_sPath)
...
-    ["repeat"] = true,
-    ["return"] = true,
-    ["then"] = true,
-    ["true"] = true,
-    ["until"] = true,
-    ["while"] = true,
+    ["_G"] = true
 }
 
+local function addHighlights( tApi, sName )
```
SquidDev, line 127: I'd rather avoid highlighting global functions for now. I understand the appeal of highlighting builtins, but would rather not for now.

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
```diff
@@ -164,10 +165,16 @@ local function writeHighlighted(sLine)
...
             tryWrite(sLine, "^%[%[.-%]%]", stringColour) or
-            tryWrite(sLine, "^[%w_]+", function(match)
+            tryWrite(sLine, "^[%w_%.]+", function(match)
                 if tKeywords[match] then
                     return keywordColour
                 end
+                if tFunctions[match] then
+                    return functionColour
+                end
+                if match == "true" or match == "false" then
```
SquidDev, line 175: Maybe make `tKeywords` a mapping of string → colours, so one may do `return tKeywords[match] or textColour`, rather than this if-else chain.

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
Ocawesome101: Good idea. I’ll look into that.

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
Ocawesome101: Again, good idea.

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
Ocawesome101: Why not?

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
Ocawesome101: Sounds good.

src/main/resources/assets/computercraft/lua/rom/programs/edit.lua
SquidDev: Any such approach is always going to be a "best guess". This won't correctly handle cases like `local read` (which do show up), and at that point it'll just end up being confusing.

Ocawesome101:
Fair. Maybe just add highlights for the standard Lua APIs like most editors
do?

Issue #433, opened 2020-04-29
Functions/events to make more responsive UIs


znepb:
### mouse_move (mouse_hover)
Basically, detects when the mouse is moved between CC pixels. Of course this would only work on advanced computers. What it returns would probably be similar to mouse_click. If the mouse were to move outside of the screen area, the event wouldn't be called until it goes inside the screen area again.

### term.getMousePosition()
This function would return the current mouse position on the screen. If it isn't within the screen's area, it would return nil.

## Why this would be useful
Currently, UIs are quite boring in CC. Adding these functions and events would give developers the opportunity to make better looking UIs.

dmarcuse:
This has come up quite a few times in the past - here's some of the common counterarguments for consideration.

- CC is not exactly fast, especially on larger servers - adding more events that are updated so frequently would just add to this
- How would it work with multiple players accessing a computer in multiplayer?
- How would CC programs detect the mouse exiting the screen or the interface being closed?

znepb:
To solve the second issue, the last person who clicked would send the mouse move events. The third issue, possibly a mouse_exit event?

SquidDev:
Having sat on this for a couple of months, I'm still not happy with this and so am going to close for now. It rather irks me - I can't place why I don't like the idea of it, it just _feels_ not CC enough, and I really don't want to do something I'll later end up regretting.

Lemmmy:
thank god for that

PR #434, opened 2020-04-30
Implement mouse_move event

Lemmmy:
This implements the first half of #433. I'm not actually interested in implementing `term.getMousePosition()` from that issue out of fear of issues with multiple players, and having to implement some kind of polling system for that. Such a feature may be better suited for the `window` API.

When the cursor is moved within the terminal on any client, a `mouse_move` event is issued with the `xPos` and `yPos` parameters (in characters, as usual). If the cursor moves from inside to outside the terminal's region, the position `-1, -1` is sent. This allows the program to deactivate any element that is currently hovered. [This program](https://pastebin.com/nvttbria) demonstrates that by turning the screen red (with throttle disabled):

mouse_move

To solve the issue with bandwidth, there is a config option for throttling, `mouse_move_throttle`. By default, this is 200ms, and is implemented as a leading-edge debounce. 200ms will be fine for servers, as I wouldn't recommend anyone build programs with a 'cursor' anyway (due to issues raised in #409). Instead, this is more useful for things like buttons. The feature can be disabled entirely by setting the throttle time in the config to `-1`.

It's worth noting that `mouse_drag` does not have a throttle, but those events are less likely to happen compared to `mouse_move` events. It's simple to copy the code to implement a drag throttle if needed, though.

Multiple players in a computer works no differently to `mouse_drag`, that is - the events are simply multiplexed with no distinction between players. I think this will be _fine_, given that the events are only sent if the mouse actually moves.

One more concern is throttle configuration. I've simply used the same configuration as everything else at the moment. In my understanding of Forge, mod configs are synchronised to the player when they connect, but I don't know if there's anything that stops the player from reconfiguring it client-side. It may be worth reviewing this.

I've tested this on all kinds of computers (Computer, Pocket Computer, Turtle, Neural Interface). Basic computers do not transmit mouse events at all, so this works perfectly.

# TO-DO
- [x] Fix multishell/window API support
- [ ] Update CCEmuX
- [ ] Port to 1.14+

# Thoughts
- I'm not 100% sure about sending -1, -1 as the position for the mouse being outside the terminal. Perhaps `nil, nil` would be better, but it could break programs that perform blanket handling of mouse events? The other solution is another event, e.g. `lose_focus`.
- The multishell code is **ugly**, as a result of not sending the additional argument with the mouse event. Perhaps it should be sent, but always as `nil` or `0`, just to be fully compatible with all the other mouse events?

# Thanks
- [znepb](https://github.com/znepb) for #433 
- [Anavrins](https://github.com/Anavrins) for the French translation
- [Abigail](https://github.com/plt-hokusai) for the Portuguese translation
- [Repflez](https://github.com/Repflez) for the Spanish translation
- [1lann](https://github.com/1lann) for the Chinese translation

### DIFF
--- a/src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
+++ b/src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
@@ -34,8 +34,6 @@ public class WidgetTerminal extends Widget
 
     private int m_lastMouseX = -1;
     private int m_lastMouseY = -1;
-    private long m_lastMouseMove = -1;
-    private boolean m_mouseMoved = false;
 
     private boolean m_focus = false;
     private boolean m_allowFocusLoss = true;
@@ -215,9 +213,10 @@ public class WidgetTerminal extends Widget
     public void handleMouseInput( int mouseX, int mouseY )
     {
         IComputer computer = m_computer.getComputer();
+        if ( computer == null || !computer.isColour() ) return;
+
         if( mouseX >= getXPosition() && mouseX < getXPosition() + getWidth() &&
-            mouseY >= getYPosition() && mouseY < getYPosition() + getHeight() &&
-            computer != null && computer.isColour() )
+            mouseY >= getYPosition() && mouseY < getYPosition() + getHeight() )
         {
             Terminal term = computer.getTerminal();
             if( term != null )
@@ -229,12 +228,12 @@ public class WidgetTerminal extends Widget
 
                 handleMouseClick( computer, charX, charY );
                 handleMouseWheel( computer, charX, charY );
-                handleMouseMove( charX, charY );
+                handleMouseMove( computer, charX, charY );
             }
         }
         else // The mouse has moved out of the terminal, send a -1, -1 mouse_move event
         {
-            handleMouseMove( -1, -1 );
+            handleMouseMove( computer, -1, -1 );
         }
     }
 
@@ -275,15 +274,16 @@ public class WidgetTerminal extends Widget
         }
     }
 
-    private void handleMouseMove( int charX, int charY )
+    private void handleMouseMove( IComputer computer, int charX, int charY )
     {
+        // Avoid sending mouse movement events if the cursor is under the same character.
+        // Note that clients can also use the mouseMoveThrottle config options to disable
+        // sending their mouse movements to a server entirely.
         if( ComputerCraft.mouseMoveThrottle >= 0 && (m_lastMouseX != charX || m_lastMouseY != charY) )
         {
-            // Simply update the last mouse move location, the work is delegated to the update handler,
-            // which will deal with throttling too!
+            computer.mouseMove( charX, charY );
             m_lastMouseX = charX;
             m_lastMouseY = charY;
-            m_mouseMoved = true;
         }
     }
 
@@ -348,22 +348,6 @@ public class WidgetTerminal extends Widget
             m_terminateTimer = 0.0f;
             m_rebootTimer = 0.0f;
             m_shutdownTimer = 0.0f;
-        }
-
-        // Handle mouse movement
-        if( m_mouseMoved && System.currentTimeMillis() - ComputerCraft.mouseMoveThrottle > m_lastMouseMove )
-        {
-            IComputer computer = m_computer.getComputer();
-            if( computer != null )
-            {
-                computer.mouseMove(
-                    m_lastMouseX == -1 ? -1 : m_lastMouseX + 1,
-                    m_lastMouseY == -1 ? -1 : m_lastMouseY + 1
-                );
-            }
-
-            m_mouseMoved = false;
-            m_lastMouseMove = System.currentTimeMillis();
         }
     }
 
--- a/src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
+++ b/src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
@@ -226,51 +226,38 @@ public class WidgetTerminal extends Widget
                 charX = Math.min( Math.max( charX, 0 ), term.getWidth() - 1 );
                 charY = Math.min( Math.max( charY, 0 ), term.getHeight() - 1 );
 
-                handleMouseClick( computer, charX, charY );
-                handleMouseWheel( computer, charX, charY );
+                if( m_lastClickButton >= 0 && !Mouse.isButtonDown( m_lastClickButton ) )
+                {
+                    if( m_focus ) computer.mouseUp( m_lastClickButton + 1, charX + 1, charY + 1 );
+                    m_lastClickButton = -1;
+                }
+
+                int wheelChange = Mouse.getEventDWheel();
+                if( m_focus )
+                {
+                    if( wheelChange < 0 )
+                    {
+                        computer.mouseScroll( 1, charX + 1, charY + 1 );
+                    }
+                    else if( wheelChange > 0 )
+                    {
+                        computer.mouseScroll( -1, charX + 1, charY + 1 );
+                    }
+
+                    if( m_lastClickButton >= 0 && (charX != m_lastClickX || charY != m_lastClickY) )
+                    {
+                        computer.mouseDrag( m_lastClickButton + 1, charX + 1, charY + 1 );
+                        m_lastClickX = charX;
+                        m_lastClickY = charY;
+                    }
+                }
+
                 handleMouseMove( computer, charX, charY );
             }
         }
         else // The mouse has moved out of the terminal, send a -1, -1 mouse_move event
         {
             handleMouseMove( computer, -1, -1 );
-        }
-    }
-
-    private void handleMouseClick( IComputer computer, int charX, int charY )
-    {
-        if( m_lastClickButton >= 0 && !Mouse.isButtonDown( m_lastClickButton ) )
-        {
-            if( m_focus ) computer.mouseUp( m_lastClickButton + 1, charX + 1, charY + 1 );
-            m_lastClickButton = -1;
-        }
-    }
-
-    private void handleMouseWheel( IComputer computer, int charX, int charY )
-    {
-        int wheelChange = Mouse.getEventDWheel();
-        if( wheelChange == 0 && m_lastClickButton == -1 )
-        {
-            return;
-        }
-
-        if( m_focus )
-        {
-            if( wheelChange < 0 )
-            {
-                computer.mouseScroll( 1, charX + 1, charY + 1 );
-            }
-            else if( wheelChange > 0 )
-            {
-                computer.mouseScroll( -1, charX + 1, charY + 1 );
-            }
-
-            if( m_lastClickButton >= 0 && (charX != m_lastClickX || charY != m_lastClickY) )
-            {
-                computer.mouseDrag( m_lastClickButton + 1, charX + 1, charY + 1 );
-                m_lastClickX = charX;
-                m_lastClickY = charY;
-            }
         }
     }
 
--- a/src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
+++ b/src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
@@ -5,6 +5,7 @@
  */
 package dan200.computercraft.client.gui.widgets;
 
+import dan200.computercraft.ComputerCraft;
 import dan200.computercraft.client.gui.FixedWidthFontRenderer;
 import dan200.computercraft.core.terminal.Terminal;
 import dan200.computercraft.shared.computer.core.IComputer;
@@ -30,6 +31,11 @@ public class WidgetTerminal extends Widget
     private int m_lastClickButton = -1;
     private int m_lastClickX = -1;
     private int m_lastClickY = -1;
+
+    private int m_lastMouseX = -1;
+    private int m_lastMouseY = -1;
+    private long m_lastMouseMove = -1;
+    private boolean m_mouseMoved = false;
 
     private boolean m_focus = false;
     private boolean m_allowFocusLoss = true;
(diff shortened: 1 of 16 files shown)


plt-hokusai:
If anyone really _wants_ a `getMousePos` it's easy enough to Do It Yourself™, in true CC fashion:

```lua
local last_x, last_y
function term.getMousePos()
  return last_x, last_y
end
parallel.waitForAny(function()
  -- your code
end, function()
  while true do
    local _, x, y = os.pullEvent('mouse_move')
    last_x, last_y = x, y
  end
end)
```

znepb:
Nice! Works like a charm. Thanks for implementing this. I made a test program with the build Lemmmy provided (here)[https://cdn.discordapp.com/attachments/530071845181849630/705220308881244170/cc-tweaked-1.12.2-SNAPSHOT-mousehover.jar]. Here's the link on pastebin: https://pastebin.com/lelkewh8

src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
```diff
@@ -221,37 +227,63 @@ public void handleMouseInput( int mouseX, int mouseY )
...
-                    m_lastClickButton = -1;
-                }
+                handleMouseClick( computer, charX, charY );
+                handleMouseWheel( computer, charX, charY );
+                handleMouseMove( charX, charY );
+            }
+        }
+        else // The mouse has moved out of the terminal, send a -1, -1 mouse_move event
+        {
+            handleMouseMove( -1, -1 );
```
plt-hokusai, line 237: Are we sure this is a good idea, instead of sending _no_ position (so that the code explodes with, say, `attempt to perform arithmetic on nil`) or a different event for losing focus entirely?

Lemmmy: My biggest concern with sending `nil` is backwards compatibility with programs that perform blanket handling of all mouse events, but `-1` could have a similar outcome.

Lemmmy:
If anyone wants to test it out, grab the latest artifact [here](https://github.com/SquidDev-CC/CC-Tweaked/suites/640824066/artifacts/5355119).

src/main/resources/assets/computercraft/lua/rom/help/events.txt
```diff
@@ -11,5 +11,6 @@ Events only on advanced computers:
 "mouse_drag" when a user moves the mouse when held. Arguments are button, xPos, yPos.
 "mouse_up" when a user releases the mouse button. Arguments are button, xPos, yPos.
 "mouse_scroll" when a user uses the scrollwheel on the mouse. Arguments are direction, xPos, yPos.
+"mouse_move" when a user moves the mouse within the terminal. Arguments xPos, yPos. xPos and yPos will be -1, -1 when the mouse is moved outside of the terminal.
```
SquidDev, line 14: Hahah, this is the first time these docs have been updated in a very long time :D:.

src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
SquidDev: I'd be inclined to send nil<sup>1</sup>, or have an entirely separate event. I don't know, both are gross. If someone's doing `event:find("^mouse_")` then we're already breaking their code due to missing the "button" argument.

Also, does this send the event every frame, or just the first one it moves outside?

1: Well, send 0 or -1, but convert it to `nil` on the server.

src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
```diff
@@ -317,6 +349,22 @@ public void update()
...
+            if( computer != null )
+            {
+                computer.mouseMove(
+                    m_lastMouseX == -1 ? -1 : m_lastMouseX + 1,
+                    m_lastMouseY == -1 ? -1 : m_lastMouseY + 1
+                );
+            }
+
+            m_mouseMoved = false;
+            m_lastMouseMove = System.currentTimeMillis();
```
SquidDev, line 366: I'd be inclined to put this processing on the server - inside `InputHandler` most likely. The packet is 3 bytes (plus discriminator) so the network traffic is not really my worry - event spam is more problematic.

One can probably drop the delay to something smaller - 50ms maybe? Goodness knows - try it out on SC and see what's tolerable.

Lemmmy: WRT the -1 event, I'm not sure if you saw my 'Thoughts' up the top, but what do you think about just making it compatible with the other events? So just always send `1` as the first (e.g. button) argument.

Lemmmy: > Also, does this send the event every frame, or just the first one it moves outside?

It just sends the first time it moves outside, as m_lastMouse[XY] get set to `-1`, and that value is remembered.

Lemmmy: > One can probably drop the delay to something smaller - 50ms maybe? Goodness knows - try it out on SC and see what's tolerable.

My concern isn't the mouse packets, it's screen redraws. That said, we can drop the default to 50 (0 also works, by the way), and just increase SwitchCraft as needed.

Lemmmy: > I'd be inclined to put this processing on the server - inside `InputHandler` most likely. The packet is 3 bytes (plus discriminator) so the network traffic is not really my worry - event spam is more problematic.

One small thing - not exactly sure how to fit it in InputHandler because it's an interface and we need a variable

Lemmmy:
That lint failure definitely wasn't me. I've made the requested changes:

- Mouse throttling has been moved to the server, specifically to the InputState class.
- The default mouse_move throttle time is now 50ms.
- When the cursor is moved outside of the terminal, or the container is closed, `nil` is now sent instead of `-1`.S
- Added an extra 'dummy' argument to ensure compatibility with programs performing blanket handling of mouse events.

With these changes, the effect is now `mouse_move 1 x y`, making it compatible with all the other mouse events. The `x` and `y` coordinates, however, are now `nil`, which is **not** compatible with the other mouse events entirely. For example, I had to add a special case to multishell's handler for the `y` position.

Side-note: I'm not sure the GitHub diff is showing up correctly. It looks like [this](https://i.lemmmy.pw/sCIN.png), but the file on my end looks like [this](https://i.lemmmy.pw/BSUD.png).

SquidDev:
Ahh, that lint failure is my fault. Looks like an illuaminate update caught some more stuff :).

src/main/resources/assets/computercraft/lua/rom/programs/advanced/multishell.lua
```diff
@@ -386,7 +386,12 @@ while #tProcesses > 0 do
             end
         elseif not (bShowMenu and y == 1) then
             -- Passthrough to current process
-            resumeProcess(nCurrentProcess, sEvent, p1, x, bShowMenu and y - 1 or y)
+            if x == nil or y == nil then -- Handle mouse_move nil events
+				resumeProcess(nCurrentProcess, sEvent, p1, nil, nil)
+            else
+				resumeProcess(nCurrentProcess, sEvent, p1, x, bShowMenu and y - 1 or y)
+			end
```
Lignum, line 393: Indent

src/main/java/dan200/computercraft/shared/computer/core/InputState.java
```diff
@@ -93,6 +99,44 @@ public void mouseScroll( int direction, int x, int y )
...
+            if( computer != null )
+            {
+                computer.mouseMove(
+                    lastMouseMoveX == -1 ? -1 : lastMouseMoveX + 1,
+                    lastMouseMoveY == -1 ? -1 : lastMouseMoveY + 1
+                );
+            }
+
+            mouseMoved = false;
+            lastMouseMove = System.currentTimeMillis();
```
SquidDev, line 136: Sorry, meant to mention this yesterday and entirely forgot:

We should probably use some monotonic clock here. We don't want people unable to move the mouse because their system clock changes! Either `System.nanoTime` or the current tick<sup>1</sup> would be fine.

1: Obviously then it depends on tps, but I can live with that.

src/main/java/dan200/computercraft/client/gui/widgets/WidgetTerminal.java
```diff
@@ -221,37 +226,64 @@ public void handleMouseInput( int mouseX, int mouseY )
...
-                        computer.mouseScroll( 1, charX + 1, charY + 1 );
-                    }
-                    else if( wheelChange > 0 )
-                    {
-                        computer.mouseScroll( -1, charX + 1, charY + 1 );
-                    }
+    private void handleMouseWheel( IComputer computer, int charX, int charY )
+    {
+        int wheelChange = Mouse.getEventDWheel();
+        if( wheelChange == 0 && m_lastClickButton == -1 )
```
SquidDev, line 252: Instead of moving these into separate functions, I'd keep it as is and just delete this if block. Yes, `handleMouseInput` will be an ugly behemoth, but it is already.

src/main/resources/assets/computercraft/lua/rom/help/events.txt
```diff
@@ -11,5 +11,6 @@ Events only on advanced computers:
 "mouse_drag" when a user moves the mouse when held. Arguments are button, xPos, yPos.
 "mouse_up" when a user releases the mouse button. Arguments are button, xPos, yPos.
 "mouse_scroll" when a user uses the scrollwheel on the mouse. Arguments are direction, xPos, yPos.
+"mouse_move" when a user moves the mouse within the terminal. Arguments xPos, yPos. xPos and yPos will be nil, nil when the mouse leaves the terminal.
```
SquidDev, line 14: This should probably mention the first argument is 1 now.

osmarks:
I've been asked to provide feedback on this. Personally I'm not really a fan - it seems as if it would spam computers with events and thus worsen performance, have problems with multiple simultaneous users (I think the suggestion was to send events from the last player who clicked, but that would be really unintuitive), and it doesn't seem to provide much value. Hover-based UIs could be very annoying to use. Also, it wouldn't translate neatly to monitors, so some programs would be unable to work on those.

Luca0208:
I'd agree on osmarks with the event spam, especially because most of the time the mouse_move events will just be ignored. I would suggest adding a function which enables/disables the mouse_move events, kind of like you need to use modem.open to receive modem_message events.

I'd put the function either in term or in os. If it is put in term, monitors could implement the method by just throwing an error informing the user, that mouse_move events don't work with monitors. Of course that would mean that there would be programs which work on computers, but crash when run on monitors, although this would only affect new programs, so I'm not sure how big of a problem that would be.

Lemmmy:
I totally agree with the event spam. As a server admin, this was my biggest concern, so I kept this in mind the entire time while implementing. What do you think about a toggle @SquidDev ?

For clarification, the events are throttled server-side and on default configuration won't be sent more than once per 50ms (this was originally 100-200 before Squid requested it to be lowered). On SwitchCraft, this will not be any less than 200ms, possibly higher. That means that no more than 5 mouse_move events will be sent per second. For comparison, mouse_drag events are sent more than once per tick.

The biggest concern with the event spam is people who have a loop like this:

```lua
while true do
  local event = os.pullEvent()
  -- ...
  redraw()  
end
```

This will result in significantly more unnecessary redraws. **However, I am not actually concerned about this** as an admin. This is why we have profilers (`/computercraft track`) and [analytics](https://analytics.switchcraft.pw/), and as @osmarks and @Luca0208 know, we are no stranger to telling people to fix their bad program design. If anything, this will make it easier to weed out bad programs.

Additionally, on SwitchCraft, most of the terminal drawing overhead actually comes from monitors that are updating at regular intervals 24/7. It's not really that common that a player is moving their mouse around in a terminal. If they're in the shell or `edit`, both hands are on the keyboard and the mouse is not moving. The mouse only ever moves if they are using a GUI, and it doesn't move as frequently as you think. And actually, in my testing, key events are more of an event spam problem (`key`, `key_up`, `char`) than `mouse_move`.

SquidDev:
Thanks to the joys of the aforementioned analytics, we can actually crunch the numbers!

Some statistics about events

This is a plot of the number of events computers process each second on SC (in the last hour).
 - Computers _on average_ process 1.4 events per second.
 - 95% of computers process 12 or less.
 - Some computers may process up to 100-150/s. Fun facts, as I'm writing this there's one computer handling [1k events in a second](https://user-images.githubusercontent.com/4346137/80970727-8a0f4200-8e13-11ea-9d85-739ace087598.png).

Given that most computers are not being interacted with, and events are being debounced/limited to 20/s (or 5/s on SC), I don't think the additional event spam is really something to be worried about.<sup>1</sup>

I think it's worth referring to #408, where we had a similar discussion about event spam. I think it's worth bearing in mind that CC is designed around an event system - if it can't handle a large number of them, it's doing something wrong.

Obviously this is something we'll monitor on SC, and this can be throttled/disabled if it becomes a problem.

---

As far as monitors go, while true I don't think it's a major issue. There's no counterpart for `mouse_up` or `mouse_scroll` events afterall.

1: As Lem says, a decent typing speed sends a lot of more events. There's 3 for every character!

dmarcuse:
As far as out-of-bounds events go, I think it'd make more sense to non-numeric values (i.e. `nil`) or even better, a separate event. This is a bit more consistent with existing input events (e.g. `mouse_up` and `key_up` are separate from `mouse_click` and `key`) and would make the API easier to understand and use IMO.

SquidDev:
Maybe `mouse_leave`? Possibly `mouse_exit`, but I think that's less clear.

It occurs to me we should probably fire that in `multishell`, when the cursor enters the menu bar at the top too.

znepb:
As with `mouse_leave`, I think if would be best if `term_focuslost` was implemented instead. This would solve the problem with programs checking for "mouse_" in the event string, as having `mouse_exit` would cause problems (fuzzy event checking, as said before).

SquidDev:
I'm incredibly unfussed about people doing `event:find("^mouse_")`. I don't think it's reasonable to worry about every possible broken thing people may be doing.

znepb:
> I'm incredibly unfussed about people doing `event:find("^mouse_")`. I don't think it's reasonable to worry about every possible broken thing people may be doing.

I used this in [zwm](https://github.com/znepb-cc/zwm), and when the `mouse_move` update was pushed to SC, it broke it. So I'm rewriting all the event forwarding 😛 

As with `term_focuslost`, this would break older programs that are no longer maintained. I'll try to find some examples of this that use fuzzy mouse searching (or whatever) and give some examples here. As for now, returning `nil, nil` works.

SquidDev:
The reality is that we can't ensure forwards compatibility for bad code. Obviously I don't want to break anything, but I also don't want to make things wildly inconsistent with the rest of CC for the sake of dealing with every eventuality.

SquidDev:
See my comments in #433. Sorry @Lemmmy!

PR #436, opened 2020-05-02
Return the peripheral name when wrapping

Lemmmy:
This is an incredibly simple, extremely useful, but potentially controversial change. This returns the name of the peripheral when wrapping it. This is useful when performing `peripheral.find()`, especially in bulk, to get the side name of the peripheral that was wrapped, for example:

```lua
local disks = table.pack(peripheral.find("drive"))

print("Found", disks.n, "disk drives."))

for _, disk in pairs(disks) do
    print("Disk drive", disk.name, disk.isDiskPresent() and "has a disk." or "does not have a disk.")
end
```

This implementation is extremely naïve, so this PR is more of a request for comments than anything. It shadows any peripheral methods called `name`, and while I'm not aware of any, this could potentially be bizarre or breaking behaviour in an extreme edge case. As such, it could perhaps be moved to run before the methods are populated when wrapping, or named to something else, such as `peripheralName`. Curious to hear your thoughts.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/apis/peripheral.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/peripheral.lua
@@ -154,6 +154,7 @@ function wrap(name)
             return peripheral.call(name, method, ...)
         end
     end
+    result.name = name
     return result
 end
 
--- a/src/main/resources/assets/computercraft/lua/rom/apis/peripheral.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/peripheral.lua
@@ -173,7 +173,7 @@ function wrap(name)
     local result = setmetatable({}, {
         __name = "peripheral",
         name = name,
-        type = peripheral.getType(name)
+        type = peripheral.getType(name),
     })
     for _, method in ipairs(methods) do
         result[method] = function(...)
--- a/src/main/resources/assets/computercraft/lua/rom/apis/peripheral.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/peripheral.lua
@@ -113,7 +113,7 @@ function getName(peripheral)
     expect(1, peripheral, "table")
     local mt = getmetatable(peripheral)
     if not mt or mt.__name ~= "peripheral" or type(mt.name) ~= "string" then
-        error("Could not get peripheral name")
+        error("bad argument #1 (table is not a peripheral)", 2)
     end
     return mt.name
 end
@@ -167,7 +167,6 @@ function wrap(name)
             return peripheral.call(name, method, ...)
         end
     end
-    result.name = name
     return result
 end
 
--- a/src/main/resources/assets/computercraft/lua/rom/apis/peripheral.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/apis/peripheral.lua
@@ -64,24 +64,33 @@ function isPresent(name)
     return false
 end
 
---- Get the type of the peripheral with the given name.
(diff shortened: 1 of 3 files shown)


SquidDev:
One possible alternative would be to store this information in the metatable. So something like:

```lua
function wrap(name)
  -- ...
  local result = setmetatable({}, { __name = "peripheral", name = name })
  -- ..
end

function getName(peripheral)
  expect(1, peripheral, "table")
  local mt = getmetatable(peripheral)
  if not mt or mt.__name ~= "peripheral" or type(mt.name) ~= "string" then
    error("idk")
  end
  return mt.name
end
```

This way you avoid any risk of shadowing things which nice. Though obviously the implementation is a bit more ugly. ¯\\\_(ツ)\_/¯

Lemmmy:
That's actually not bad at all. So this essentially implements `peripheral.getName(wrappedPeripheral)`, where wrappedPeripheral is an instance of `peripheral.wrap(...)`?

SquidDev:
Yep, exactly.

src/main/resources/assets/computercraft/lua/rom/apis/peripheral.lua
```diff
@@ -148,12 +161,13 @@ function wrap(name)
...
     end
 
-    local result = {}
+    local result = setmetatable({}, { __name = "peripheral", name = name })
     for _, method in ipairs(methods) do
         result[method] = function(...)
             return peripheral.call(name, method, ...)
         end
     end
+    result.name = name
```
SquidDev, line 170: I think it's safe to remove this line now?

SquidDev:
```suggestion
```

src/main/resources/assets/computercraft/lua/rom/apis/peripheral.lua
```diff
@@ -105,6 +105,19 @@ function getMethods(name)
...
 
+--- Get the name of a peripheral wrapped with @{peripheral.wrap}.
+--
+-- @tparam table peripheral The peripheral to get the name of.
+-- @treturn string The name of the given peripheral.
+function getName(peripheral)
+    expect(1, peripheral, "table")
+    local mt = getmetatable(peripheral)
+    if not mt or mt.__name ~= "peripheral" or type(mt.name) ~= "string" then
+        error("Could not get peripheral name")
```
SquidDev, line 116:
```suggestion
        error("bad argument #1 (table is not a peripheral)", 2)
```

src/main/resources/assets/computercraft/lua/rom/apis/peripheral.lua
Lemmmy: Whoops!

Lemmmy:
One more thing that could be done while we're here is to change getType to accept either a string or a peripheral instance - thoughts?

SquidDev:
Yeah, I did wonder about that. It's probably a good idea for completeness - not sure whether we should store the type, or just recompute it.

Sorry to be a pain, but would you also be able to add a test to `peripheral_sepc.lua`. Something like:

```lua
it_modem("can get the name of a wrapped peripehral", function()
  expect(peripheral.getName(peripheral.wrap("top")):eq("top")
end)
```

Lemmmy:
I thought that too, and figured recomputing would be fine since wrap() and find() aren't called _that_ often so the performance loss is negligible.

I also forgot to add it to the help pages, I'll do that too.

Lemmmy:
Hopefully I got the tests right - I assume they were run by `./gradlew test`. Everything seems to be working correctly:

Issue #142, opened 2019-03-16
[


BigBang1112:
## Useful information to include:
 - Minecraft version
 - CC: Tweaked version
 - Detailed reproduction steps: sometimes I can spot a bug pretty easily, but often it's much more obscure. The more information I have to help reproduce it, the quicker it'll get fixed.

PR #437, opened 2020-05-02
Preserve computer ids on unlabelled computers

SquidDev:
Breaking an unlabelled computer/turtle will now preserve all information about the computer, rather than resetting it.

In order to make things a little clearer, we display the id of unlabelled computers in the tooltip:

Computer ID: 142

I'm not entirely happy with the tooltip - there's an argument it should be done in the label instead (i.e. `Advanced Computer #142`). I don't know which one is clearer.


Lemmmy:
I think I like the way the tooltip is done right now - including the ID in the actual label could cause some confusion/clutter with the advanced tooltip from F3+H.

Issue #90, opened 2018-12-24
Unable to mount file


LemADEC:
- Minecraft version 1.12.2
 - CC: Tweaked version 1.80pr1.12

My WarpDrive mod attach a startup file to autoboot the computer when attached to certain blocks.
With CC-Tweaked, it appears file attachment is no longer supported:
```
[18:54:20] [ComputerCraft-Computer-Runner-0/ERROR] [computercraft]: Could not load mount from mod jar
java.nio.file.NotDirectoryException: assets/warpdrive/lua.ComputerCraft/common/updater
>       at com.sun.nio.zipfs.ZipDirectoryStream.<init>(ZipDirectoryStream.java:73) ~[zipfs.jar:1.8.0_191]
>       at com.sun.nio.zipfs.ZipPath.newDirectoryStream(ZipPath.java:651) ~[zipfs.jar:1.8.0_191]
>       at com.sun.nio.zipfs.ZipFileSystemProvider.newDirectoryStream(ZipFileSystemProvider.java:262) ~[zipfs.jar:1.8.0_191]
>       at java.nio.file.Files.newDirectoryStream(Files.java:457) ~[?:1.8.0_191]
>       at java.nio.file.Files.list(Files.java:3451) ~[?:1.8.0_191]
>       at dan200.computercraft.core.filesystem.FileSystemMount.<init>(FileSystemMount.java:29) ~[FileSystemMount.class:?]
>       at dan200.computercraft.ComputerCraft.createResourceMount(ComputerCraft.java:929) [ComputerCraft.class:?]
>       at sun.reflect.GeneratedMethodAccessor44.invoke(Unknown Source) ~[?:?]
>       at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_191]
>       at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_191]
>       at dan200.computercraft.api.ComputerCraftAPI.createResourceMount(ComputerCraftAPI.java:152) [ComputerCraftAPI.class:?]
>       at cr0s.warpdrive.block.TileEntityAbstractInterfaced.attach(TileEntityAbstractInterfaced.java:389) [TileEntityAbstractInterfaced.class:?]
>       at pl.asie.computronics.api.multiperipheral.WrappedMultiPeripheral.attach(WrappedMultiPeripheral.java:43) [WrappedMultiPeripheral.class:?]
[...]

PR #440, opened 2020-05-03
Improvements to the various file system programs (rm, mv, rename)

lupus590:
instead of using the pcall error


lupus590:
remind me, how do I run the rom tests locally?

SquidDev:
`./gradelw test` will run the test suite.

You can also point CCEmuX towards the rom and test suite: `java -jar CCEmux-cct.jar -a …/CC-Tweaked/src/main/resources -c …/CC-Tweaked/src/test/resources` and then do `test-rom/myfly test-rom/spec` to run it.

lupus590:
Do you want the `isMountPoint` function to go into the FS API?

src/main/resources/assets/computercraft/lua/rom/programs/delete.lua
```diff
@@ -8,7 +8,7 @@ end
 local function isMountPoint(path)
     if not fs.isDir(path) then
         return false
-    elseif fs.getDrive(path) == fs.getDrive(fs.combine(path, "..")) then
+    elseif fs.combine(path, "..") == ".." or fs.getDrive(path) == fs.getDrive(fs.combine(path, "..")) then -- the == ".." check prevents errors when path is "/"
```
lupus590, line 11: I would like to remove the need for this comment but I'm not sure how clear this code is on it's own

lupus590:
It's been a couple of days @SquidDev and I need your feedback for me to continue.

lupus590:
So I've moved the `isMountPoint` function to the FS API in the bios and also done similar changes to move and rename. I'm now going to check the copy program and then I think that it all of the file system programs.

Maybe I should rename the pullrequest to FS program improveents.

lupus590:
While I'm thinking of it, why are rename and move seperate programs and not one an alias of the other?

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -797,6 +797,17 @@ function fs.complete(sPath, sLocation, bIncludeFiles, bIncludeDirs)
     return tEmpty
 end
 
+function fs.isMountPoint(sPath)
+    expect(1, sPath, "string")
+    if not fs.isDir(sPath) then
```
SquidDev, line 802: _Technically_ files can also be mounts (see #90), so I'm not sure if we want this.

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -797,6 +797,17 @@ function fs.complete(sPath, sLocation, bIncludeFiles, bIncludeDirs)
...
+        return false
+    elseif fs.combine(sPath, "..") == ".." or fs.getDrive(sPath) == fs.getDrive(fs.combine(sPath, "..")) then -- the == ".." check prevents errors when path is "/"
```
SquidDev, line 804: This might be cleaner as:

```lua
local parent = fs.getDir(path)
return parent == ".." or fs.getDrive(path) ~= fs.getDrive(parent)
```

Hrmr, question: do we want to count `/` as a mount point? I'd argue yes.

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -797,6 +797,17 @@ function fs.complete(sPath, sLocation, bIncludeFiles, bIncludeDirs)
     return tEmpty
 end
 
+function fs.isMountPoint(sPath)
```
SquidDev, line 800: "Mount point" feels like a bit of an odd name. It's obviously technically correct, but but not really something CC uses elsewhere. I don't know if `isDriveRoot` is any clearer (given `getDrive` is a thing)?

Can you add some docs to `doc/stub/fs.lua` too? Hopefully one day we'll be smart enough to pull them in from the bios.

src/main/resources/assets/computercraft/lua/rom/programs/delete.lua
```diff
@@ -6,12 +6,20 @@ if args.n < 1 then
...
-    local files = fs.find(shell.resolve(args[i]))
+    local resolvedPath = shell.resolve(args[i])
+    local files = fs.find(resolvedPath)
     if #files > 0 then
         for _, file in ipairs(files) do
-            local ok, err = pcall(fs.delete, file)
-            if not ok then
-                printError((err:gsub("^pcall: ", "")))
+            if file ~= resolvedPath and fs.isReadOnly(file) then
+                print("Skipping read only " .. file)
```
SquidDev, line 14: Saying "Skipping" seems a little odd, as it seems like a debug message when this is an error. I'd be inclined to say something more direct like `"Cannot delete read-only file /%s"` (and use `printError`).

src/main/resources/assets/computercraft/lua/rom/programs/delete.lua
```diff
@@ -6,12 +6,20 @@ if args.n < 1 then
...
+            elseif fs.isMountPoint(file) then
+                print("Skipping seperate mount " .. file)
+                print("To delete it's contents run rm " .. fs.combine(file, "*"))
```
SquidDev, line 17: `its` :p.

You may also want to do `"rm /%s"` - otherwise this'll be wrong when running from outside `/`.

src/main/resources/assets/computercraft/lua/rom/programs/move.lua
```diff
@@ -8,11 +8,31 @@ end
...
 local tFiles = fs.find(sSource)
+
+local function sanityChecks(sSource, sDestination)
+    if fs.isMountPoint(sSource) then
+        printError("Can't move mounts")
+        return false
+    elseif fs.isReadOnly(sSource) then
+        printError("Source is read-only")
+        return false
+    elseif fs.exists(sDest) then
```
SquidDev, line 19: I feel the destination checks should be done outside the loop - otherwise we'll print them for every file which feels a little odd.

lupus590:
I feel like `fs.attributes` could do with a `isMountPoint` in its returned table. I don't know how to do that though.

src/main/resources/assets/computercraft/lua/bios.lua
Lemmmy: I feel like `isDriveRoot` would be confusing, it sounds like it should only return true for `/`. Might be helpful to consider that disks call this `getMounthPath`, so `isMountPath` might also work.

lupus590:
I'm not sure if I did exactly what you wanted for `doc/stub/fs.lua`

src/main/resources/assets/computercraft/lua/bios.lua
lupus590: how about `isMountRoot`?

src/main/resources/assets/computercraft/lua/bios.lua
```diff
@@ -797,6 +797,11 @@ function fs.complete(sPath, sLocation, bIncludeFiles, bIncludeDirs)
     return tEmpty
 end
 
+function fs.isDriveRoot(sPath)
+    expect(1, sPath, "string")
+    return fs.getName(sPath) == "root" or fs.getDrive(sPath) ~= fs.getDrive(fs.getDir(sPath))
```
SquidDev, line 802: This means `fs.isDriveRoot("foo/root")` returns true, which is definitely not correct.

doc/stub/fs.lua
```diff
@@ -18,6 +18,7 @@ function getDrive(path) end
 function getFreeSpace(path) end
 function find(pattern) end
 function getDir(path) end
+function isDriveRoot(path) end
```
SquidDev, line 21: Docs!

src/main/resources/assets/computercraft/lua/rom/programs/delete.lua
```diff
@@ -6,12 +6,22 @@ if args.n < 1 then
...
 for i = 1, args.n do
-    local files = fs.find(shell.resolve(args[i]))
+    local resolvedPath = shell.resolve(args[i])
+    local files = fs.find(resolvedPath)
     if #files > 0 then
         for _, file in ipairs(files) do
-            local ok, err = pcall(fs.delete, file)
-            if not ok then
-                printError((err:gsub("^pcall: ", "")))
+            if file ~= resolvedPath and fs.isReadOnly(file) then
```
SquidDev, line 13: Why the `~= resolvedPath` check?

src/main/resources/assets/computercraft/lua/rom/programs/move.lua
```diff
@@ -8,11 +8,33 @@ end
...
+    return false
+elseif fs.isReadOnly(sDest) then
+    printError("Destination is read-only")
+    return false
+end
+
 if #tFiles > 0 then
     for _, sFile in ipairs(tFiles) do
         if fs.isDir(sDest) then
+            if not sanityChecks(sFile) then return end
```
SquidDev, line 34: Hrmr, I wonder if it's better to try to copy the other files rather than aborting entirely - like we do for `rm`.

src/main/resources/assets/computercraft/lua/rom/programs/move.lua
```diff
@@ -8,11 +8,33 @@ end
...
+    elseif fs.isReadOnly(sSource) then
+        printError("Source is read-only")
+        return false
+    end
+    return true
+end
+
+if fs.exists(sDest) then
+    printError("Destination exists")
+    return false
```
SquidDev, line 25: Don't need to return `false` here.

src/main/resources/assets/computercraft/lua/bios.lua
lupus590: I'm a bit curious as to why fs.getName("/") gives `"root"` then.

SquidDev: ¯\\\_(ツ)\_/¯ Because Dan?

src/main/resources/assets/computercraft/lua/rom/programs/delete.lua
lupus590: No idea, it seems to work fine without it so I've removed it.

PR #441, opened 2020-05-03
Add a monitor renderer using TBOs

SquidDev:
This uses the system described in #409 (or at least, how I understand it), to render monitors in a more efficient manner.

Each monitor is backed by a texture buffer object (TBO) which contains a relatively compact encoding of the terminal state. This is then rendered using a shader, which consumes the TBO and uses it to index into main font texture.

Performance wise, this has little impact on general FPS, but is significantly better on updates. Compare with #412 and #356:

image

That small blip is where we jump to >30fps is where the monitors update.

---

This is very much a WIP PR. My OpenGL skills are pretty much non-existent, so the implementation of this is no doubt terrible.

 - [x] Get this working on an older version of OpenGL. TBOs are introduced in OpenGL 3.1 (#409), so we should be able to go as far back as that. However, I don't _believe_ 3.1 supports `layout` qualifiers, so we may have to remove them.

 - [x] Test with Optifine's shaders. We may just need to fallback to the VBO renderer if Optifine is installed.

 - [x] Backport to 1.12. This is currently targetting 1.15, as it has a much nicer render system. However there's no reason we can't backport this to 1.12.

PR #443, opened 2020-05-04
 Add a monitor renderer using TBOs

SquidDev:
This is a backport of #441 for 1.12. See there for a description of the changes and improvements to performance.

This PR is almost identical - the only difference is we use `glGetFloatv` to fetch the model/view matrix, as it is not available to the client.

### DIFF
--- a/src/main/java/dan200/computercraft/client/gui/FixedWidthFontRenderer.java
+++ b/src/main/java/dan200/computercraft/client/gui/FixedWidthFontRenderer.java
@@ -50,12 +50,12 @@ public final class FixedWidthFontRenderer
     {
     }
 
-    private static float toGreyscale( double[] rgb )
+    public static float toGreyscale( double[] rgb )
     {
         return (float) ((rgb[0] + rgb[1] + rgb[2]) / 3);
     }
 
-    private static int getColour( char c, Colour def )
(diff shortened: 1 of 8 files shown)


Lemmmy:
Confirmed working configurations. All tests done with the Urn monitor updating 20 times per second. Will update this over the next half hour or so:
- [NVIDIA GTX 760 Ti (Linux, proprietary driver)](https://www.youtube.com/watch?v=-rgSmSLktt0)
- [AMD R9 390 (Linux, Radeon driver, locked vsynced)](https://i.cloudninja.pw/ObedientHighfalutinCockroach.png)
- [NVIDIA MX150 (Linux, nvprime)](https://plt.abby.how/means/ParallelLateTown.png)
- [NVIDIA RTX 2080 (Windows)](https://i.lemmmy.pw/tgJq.png)
- [Intel Integrated (Linux, i915/mesa)](https://plt.abby.how/means/PracticalLinearTurn.png)
- [AMD Vega 8 (Windows)](https://i.lemmmy.pw/iqF9.png)

These are a relatively diverse set of particularly extreme conditions, mainly to ensure the shader compiles, but also to compare the performance benefits.

All users compiled the shader successfully (and confirmed they were actually using TBO), no errors in the logs.

PR #447, opened 2020-05-09
Replace getMethodNames/callMethod with annotations

SquidDev:
This is an attempt to redesign ComputerCraft's Java API to be a little more user friendly and hopefully more future proof as CC develops. This is a breaking change, hence the fact that it is targetting Minecraft 1.15.

These changes are still very much in the development stage - I want to get some feedback on specific concepts, and give people an idea of what I'm working towards

### Major changes

 - Instead of defining methods using `getMethodNames`/`callmethod`, one may annotate a methods with the `@LuaFunction` annotation. This is then exposed to the user as a peripheral/lua method.

   This has several advantages:
    - No more faffing around with method arrays/indices. The method and method name appear in one place.
    - We can be more flexible in both the return and argument types - for instance, one can simply declare something as `void`. This also gives us greater flexibility to extend the API in the future.
    - One can make functions run on the main thread simply by annotating them as `mainThread = true`.

 - Instead of having a `ILuaAccess.yield` method, one now returns a `MethodResult.yield`. This is then interpreted by the Lua runtime (i.e. Cobalt), acting as a standard Lua yield.

   As Java-side yields are now non-blocking, this means we no longer spawn new threads on peripheral calls.

### Further changes
There's some further changes I'd quite like to make, but I'm less sure about.

 - [x] **Arguments object:** Instead of accepting an object array, we should provide an `IArguments` interface (much like OC does). This'd encourage people towards "proper" argument validation - `ArgumentHelper` isn't incredibly obvious right now.

 - [x] **Generate argument conversion code automatically:** Namely, one should be able to write `public final Object[] exec(String command)`, and we'll generate the appropriate `ArgumentHelper` calls.

 - [ ] **Pick up static methods:** `public static Object[] exec(TileEntityCommandBlock block, String foo)` should be applied to any `TileEntityCommandBlock`.

   This comes dangerously close to a peripheral system like Plethora, which is kind of the point - I'd like to merge some integration code into CC in the future. This is probably safe to leave off in a separate PR.

 - [x] **Benchmark and optimise:** There's inevitably going to be some additional overhead here - it'd be nice if we could minimise this. At a bare minimum, this PR would be equally fast as master - ideally faster.

 - [x] **Tests** 

### Problems
 - [x] **No composition:** One thing we can do with the previous API was to extend arbitrary `ILuaObject`s with additional methods (see `HttpResponseHandle`). However, as we're now tied to Java methods (and so Java's inheritance model) this is no longer possible.
   
   I'm not sure what the correct thing to do here is. One dodgy idea would be some `getComponents(): List<Object>` function (or alternatively, some `ObjectCollection`, which merges all objects together). It feels a wee bit of a hack though, and is not especially compatible with peripherals.

 - [x] **Peripheral stack levels are broken again:** In order to solve https://github.com/SquidDev-CC/CC-Tweaked/issues/363#issuecomment-604012795, we now catch errors thrown from peripherals and bump their level by one.

   This is now broken, as we only catch errors thrown immediately - not those thrown after a yield. One fix would be some `MethodResult.bumpError(level: int): MethodResult` method, but that's helluva ugly!


SquidDev:
Ughr, I was hoping I'd turned that off.

PR #453, opened 2020-05-18
Rewrite monitor networking

SquidDev:
This implements the initial rewrite of monitor networking, as described in https://github.com/SquidDev-CC/CC-Tweaked/pull/409#issuecomment-629648943.

 - [x] **Send monitor data as a separate packet.**

 - [x] **Compress the monitor data:** See the description in e3fab2a8a6d94489fde5f5af807dfc7a21801b6c, but this generally results in a 2-4x reduction in packet size, at the cost of server time.

   We probably need to do some benchmarking to see the performance impact of this. gzip is reasonably fast, but there's going to be a point where it causes issues to TPS. Can we get away with a lower level of compression?

 - [x] **Bandwidth limits:** Looking at the implementation, we may have to do this as a server-wide limit rather than a per-player one. However, this would provide some mitigation for the above - there's a hard limit on how much we can compress each tick and so (hopefully) on how much server time we use.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/network/client/TerminalState.java
+++ b/src/main/java/dan200/computercraft/shared/network/client/TerminalState.java
@@ -100,10 +100,7 @@ public class TerminalState
             buf.writeVarInt( height );
 
             ByteBuf sendBuffer = getCompressed();
-            int index = sendBuffer.readerIndex();
-            buf.writeVarInt( sendBuffer.readableBytes() );
-            buf.writeBytes( sendBuffer );
-            sendBuffer.readerIndex( index );
+            buf.writeBytes( sendBuffer, sendBuffer.readerIndex(), sendBuffer.readableBytes() );
         }
     }
 
@@ -124,7 +121,7 @@ public class TerminalState
         if( !compress ) return buffer;
         if( compressed != null ) return compressed;
 
-        ByteBuf compressed = this.compressed = Unpooled.directBuffer();
+        ByteBuf compressed = Unpooled.directBuffer();
         OutputStream stream = null;
         try
         {
@@ -140,7 +137,7 @@ public class TerminalState
             IoUtil.closeQuietly( stream );
         }
 
-        return compressed;
+        return this.compressed = compressed;
     }
 
     private static ByteBuf readCompressed( ByteBuf buf, int length, boolean compress )
--- a/src/main/java/dan200/computercraft/shared/network/client/TerminalState.java
+++ b/src/main/java/dan200/computercraft/shared/network/client/TerminalState.java
@@ -110,6 +110,11 @@ public class TerminalState
         return buffer != null;
     }
 
+    public int size()
+    {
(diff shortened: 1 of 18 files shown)


src/main/java/dan200/computercraft/shared/network/client/TerminalState.java
```diff
@@ -0,0 +1,180 @@
...
+        terminal.read( new PacketBuffer( buffer ) );
+    }
+
+    private ByteBuf getCompressed()
+    {
+        if( buffer == null ) throw new NullPointerException( "buffer" );
+        if( !compress ) return buffer;
+        if( compressed != null ) return compressed;
+
+        ByteBuf compressed = this.compressed = Unpooled.directBuffer();
```
SquidDev, line 127: I do not this this is safe. Packets may be sent on multiple threads at once, and so this is could lead to a race condition with multiple threads reading from `buffer`.

I kind of want to avoid compressing on the main thread (and compressing if there's no one watching this chunk), but that may be a premature optimisation.

src/main/java/dan200/computercraft/shared/peripheral/monitor/TileMonitor.java
```diff
@@ -148,7 +152,11 @@ public void updateTick()
             }
         }
 
-        if( m_serverMonitor.pollTerminalChanged() ) updateBlock();
+        if( m_serverMonitor.pollTerminalChanged() && m_xIndex == 0 && m_yIndex == 0 && m_serverMonitor != null )
```
SquidDev, line 155: We also need to send the terminal state when a player starts tracking a chunk.

src/main/java/dan200/computercraft/shared/peripheral/monitor/TileMonitor.java
```diff
@@ -286,6 +288,16 @@ protected final void readDescription( @Nonnull NBTTagCompound nbt )
             updateBlock();
         }
     }
+
+    public final void read( TerminalState state )
+    {
+        // This message may appear before the NBT update one, and so we construct a new monitor, but don't do anything
+        // with it.
+        // No, I'm not convinced this is correct.
```
SquidDev, line 296: Yeah, this is still super sketchy.

Lemmmy:
Might be worth checking the numbers on bandwidth caused by the cinema and compare

SquidDev:
Oh, definitely. How feasible is it to run the cinema locally nowadays - probably worth doing some stress testing before actually pushing this anywhere.

Lemmmy:
I'll have to ask 1lann in a few hrs but iirc it's relatively simple to deploy, just need to grab the latest build of the encoder and run it somewhere

SquidDev:
We deployed this on SC a couple of hours ago, and nothing exploded, so we're probably OK. I'll probably leave everything a few days, so we can keep an eye on any possible performance regressions.

SquidDev:
_Technically_ it's been a couple of days, even if in practice it's not been 36 hours.

PR #454, opened 2020-05-23
Simplify monitor TBO shader

Lignum:
Not sure how I missed this in the initial PR, but the fixed-function pipeline's MVP matrix is exposed as the built-in shader variable `gl_ModelViewProjectionMatrix`, there's no need to retrieve it ourselves. This saves us some uniforms and `glGetFloat` calls, which can incur some overhead.

### DIFF
--- a/src/main/java/dan200/computercraft/client/render/MonitorTextureBufferShader.java
+++ b/src/main/java/dan200/computercraft/client/render/MonitorTextureBufferShader.java
@@ -13,7 +13,6 @@ import dan200.computercraft.shared.util.Palette;
 import net.minecraft.client.renderer.OpenGlHelper;
 import org.apache.commons.io.IOUtils;
 import org.lwjgl.BufferUtils;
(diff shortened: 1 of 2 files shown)


SquidDev:
Huh, that's neat. I had assumed the two were entirely separate.

SquidDev:
Hrmrm. [This][1] and [this][2] seem to suggest that this functionality was removed in OpenGL 3.1. Is it safe to rely on these?

[1]: https://www.khronos.org/opengl/wiki/GLSL_:_recommendations#Built-in_Uniforms
[2]: https://www.khronos.org/opengl/wiki/OpenGL_Context#Context_types

Lignum:
Half the API was removed in OpenGL 3.1, but it's still perfectly fine to use it if your context is created with the compatibility profile, which Minecraft is. This is perfectly standards compliant.

SquidDev:
Thanks!

PR #455, opened 2020-05-23
Further monitor rendering improvements

Lignum:
This PR makes changes to how we pass data to the monitor shader.
- Frequently changing `uniform`s have been moved into a Uniform Buffer Object. Calling `glUniform` is slow, which this change fixes.
- The monitor TBO is no longer reallocated every frame (`glBufferData`). It is now recreated when it changes its size, but otherwise it is updated with `glMapBuffer`, which lets us write memory directly into the graphics driver, instead of having the driver copy it from us.

Further possible improvements:
- `glMapBuffer` gives us direct access to the driver's memory. Furthermore, we can use `glMapBufferRange` to only request access to part of the buffer. We can exploit these features by only uploading partial updates to the GPU. However, it is to be noted that for smaller buffers, it can sometimes be more efficient to just update the entire buffer in practice.

Please test and benchmark thoroughly before merging. In particular, since the new UBO contains palette information, I would be interested in how programs that frequently change the palette perform.


Lignum:
The renderer seems to explode when there are multiple monitors on screen.

src/main/java/dan200/computercraft/shared/peripheral/monitor/ClientMonitor.java
```diff
@@ -164,5 +164,7 @@ public static void destroyAll()
                 iterator.remove();
             }
         }
+
+        MonitorTextureBufferShader.deleteUBO();
```
SquidDev, line 168: I don't think we want this.

Both the UBO and shader are shared across multiple worlds, while monitors are not. It makes sense to delete the latter, but not the former.

src/main/java/dan200/computercraft/shared/util/Palette.java
```diff
@@ -120,4 +122,19 @@ public void readFromNBT( NBTTagCompound nbt )
...
         }
     }
+
+    @Override
+    public boolean equals( Object o )
+    {
+        if( this == o ) return true;
+        if( o == null || getClass() != o.getClass() ) return false;
+        Palette palette = (Palette) o;
+        return Arrays.equals( colours, palette.colours );
```
SquidDev, line 132: This line will always return `false`. It checks reference equality for each colour's `double[]`, but those are unique for each palette.

src/main/java/dan200/computercraft/shared/peripheral/monitor/ClientMonitor.java
```diff
@@ -33,6 +34,7 @@
     public BlockPos lastRenderPos = null;
 
     public int tboBuffer;
+    public ByteBuffer monitorBuffer;
```
SquidDev, line 37: Can we call this something like `tboContents`? `monitorBuffer` could refer to _anything_.

src/main/java/dan200/computercraft/shared/peripheral/monitor/ClientMonitor.java
```diff
@@ -68,13 +70,11 @@ public boolean createBuffer( MonitorRenderer renderer )
 
                 tboBuffer = OpenGlHelper.glGenBuffers();
                 OpenGlHelper.glBindBuffer( GL31.GL_TEXTURE_BUFFER, tboBuffer );
-                GL15.glBufferData( GL31.GL_TEXTURE_BUFFER, 0, GL15.GL_STATIC_DRAW );
+                monitorBuffer = null;
```
SquidDev, line 73: Maybe move this to `deleteBuffers`? It feels a little odd to have it kicking around here.

src/main/java/dan200/computercraft/shared/peripheral/monitor/ClientMonitor.java
```diff
@@ -68,13 +70,11 @@ public boolean createBuffer( MonitorRenderer renderer )
...
+
                 tboTexture = GlStateManager.generateTexture();
                 GL11.glBindTexture( GL31.GL_TEXTURE_BUFFER, tboTexture );
                 GL31.glTexBuffer( GL31.GL_TEXTURE_BUFFER, GL30.GL_R8, tboBuffer );
-                GL11.glBindTexture( GL31.GL_TEXTURE_BUFFER, 0 );
```
SquidDev, line 75: Just curious as to why these lines have been removed? I assume because we end up binding them later on anyway? If so, may be worth leaving a comment here.

src/main/java/dan200/computercraft/shared/peripheral/monitor/ClientMonitor.java
Lignum: Ah, didn't think of that. Where do you suggest I put this?

src/main/java/dan200/computercraft/client/render/MonitorTextureBufferShader.java
```diff
@@ -55,25 +58,29 @@ static void setupUniform( int width, int height, Palette palette, boolean greysc
         MATRIX_BUFFER.rewind();
         OpenGlHelper.glUniformMatrix4( uniformP, false, MATRIX_BUFFER );
 
-        OpenGlHelper.glUniform1i( uniformWidth, width );
-        OpenGlHelper.glUniform1i( uniformHeight, height );
+        GL15.glBindBuffer( GL31.GL_UNIFORM_BUFFER, uniformBuffer );
+        uboBuffer = GL15.glMapBuffer( GL31.GL_UNIFORM_BUFFER, GL15.GL_WRITE_ONLY, uboBuffer );
+        uboBuffer.putInt( UNIFORM_OFFSETS.get( 0 ), width );
```
SquidDev, line 63: The whole `.get(_)` everywhere is _kinda_ gross. Ughr, I'd prefer if we had individual offset/stride fields, but then I realise that makes querying them more ugly.

src/main/java/dan200/computercraft/shared/util/Palette.java
Lignum: `equals` and `hashCode` aren't supposed to be here anyway, they're a remnant of when I was working on something else.

src/main/java/dan200/computercraft/shared/peripheral/monitor/ClientMonitor.java
SquidDev: Do we need it at all? We never clean up the shader after all, and I assume the lifetimes of the two should be the same?

src/main/java/dan200/computercraft/shared/peripheral/monitor/ClientMonitor.java
Lignum: Sure.

src/main/java/dan200/computercraft/shared/peripheral/monitor/ClientMonitor.java
Lignum: It's standard in OpenGL that you bind all resources that you need when you need them, therefore we don't need to waste time unbinding everything. Though I wouldn't be surprised if Minecraft didn't follow this principle, to be honest. Seems to work fine, though.

src/main/java/dan200/computercraft/shared/peripheral/monitor/ClientMonitor.java
Lignum: This line doesn't actually do anything, don't know why it's here.

SquidDev: Well, I guess there's no harm in having it, as it frees a (tiny) bit of memory. But yes, feels a little odd.

src/main/java/dan200/computercraft/shared/peripheral/monitor/ClientMonitor.java
Lignum: It's kinda like not `free`ing memory that resides for the lifetime of the entire program anyway. Technically the OS will clean it up for us, it's just kinda ugly not to do it.

So no we don't need it, but it makes me feel better about myself. I'll happily remove it if you like.

SquidDev: Please. I'm not even sure Minecraft has a shutdown handler.

It's not like we or MC free any of the other buffers we allocate afterall.

src/main/java/dan200/computercraft/client/render/MonitorTextureBufferShader.java
Lignum: > Ughr, I'd prefer if we had individual offset/stride fields, but then I realise that makes querying them more ugly.

You'd effectively just be moving all the `.get(_)`s somewhere else. LWJGL without a wrapper just isn't pretty, I guess we have to live with that.

SquidDev: Oh I know. But the ugliness would be in one place, rather than two.

src/main/java/dan200/computercraft/shared/peripheral/monitor/ClientMonitor.java
Lignum: Not even that, `monitorBuffer` is `null` to begin with. I think I wrote that down to clarify that it is supposed to start off empty? Either way, I removed it in the latest commit.

src/main/java/dan200/computercraft/client/render/MonitorTextureBufferShader.java
Lignum: Fair enough, done.

SquidDev:
So LWJGL is refusing to map the buffer? Odd.

Lignum:
Okay, the issue only seems to happen when OptiFine is installed. The only reason why `glMapBuffer` would return `null` is if there was an OpenGL error. However, those would show up in the terminal, yet I see none in `latest.log`.

I'm tempted to just add a null check there and hope nothing breaks.

Lignum:
That seems to have fixed it without any apparent breakage. ¯\_(ツ)_/¯

lupus590:
> That seems to have fixed it without any apparent breakage. ¯_(ツ)_/¯
@Lignum 

"you dropped this \\" - /u/missingArmBot

src/main/java/dan200/computercraft/client/render/TileEntityMonitorRenderer.java
```diff
@@ -162,22 +159,36 @@ private static void renderTerminal( ClientMonitor monitor, float xMargin, float
...
+                        monitor.tboBufferSize = width * height * 3;
+                        GL15.glBufferData( GL31.GL_TEXTURE_BUFFER, monitor.tboBufferSize, GL15.GL_DYNAMIC_DRAW );
+                    }
+
+                    monitor.tboContents = GL15.glMapBuffer( GL31.GL_TEXTURE_BUFFER, GL15.GL_WRITE_ONLY, monitor.tboBufferSize, monitor.tboContents );
+                    ByteBuffer monitorBuffer = monitor.tboContents;
+
+                    // When OptiFine is installed, glMapBuffer returns null sometimes for some reason.
+                    // TODO: Investigate this and fix it properly rather than using this hack.
+                    if ( monitorBuffer != null )
```
SquidDev, line 175: If the buffer is null, does this mean the monitor entirely fails to update? If so, is it worth falling back to a version which creates a new buffer and calls `glBufferData`?

Lignum: I honestly have no idea. In the testing I've done with OptiFine, it seems to work fine, but we do need to test this thoroughly.

> If so, is it worth falling back to a version which creates a new buffer and calls glBufferData?

No, but that gave me an idea. We can fall back to using `glBufferSubData` instead. It's more expensive than `glMapBuffer`, but there's pretty much no good reason why it should fail.

SquidDev: Yeah. I'm somewhat surprised it's not printing anything to the logs. I was fairly sure Minecraft dumped `glGetError` after each frame.

Lignum: It does. Which could potentially mean that `glMapBuffer` isn't erroring, [but as the spec states](https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/glMapBuffer.xhtml):

> If an error is generated, a `NULL` pointer is returned.

There is no other situation I can find in the spec where it should return `null`. Unless there's an exception for LWJGL's bindings?

Lignum: It's worth noting the existence of the [OpenGL 4.3's debugging features](https://www.khronos.org/opengl/wiki/Debug_Output).

Instead of polling `glGetError`, it lets us hand a callback to the driver, which will give us detailed errors, warnings and informational messages. However, I have no idea how well this would interact with Minecraft's existing code.

I don't see how this would give us errors that `glGetError` doesn't detect though.

SquidDev: I _think_ Minecraft 1.15 does register this callback, so I might have a fiddle in porting this across. Though yes, don't know if it'd actually reveal anything.

Lignum:
b49a55d combined with #459 should provide stable gameplay. A JAR built with both PRs is available [here](https://lignum.pw/cc-tweaked-1.12.2-1.88.0.jar). I've recruited some playtesters, and I encourage anyone reading this to join in.

Lignum:
Bug: Switching to the Vertex Buffer renderer and back to Texture Buffers causes all monitors rendered during the switch to become black, until they are reloaded.

SquidDev:
Set `tboContents` to null when calling `deleteBuffers`? That should work.

Lignum:
> Set tboContents to null when calling deleteBuffers? That should work.

Unfortunately not. I'll look into this later.

SquidDev:
Oh right. Set `tboBufferSize` to `-1` instead. Sorry, forgot the condition for invalidating buffers.

Lignum:
@SquidDev Cheers, that worked.

SquidDev:
I just tested this locally, and it has _significantly_ worse performance on my system. As in, [display list][dl] levels of performance.

I'm afraid I've not done too much profiling yet - I'm not sure if it's monitor updates which are slower, or just rendering in general. Don't know if you have any ideas?

[dl]: https://user-images.githubusercontent.com/4346137/74032606-48096800-49ac-11ea-9801-c4da9833e77b.png

Lignum:
My best guess is that the UBO updates are causing the slowdown. Since the UBO is shared by every monitor, it has to be updated every frame, and [we're not streaming it properly](https://www.khronos.org/opengl/wiki/Buffer_Object_Streaming). I figured this would still be better than just reallocating the buffer every single time we change our uniforms. However,...

> Since allocating storage is (likely) faster than the implicit synchronization, you gain significant performance advantages over synchronization.

The page I linked suggests otherwise. Can you try replacing the UBO update with a `glBufferData` call? The wiki also suggests:
> The very first thing you should do is make sure that STREAM is in your buffer's hint. 

Currently we use `DYNAMIC`, see if that changes anything.

If this is the problem, another solution would be to use [multiple buffers](https://www.khronos.org/opengl/wiki/Buffer_Object_Streaming#Explicit_multiple_buffering), but that's gonna get messy.

Since the actual monitor contents are only updated whenever the terminal changes (which, mind you, also only happens 20 times a second, rather than every frame!), doing updates with `glMapBuffer` should work fine here.

I can't really test this myself since my GPU is generally pretty fast no matter what, sorry.

SquidDev:
Part of the problem does appear to be `glMapBuffer` - switching over to `glBufferData` in both cases _appears_ to bring things back to similar levels.

I'll have a fiddle with your other suggestions and get back to you.

Lignum:
It's possible that the TBO is small enough that reallocating it is much cheaper than synchronising updates with the GPU. However, I think implementing multiple buffer streaming is still worth a shot. As it turns out, `glMapBuffer` calls that aren't streamed properly are much slower than I initially thought.

SquidDev:
OK, I have absolutely no clue. The above changes definitely improved things (now above 30fps), but it's still worse than master.

I reverted the changes to TBO updates (to always use `glBufferData` just to give me a baseline, but even then there's a good 10 fps difference. Switching between DYNAMIC/STREAM doesn't appear to have any significant impact on my machine either :/.

SquidDev:
@Lignum I've spent a little bit longer looking at this, and am unable to get the performance on par on my machine.

As of such, I'm afraid I'm inclined to close this - I'll cherry pick the "`texture2D` -> `texture`" change across, but I'm not sure what else is applicable. I guess it might be worth caching the `ByteBuffer` we use, so we're not allocating buffers for each monitor update.

Don't know if you have any thoughts, or other suggestions on things I could try to get the performance up?

Lignum:
> Don't know if you have any thoughts, or other suggestions on things I could try to get the performance up?

Nope, sorry. Though we have learned that the usual OpenGL optimisation tricks don't work on monitors, which is quite valuable information. I suppose this is because games usually have far more complex models with fancy animations. GPUs aren't really made for rendering bitmaps, and I suppose there isn't much that can be done about that.

> I'll cherry pick the "texture2D -> texture" change across, but I'm not sure what else is applicable. I guess it might be worth caching the ByteBuffer we use, so we're not allocating buffers for each monitor update.

:heavy_check_mark:

PR #458, opened 2020-05-25
Add functions for parsing and drawing nft

JakobDev:
This PR add functions for parsing and drawing a nft image as described in [CraftOS-Standards](https://github.com/oeed/CraftOS-Standards/blob/master/standards/6-nft.md).


JakobDev:
Why is this ignored @SquidDev?

src/main/resources/assets/computercraft/lua/rom/apis/paintutils.lua
```diff
@@ -278,3 +278,76 @@ function drawImage(image, xPos, yPos)
...
+--
+-- @tparam string path The file to load.
+--
+-- @treturn table|nil The parsed image data, suitable for use with
+-- @{paintutils.drawTextImage}, or `nil` if the file does not exist.
+function loadTextImage(path)
+    expect(1, path, "string")
+
+    if fs.exists(path) then
+        local file = io.open(path, "r")
```
SquidDev, line 332: Instead of doing `fs.exists`, I'd just do:

```lua
local file = io.open
if not file then return nil end
```

Otherwise this will error on directories.

src/main/resources/assets/computercraft/lua/rom/apis/paintutils.lua
```diff
@@ -278,3 +278,76 @@ function drawImage(image, xPos, yPos)
...
+    local sBackground = "F"
+    local sForeground = "0"
+    for c in image:gmatch"." do
+        if c == "\n" then
+            nLine = nLine + 1
+        elseif c:byte() == 30 then
+            sMode = "background"
+        elseif c:byte() == 31 then
+            sMode = "foreground"
+        elseif sMode == "background" then
```
SquidDev, line 301: I'm really not a fan of doing the `sMode` thing. I think a cleaner thing would be something like this:

```lua
local i, len = 1, #image
while i <= len do
  local c = image:sub(i, i)
  if c == "\31" and i < len then
    i = i + 1
    sForeground = image:sub(i, i)
  elseif -- ...
  end

  i = i + 1
end
```

JakobDev:
@SquidDev The requested changes has be done.

I prefer using image.nft instead of cc.image.nft because I think we should only put cc internal modules into the cc directory.

SquidDev:
My stance is been that `cc.*` should be for any module provided by CraftOS, irrespective of whether it's internal. This way we have a consistent namespace for all modules, and can (mostly) guarantee we won't conflict with any user modules.

JakobDev:
I had moved the modul inside the cc  namespace.

src/test/resources/test-rom/spec/modules/cc/image/nft_spec.lua
```diff
@@ -1,5 +1,5 @@
-describe("image.nft", function()
-    local nft = require("image.nft")
+describe("cc,image.nft", function()
```
Lupus590, line 1: you have a comma instead of a dot here

JakobDev:
I looks like you left none of my code over, but as long as it works I'm OK with it.

Lemmmy:
If you actually read the [contributing guidelines](https://github.com/SquidDev-CC/CC-Tweaked/blob/mc-1.15.x/CONTRIBUTING.md) maybe some of your code would be left over, your PR shouldn't have lint errors

JakobDev:
@Lemmmy If you take a look at the commits, you will see, that way more changed than just a few lint errors.

PR #459, opened 2020-05-27
Oops! No standards compliance

Lignum:
As of #454, monitors no longer render on strict drivers (e.g. Windows AMD driver) since apparently `GL_ARB_compatibility` is disabled. This should normally be enabled in the compatibility profile, but for some reason it's not, so less permissive drivers fail to compile the shader.

This PR fixes this by explicitly requiring it in our vertex shader.

### DIFF
--- a/src/main/resources/assets/computercraft/shaders/monitor.vert
+++ b/src/main/resources/assets/computercraft/shaders/monitor.vert
@@ -1,6 +1,4 @@
-#version 140
-
-#extension GL_ARB_compatibility : require
+#version 130
 
 in vec3 v_pos;
 
--- a/src/main/resources/assets/computercraft/shaders/monitor.vert
+++ b/src/main/resources/assets/computercraft/shaders/monitor.vert
@@ -1,5 +1,7 @@
 #version 140
 
+#extension GL_ARB_compatibility : require
+
 in vec3 v_pos;
 
 out vec2 f_pos;


SquidDev:
> Obviously they'll be there because compatibility, but I'm not 100% sure it's standards compliant.

> This is perfectly standards compliant.

🤔

SquidDev:
Do we need to check if the extension is present? From what I gather, older versions of Mesa don't support it.

Lignum:
> 🤔

Whoopsident.

> Do we need to check if the extension is present? From what I gather, older versions of Mesa don't support it.

Oof, that might be a good idea, yes.

SquidDev:
I assume we'll need two versions of the shader? One which uses the GLSL matrix, and the other using the older fetch + uniform code?

Obligatory "how much performance impact does this really have?". Can we try profiling with a lot of 1x1 monitors?

Lignum:
> I assume we'll need two versions of the shader? One which uses the GLSL matrix, and the other using the older fetch + uniform code?

Not necessarily, we can use preprocessor macros in GLSL.

> Obligatory "how much performance impact does this really have?". Can we try profiling with a lot of 1x1 monitors?

`glUniform` is generally quite slow, which is why I introduced UBOs, so it's not negligible. On Mesa versions old enough not to support `GL_ARB_compatibility` it's probably a hopeless feat to try and optimise anyway.

Profiling is a good idea though.

> I don't really see any urgency in merging this. I want to get it correct this time, and the broken code isn't in a released version after all.

Fair enough.

Lignum:
Nevermind all that, I did some research and found a relatively elegant solution. `gl_ModelViewProjectionMatrix` was deprecated specifically in GLSL 1.40. Since [mixing GLSL versions in a program is allowed](https://stackoverflow.com/a/30943770/7506861), we can just set our vertex shader to GLSL 1.30, which works fine.

SquidDev:
> mixing GLSL versions in a program is allowed

Thanks, I hate it. But if it works, I'll live.

Can we get folks from #443 to do a set of tests again? You should be able to download the jar from the GH actions page.

Ale32bit:
Just tested it:
Monitors are working correctly, whereas the version from master is broken.

(As requested by Lignum:)
I'm using Arch Linux 5.6.14-zen1-1-zen with Nvidia RTX 2080, driver version 440.82

SquidDev:
> Obligatory "how much performance impact does this really have?". Can we try profiling with a lot of 1x1 monitors?

Did a small test with 324 1x1 monitors with an unchanging white screen. I honestly can't see any performance difference between the two - it's ~75fps in either case.

I guess it might be worth setting up a few thousand monitors, but I haven't figured out a good way to automate the modems, so I'll pass. Either way, this approach is simpler than using `getUniformMatrix4`, so I guess there's no harm :).

Lignum:
> Did a small test with 324 1x1 monitors with an unchanging white screen. I honestly can't see any performance difference between the two - it's ~75fps in either case.

Note that drivers can and do optimise things. For example, the NVIDIA drivers often optimise display lists into VBOs where possible. So to get any meaningful data, this would have to be done on multiple systems. Or it's possible that it really is a negligible difference.

SquidDev:
> Note that drivers can and do optimise things

I'm using Intel integrated graphics. So while it's possible, I'd be surprised :p.

PR #465, opened 2020-06-02
make loadfile consistent with standard lua

kepler155c:
A simple change to have loadfile include the full path to source. This is consistent with standard lua: see: https://www.lua.org/source/5.1/lauxlib.c.html#luaL_loadfile

This change will allow a debugger to easily find the corresponding source file.

### DIFF
--- a/src/main/resources/assets/computercraft/lua/bios.lua
+++ b/src/main/resources/assets/computercraft/lua/bios.lua
@@ -7,7 +7,7 @@ local expect
 
 do
     local h = fs.open("rom/modules/main/cc/expect.lua", "r")
-    local f, err = loadstring(h.readAll(), "@expect.lua")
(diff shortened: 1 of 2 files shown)


SquidDev:
This something I've kinda wanted for a while, but I'm not entirely sure about it.

It's definitely more consistent, but we end up with the issue that file names/errors often end up taking significantly more space. This makes them rather unreadable, especially on turtles.

I'd appreciate some people [playing around with this][cc] (do change the resolution too!) and seeing how they feel about it.

[cc]: https://copy-cat.squiddev.cc/gh/pull/465/

SquidDev:
I'm afraid I'm going to close this for the time being. As mentioned above, it's a change I'd love to make, but I'm not currently comfortable enough that it won't make things worse.

PR #467, opened 2020-06-05
Pastebin with connection

Serval6:
I was told that pastebin will not necessarily stay ... I quickly made this code to be able to send to my account and then modify it from my browser (or download to edit on vim) ... I kept it working usual as long as one is not connected, the only modification when one is connected is that `put` sends to the user's account. I then add commands that seem useful to me.
Do not hesitate to tell me that this is not useful for you!

PS: if I have made spelling mistakes / coding style errors, do not hesitate to notify me.

If you are logged in `pastebin put <filename>` will send using your pastebin account

`pastebin connect` and `pastebin disconnect` is added to connect / disconnect
`pastebin list` gets your pastebin list
`pastebin infos` get your personal information
`pastebin delete <code>` allows you to delete one of your pastebin

### DIFF
--- a/src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
+++ b/src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
@@ -126,7 +126,7 @@ elseif sCommand == "disconnect" then
 elseif sCommand == "list" then
     local user_key = settings.get("pastebin.key")
     if not user_key then
-        io.stderr:write("You aren't logged")
+        io.stderr:write("You aren't logged\n")
         print("first use: pastebin connect")
         return
     end
@@ -144,7 +144,7 @@ elseif sCommand == "list" then
         response.close()
         if sResponse == "Bad API request, invalid api_user_key" then
             print("Failed.")
-            print("You aren't logged")
+            io.stderr:write("You aren't logged\n")
             print("first use: pastebin connect")
         elseif sResponse == "No pastes found." then
             print("Success.")
@@ -160,7 +160,7 @@ elseif sCommand == "list" then
 elseif sCommand == "infos" then
     local user_key = settings.get("pastebin.key")
     if not user_key then
-        io.stderr:write("You aren't logged")
+        io.stderr:write("You aren't logged\n")
         print("first use: pastebin connect")
         return
     end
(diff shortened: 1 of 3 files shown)


src/main/resources/assets/computercraft/lua/rom/programs/http/pastebin.lua
```diff
@@ -70,10 +75,172 @@ local function get(url)
...
+    username = read()
+    write("Password: ")
+    password = read("*")
+
+    write("Connecting to pastebin.com... ")
+    local response = http.post(
+    "https://pastebin.com/api/api_login.php",
+    "api_dev_key=" .. dev_key .. "&" ..
+    "api_user_name=" .. textutils.urlEncode(username) .. "&" ..
+    "api_user_password=" .. textutils.urlEncode(password)
```
lupus590, line 106: You are saving the password as plain text, personally I would ask for the users password each time theprogram needs it.

lupus590, line 106: Is there a way to get an API key from the website and then paste that into the program? I would rather not put my password into CC ad see it transmit my password as plain text.

Serval6: I agree with you, unfortunately I don't think they offer it

SquidDev: It's not transmitting your password as plain text - it's done over an encrypted connection.

Serval6: if necessary I can remove the fact that I save in the settings and ask each time to connect

lupus590:
To fix the build you will need to write tests, [here is the file](https://github.com/Serval6/CC-Tweaked/blob/master/src/test/resources/test-rom/spec/programs/http/pastebin_spec.lua) you need to change, you can run tests locally with `./gradlew test` on the command line at the repo root.

SquidDev:
Sorry for taking so long to look at this, I'm not very good at this maintaining lark, and it's taken me a while to work out what I want to do with this.

I think this a really cool change, but not sure it's something I want in CC:T itself. Interfacing with the API in more depth adds a fair bit of extra complexity to the pastebin program, and I think that's probably better served by an external program people can download, rather than something bundled with CC:T itself.

Sorry.

PR #474, opened 2020-06-16
Add Module to translate CraftOS

JakobDev:
I added a module to translate CraftOS. The language files are in the same format as the language files of forge. The selected language is controlled with a setting. I only translated copy.lua, to show the usage.


Lignum:
e

SquidDev:
I'd appreciate some comments from other people on this. In principle I'm in favour of this, but it does add significant complexity and our current translations are pretty patchy, so I'm wary of adding an order of magnitude more.

JakobDev:
I just added a few more translations.

I think we should give users the possibility to translate programs. I will add a full German translation. Maybe some people will add translation for their language too. It's not a big problem if only the important things are translated.

Should we translate motd too? This should be done by creating a folder named motd with the files e.g. en_us.txt, de_de.txt inside.

Lupus590:
This all works fine for well written command line programs, but anything with a GUI is going to get into a problem with its UI strings changing length. What solutions do we have for that?

SquidDev:
> I think we should give users the possibility to translate programs. [...] It's not a big problem if only the important things are translated.

I quite agree that it's a good idea, but I do maintain it's a problem if it's not used. If translations aren't kept in sync, it just ends up being a maintenance nightmare.

I guess one possibility would be to integrate with something like [Weblate] or CrowdIn, which should make it much easier for people to add/edit translations, rather than having to create GH PRs.

> Anything with a GUI is going to get into a problem with its UI strings changing length

I assume we'd have to be careful when translating. It's also an issue with Minecraft itself - the config GUI is hilariously broken for some languages :).

---

@JakobDev I notice you've accidentally reverted some changes in this PR - for instance `shell.lua` is missing 086fccd997a7339187b0ed17ac0727e5f2d92321, and there's a couple of whitespace changes. Would you be able to fix those?


[weblate]: https://weblate.org/en-gb/hosting/

SquidDev:
> I guess one possibility would be to integrate with something like Weblate or CrowdIn

[Well, there we go](https://i18n.tweaked.cc/projects/cc-tweaked/minecraft/). We'll see how well this goes.

JakobDev:
@SquidDev Should be fixed now.

I know that gui programs could be a problem. I had to short some german translations in edit to avoid problems.

JakobDev:
I had now made all main programs translateable. Only the programs in the folders are missing.

JakobDev:
After a lot of hours I had now finished the translation except some GUI programs, that need more work that is better done in a sperate PR, adventure, because it has too many strings and pastebin, because I don't want a conflict with #467.

@SquidDev It should be ready to merge now.

kepler155c:
In order to allow people who write programs and would like to take advantage of the translation API, I would suggest loading translation files from a configurable path - similar to the Help API. You could use the settings API to allow for specifying the translation path.

SquidDev:
@kepler155c Personally I don't think this is something which is needed. As long as we expose suitable loading files, it should be possible to do require the module and then load your custom directory:

```lua
local i18n = require "cc.translate"
i18n.loadDirectory("my_custom_dir/")
local translate = i18n.translate
```

I wonder if it's worth exposing [several methods like `i18n.lua`][i18n] has, namely:
 - `loadDirectory(path: string)` - like we have now. I'd argue this should just be `loadDir`, as that's consistent with the `fs` API.
 - `loadFile(lang: string, path: string)` - Load a file.
 - `load(lang: string, items: { [string] = string })` - Load a table, for those who just want to embed things within their program.
 - `set(lang: string, key: string, txt: string)` - Set a single key.

Might also be worth changing `translate` to allow an setting the language as an optional second argument.

@JakobDev Would you be able to fix the lint warnings and other build failures? There's a whole load of global assignments which look rather problematic.

[i18n]: https://github.com/kikito/i18n.lua

 (and I'd argue `loadFile`

JakobDev:
@SquidDev Alsmost all Java Tests fail with ./gradlew test, but I haven't changed a single line of Java Code. Can you take a look at this? And is it possible, to just run the Lua tests?

What load of global assignments did you mean? translate is always local.

SquidDev:
> Almost all Java Tests fail with ./gradlew test, but I haven't changed a single line of Java Code

That's very odd - I only get 10 failures, and all within the Lua tests. Can you upload your builds log - I'm not sure why they'd be failing.

For the time being, you can run just the Lua ones with `./gradlew test --tests dan200.computercraft.core.ComputerTestDelegate` (or just run them from within IntelliJ, which makes things much easier).

> What load of global assignments did you mean? translate is always local.

Within the `cc.translate` module itself - you can go through the build log yourself, but a couple of snippets:
```
 src/main/resources/assets/computercraft/lua/rom/modules/main/cc/translate.lua:[9:5-9:8]: Setting unknown global variable "data". [var:set-global]
   │
 9 │     data = textutils.unserialiseJSON(file.readAll())
   │     ^^^^
src/main/resources/assets/computercraft/lua/rom/modules/main/cc/translate.lua:[16:10-16:22]: Setting unknown global function "loadDirectory". [var:set-global]
    │
 16 │ function loadDirectory(path)
    │          ^^^^^^^^^^^^^
src/main/resources/assets/computercraft/lua/rom/modules/main/cc/translate.lua:[17:5-17:9]: Setting unknown global variable "files". [var:set-global]
    │
 17 │     files = fs.list(path)
    │     ^^^^^
```

JakobDev:
The lua checks and lint works now without problems. I also added a few new function to the translate module.

JakobDev:
@SquidDev Could you plese merge? There are no comment since 15 days. This PR affects almost all programs in CraftOS, so the Chance that it will cause conflicts with other PRs/Commits are high.

SquidDev:
There's a couple of tests still failing - having had a look at the reports, the problem comes from loading the translate within `shell.lua`. We've basically got:

```lua
-- shell.lua
local translate = dofile("rom/modules/main/cc/translate.lua").translate

-- cc/translate.lua
local expect = require("cc.expect").expect
```

We have to use `dofile` in `shell`, as `require` isn't available yet. But that means we also need to use it within `cc.translate`. I guess something like this should work:

```lua
local expect = (require and require "cc.expect" or dofile("rom/modules/main/cc/expect.lua")).expect
```

Hardly going to win any prizes for good code, but it _should_ work.

---

As far as merging, I'm happy with the code itself, but still not 100% sold on the feature. As mentioned above, it adds a _lot_ of additional complexity and I am really worried about whether I'll be able to continue to maintain it.

I'd _really_ like some feedback from other people here, especially from some non-English speakers. As a Brit, I'm rather poorly equipped to judge how worth it this is :/. I've prodded people on Discord, so hopefully will get something.

neumond:
For cyrillic languages/chinese/japanese/etc CraftOS translation is useless, if not dangerous. Terminals are unable to display all these characters. There must be some linter to detect non-displayable characters in translations.

I think every Russian who's able to write code/run terminal commands will feel themselves comfortable with English in CC.

daelvn:
From a Spanish perspective, and that is, from myself and other people whose English level is not as advanced as mine, and barely counts as blabbering it, I also see no point in adding this to CraftOS.

If you do literally anything on the computer, you'll have to deal with something in English at some point. Increased chances if you're playing Minecraft, even more if you're installing mods and you're following a tutorial or going to download pages (which are not translated) for that, and specifically for this mod, programming, which has all to do with knowing English. I don't see the point in adding something which everyone can already understand well, given that they got to the point of installing the mod. This is also coming from my friends who are almost "illiterate" when it comes to English, but they can get around with untraslated mods because they're completely different paradigms, and they just get around because that's how playing with mods usually is.

Another point is that only the errors for the CraftOS programs are translated, but not other errors. I understand that some errors, like the ones that come from Java, you can't really translate, but consider this.

a. If you translate them, the users might understand the error, but if they are to find solutions for it online, well, good luck.
b. If you don't translate them, the users might not understand, and if they want to fix it they will have to go through English forum posts and chats to fix it, and if you already don't have the level for it, well, good luck.
c. Whatever you do, you're going to get a mixed translation at best where it's impossible to keep it consistent, and it will mess with the users who thought they could have a seamless and consistent experience.

People already come to the Discord server to get their errors fixed every other day, the forums used to be like that but it's not like I am active in them anymore. Because unless you got years of Lua practice with you, your programs will error, and a lot of them aren't going to know what's up. They will undeniably have to understand English to fix it, unless they want to give up on the mod. CraftOS isn't a platform where nothing is expected to error. Users are expected to make their own programs, and that goes hand by hand with erroring. It is going to happen no matter how hard you try.

I say that this doesn't go through really, it's pretty unnecessary and confusing. But that's just my opinion. The short version of this opinion is the comment above me, which I also agree with. Just, yeah...

SquidDev:
> For cyrillic languages/chinese/japanese/etc CraftOS translation [...] terminals are unable to display all these characters. 

OK, that's a good point I hadn't considered. Gah, full unicode support would be so nice to have, but we've never found a good design for it which doesn't break all existing programs. Thanks CC 1.76 for breaking that one :).

> Another point is that only the errors for the CraftOS programs are translated, but not other errors.

On one level, I don't think this is a major issue. You're effectively translating the UI, but not the internal errors. Though obviously given that CC is very much a programming platform, and as you say people will bump into them, it is worth worrying about.

I think for now this is something I'm going to close. Ughr, I'd like to make CC more accessible, and i18n is definitely one way to go about it, but right now I don't think we're in a position to properly deal with it.

Issue #452, opened 2020-05-18
OpenPeripheral, Plethora and the future


SquidDev:
## Background
Historically, CC has never had much integration with other mods. That has changed more recently, with support for JEI, MCMultipart, CraftTweaker and Charset.

However, what hasn't changed is that CC:T does not provide peripheral methods for external mods (or even Vanilla!). This has instead been provided by other mods, such as OpenPeripheral and Plethora. This is _fine_, but not without its problems:

 - **It's confusing:** Most people aren't aware that these methods come from other mods (see #449, and numerous questions on Discord and Reddit).
 - **We need to wait on other mods to update:** Both Plethora and OP added gameplay features, which add an additional burden to updating the mod.

This is, in short, a Problem.

## Proposal
I think the "best" solution for now is to merge the _vanilla_ (and Forge) integration into CC: Tweaked. This means we don't add any additional dependencies (which make updating the mod harder), but should provide enough functionality for most users.

Effectively, this means the following changes:
 - Merge _some_ of Plethora's API into CC: Tweaked.
 - Port over inventory, fluid and energy methods, and some vanilla/CC meta providers.
 - All remaining mod-integration code will stay in Plethora.

There's a lot of stuff about how the API should look which really needs to be hashed out before doing any work on this. Plethora has a lot of nice functionality, but it ends up being stupidly complex and somewhat broken (I'm looking at you converters).

SquidDev:
Yes, I realise there's nothing really actionable right now. I got halfway through writing this issue, and then realised that everything else was just going to be a list of problems with Plethora and me not having a clue on how to fix them. That can probably wait 'til later.

Lemmmy:
I agree with moving some of Plethora's basic features to CC:T. It always seemed to make sense to me that inventory interaction should be a native CC feature, given the sheer amount of things made possible by it. One thing that comes to mind, though, is the introspection module - I'm imagining you don't plan on bringing that or any of the other modules over, but it provides some basic functionality for turtles that wouldn't be available otherwise.

SquidDev:
I think the only the introspection module can do that turtles can't is getting complete item metadata?

Ideally `getItemDetails` would include that. However, it currently runs on the computer thread, which doesn't play well with meta providers.

Lemmmy:
[...]

PR #478, opened 2020-06-21
"Generic" peripherals

SquidDev:
This is a draft implementation of some of the features discussed in #452. The functionality is incredibly barebones for now; I'm much more interested in getting the framework together.

I'll try to get this merged into the next release, but gated behind a config option marked "experimental". I do want to make this available to users sooner rather than later, but be aware that things will change!


src/main/java/dan200/computercraft/core/asm/TaskCallback.java
```diff
@@ -6,19 +6,17 @@
 
 package dan200.computercraft.core.asm;
 
-import dan200.computercraft.api.lua.ILuaCallback;
-import dan200.computercraft.api.lua.LuaException;
-import dan200.computercraft.api.lua.MethodResult;
+import dan200.computercraft.api.lua.*;
```
hola-holo, line 10: our static analysis engine detects that your pull request includes a change from an explicit import to a wildcard import. According to the best practice (http://www.javadude.com/articles/importondemandisevil.html), import star (import on demand) is a bad practice, wildcard import can sometimes cause name conflicts and ambiguities.

src/main/java/dan200/computercraft/core/asm/TaskCallback.java
```diff
@@ -6,19 +6,17 @@
...
```
hola-holo, line 9: our static analysis engine detects that your pull request includes a change from an explicit import to a wildcard import. According to the best practice (http://www.javadude.com/articles/importondemandisevil.html), import star (import on demand) is a bad practice, wildcard import can sometimes cause name conflicts and ambiguities.

SquidDev: Unsubscribe

PR #484, opened 2020-06-28
Expose few publics data to turtle.getItemDetail()

### LINKED ISSUE #471: Add ability to read NBT with turtle.getItemDetail()
 - Explanation of how the feature/change should work :
When use turtle.getItemDetail() on a item(stack) the return should also include NBT data
To keep compatibility with previous version an optional arg "includeNBT = false" can be added.

 - Some rationale/use case for a feature. My general approach to designing new features is to ask yourself "what issue are we trying to solve" and _then_ "is this the best way to solve this issue?".
This would be very usefull to sort items based on they enchantments (to only keep enchants we don't already have)
Currently the only way to deal with the combinatorial explosion of enchantments is to use a very large amounts of NBT filters from other mods.

Naheulf:
- displayName (WARN : it's server localised)
- item damage (current and maximum), if you want I can change it to only display damage percent.
- enchantments
- storedEnchantments (used in anvil to apply enchants from enchanted books)
The last two respect the HideFlags NBT tag.

Resolve #471 (at least for my usage)
Can impact #452.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 # Mod properties
-mod_version=1.89.1
+mod_version=1.90.0
 
 # Minecraft properties (update mods.toml when changing)
 mc_version=1.15.2
--- a/src/main/java/dan200/computercraft/core/lua/CobaltLuaMachine.java
+++ b/src/main/java/dan200/computercraft/core/lua/CobaltLuaMachine.java
@@ -296,7 +296,6 @@ public class CobaltLuaMachine implements ILuaMachine
             LuaTable table = new LuaTable();
             values.put( object, table );
 
-            // Set<Map.Entry<K,V>> entrySet()
             for( Map.Entry<?, ?> pair : ((Map<?, ?>) object).entrySet() )
             {
                 LuaValue key = toValue( pair.getKey(), values );
@@ -327,8 +326,8 @@ public class CobaltLuaMachine implements ILuaMachine
 
         if( object instanceof CompoundNBT )
         {
-            CompoundNBT NBT = (CompoundNBT) object;
-            LuaTable table = new LuaTable(NBT.size(), 0);
+            CompoundNBT nbt = (CompoundNBT) object;
+            LuaTable table = new LuaTable( nbt.size(), 0 );
             values.put( object, table );
 
             // Set<K> keySet()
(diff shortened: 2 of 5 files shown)


gradle.properties
```diff
@@ -1,7 +1,7 @@
 # Mod properties
-mod_version=1.89.1
+mod_version=1.90.0
```
SquidDev, line 2: Can you revert this for the time being? I don't tend to bump this until I actually release, as this forces me to write (and check) the changelog.

src/main/java/dan200/computercraft/core/lua/CobaltLuaMachine.java
```diff
@@ -323,6 +324,63 @@ private LuaValue toValue( @Nullable Object object, @Nullable Map<Object, LuaValu
             return table;
         }
 
+        if( object instanceof CompoundNBT )
```
SquidDev, line 327: `dan200.computercraft.core` generally should avoid having any hard dependencies on Minecraft - we want to be able to run it outside of Minecraft too (such as in an emulator). What we do elsewhere (such as within the `commands` API) is to [convert the NBT to plain a `Map<?, ?>`](https://github.com/SquidDev-CC/CC-Tweaked/blob/514db30fb1d40c5a73e70c191120513166f356ea/src/main/java/dan200/computercraft/shared/util/NBTUtil.java#L91) within the method itself.

gradle.properties
Naheulf: Should I also revert the version number in whatsnew.txt and changelog.txt ? (to force recheck with :checkRelease ?)

SquidDev: Oh, yes please!

Naheulf: > I don't know if you've seen the changes in #478?
Nop. I haven't seen changes in #478. I just tried to use it but I did not succeed.

> As far as enchantments goes, I wonder if it's worth doing [what Plethora does](https://github.com/SquidDev-CC/plethora/blob/07ccc4d9b2652641c9290795996f2cebc306e320/src/main/java/org/squiddev/plethora/integration/vanilla/meta/VanillaMeta.java#L131), rather than just returning the raw NBT. This way we're not tied to how Mojang (currently) represents this information, and allows us to also include the enchantment's display name.
Yep.

I just thought of something: If we send the metadata acquisition to the main thread, it will necessarily cause a pause in the execution of the Lua program until the next tick ?. Isn't that likely to break programs? (Thx Google Translate)

SquidDev: >  it will necessarily cause a pause in the execution of the Lua program until the next tick ?. Isn't that likely to break programs? (

Yep, hence the `detailed` parameter. We'd preserve the existing no-delay functionality when it's missing, and return detailed information from the main tick if it's `true`.

Naheulf: This morning I thought of something: what if we added not one but two additional parameters ? 
`getItemDetail([slot, [detailed = false, [unsafeFastMode = false])`
detailed : get even more details (enchants, embedded items, ...)
unsafeFastMode : false = ask main thread to get advanced details, true = execute in lua thread (with a **_BIG RED WARNING_** in the doc)

SquidDev: No.

SquidDev:
As of 36bb8b67c927084c68d04823e91d98d95426af7e `turtle.getItemDetail` should support the detailed argument. It should now be sufficient just to add enchantment handling code to the `ItemDetail.fill` method.

Naheulf:
I will create another PR starting from [36bb8b6](https://github.com/SquidDev-CC/CC-Tweaked/commit/36bb8b67c927084c68d04823e91d98d95426af7e). I think it will be easier to merge both branches from your repo than from my repo.

Issue #133, opened 2019-03-01
Talkin' 'bout documentation


SquidDev:
### Update
[I'm hosting some proof-of-concept documentation](https://tweaked.cc/). Please give any comments you have on the accessibility, usefulness, etc... of the docs.

---

This is my documentation, this is my documentation baby.

Because this can't possibly end badly.

# The problem
CC:Tweaked's documentation is a bit of a mess right now. There's effectively three "sources" of documentation, none of them perfect.

 - **CraftOS's help files:** Ideally this would be the "canonical" source of documentation. However, I have failed to keep this updated as the rest of the mod has changed (see, for instance, `changelog`) and so it is woefully outdated.

   Moreover, the documentation itself is pretty flawed. It's very hard to have something readable given CC's small resolution, but the large monochromatic walls of text don't help. While documentation for programs is generally pretty good, documentation for APIs often ends up being little more than a list of functions provided, which isn't very helpful.

 - **[GitHub wiki](https://github.com/SquidDev-CC/CC-Tweaked/wiki):** This was added to document the changes to the original mod, which is fine, but again makes it very hard to consume as you have to refer to two sources.

 - **[cc.cc wiki](https://wiki.computercraft.cc/Main_Page):** This is definitely the best documentation we have right now. However, I personally find it hard to keep it up-to-date (despite the fantastic templates, adding new functions takes much longer than I'm comfortable with), which means some features are still not documented.

# The ideal
As far as I see it, there are three-to-four separate things to document:

 - **API/library reference:** A list of all functions, a summary of what they do and documentation on parameter/return values. Effectively the same sort of stuff you see for any library.

 - **Program reference:** List of all programs, what they do and what options/arguments they accept. Should probably also include some example usages - think `man` pages.

 - **Guides and tutorials:** I'm not sure if this counts as one or two things. But effectively more long-form write ups of one bit of the mod. This would probably include documentation on the blocks/items themselves, but also more general topics like "terminal redirects", "using the turtle API", etc...

In my opinion, the first two would be best served as being part of CC:Tweaked/CraftOS itself, while tutorials and long-form stuff should be driven by the community. Having reference documentation closer to the code should make it easier to ensure the documentation is in sync\*

\* Obviously this hasn't been the case so far - I'm definitely going to start requiring documentation updates as part of changes I or other people do.

# The proposal
There's a couple of changes I'd like to make:

[...]

PR #490, opened 2020-07-03
Added Javadoc for currently undocumented functions

MCJack123:
This PR adds some documentation for APIs that did not have docs in the
source yet. This includes:
* the drive peripheral
* the FS API
* the OS PAI
* the printer peripheral
* the speaker peripheral

The term API also needs documenting, but the TermAPI.java file does not
have the actual functions inside it, so they'll need to go elsewhere.
The turtle API also needs some further documentation, but I don't know
much about it. The official CC wiki can fill in the holes eventually.

This will help towards resolving #133, and builds upon 9f8774960fe01f673fc9eae48a2dbfff100fb770 by enhancing the resulting docs.

Note: I mostly used the standard Javadoc `@param` and `@return` tags, while using the `@cc` tags when dealing with optional parameters. I'm not sure what code style is preferred, but I tried to follow the style of other comments as best as possible.

### DIFF
--- a/src/main/java/dan200/computercraft/core/apis/FSAPI.java
+++ b/src/main/java/dan200/computercraft/core/apis/FSAPI.java
@@ -298,6 +298,7 @@ public class FSAPI implements ILuaAPI
      * @param mode The mode to open the file with.
      * @return A file handle object for the file, or {@code nil} + an error message on error.
      * @throws LuaException If an invalid mode was specified.
+     * @cc.treturn Handle A file handle object for the file, or {@code nil} + an error message on error.
      */
     @LuaFunction
     public final Object[] open( String path, String mode ) throws LuaException
@@ -359,6 +360,7 @@ public class FSAPI implements ILuaAPI
      * @param path The path to get the drive of.
      * @return The name of the drive that the file is on; e.g. {@code hdd} for local files, or {@code rom} for ROM files.
      * @throws LuaException If the path doesn't exist.
+     * @cc.treturn string The name of the drive that the file is on; e.g. {@code hdd} for local files, or {@code rom} for ROM files.
      */
     @LuaFunction
     public final Object[] getDrive( String path ) throws LuaException
--- a/src/main/java/dan200/computercraft/core/apis/FSAPI.java
+++ b/src/main/java/dan200/computercraft/core/apis/FSAPI.java
@@ -61,6 +61,12 @@ public class FSAPI implements ILuaAPI
         fileSystem = null;
     }
 
+    /**
+     * Returns a list of files in a directory.
+     *
+     * @param path The path to list.
+     * @return A table with a list of files in the directory.
+     */
     @LuaFunction
     public final String[] list( String path ) throws LuaException
     {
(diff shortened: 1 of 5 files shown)


src/main/java/dan200/computercraft/core/apis/FSAPI.java
```diff
@@ -268,6 +377,14 @@ public final void delete( String path ) throws LuaException
         }
     }
 
+    /**
+     * Returns the amount of free space available on the drive the path is
+     * located on.
+     *
+     * @param path The path to check the free space for.
+     * @return The amount of free space available, in bytes.
```
SquidDev, line 385: Can you add a `@cc.treturn number|"unlimited"` here. And possibly explain the whole "unlimited" thing.

src/main/java/dan200/computercraft/core/apis/OSAPI.java
```diff
@@ -153,24 +153,57 @@ private static long getEpochForCalendar( Calendar c )
         return c.getTime().getTime();
     }
 
+    /**
+     * Adds an event to the event queue. This event can later be pulled with
+     * os.pullEvent.
+     *
+     * @param name The name of the event to queue.
+     * @param args The parameters of the event.
+     * @cc.param [opt] ... The parameters of the event.
```
SquidDev, line 162: I've generally taken the approach that variadic arguments don't need to be `[opt]` - you can assume you can get away with passing 0 arguments to them.

I have a feeling you'll also need a `@cc.param string name` here. It's kind of an "all or nothing" approach.

src/main/java/dan200/computercraft/core/apis/OSAPI.java
```diff
@@ -271,6 +370,22 @@ public final int day( Optional<String> args ) throws LuaException
...
+     *   world was created. This is the default.
+     * * If called with {@code utc}, returns the number of seconds since 1
+     *   January 1970 in the UTC timezone.
+     * * If called with {@code local}, returns the number of seconds since 1
+     *   January 1970 in the server's local timezone.
+     *
+     * @param args The locale to get the seconds for. Defaults to {@code ingame} if not set.
+     * @return The seconds since the epoch depending on the selected locale.
+     * @cc.tparam [opt] string The locale to get the seconds for. Defaults to {@code ingame} if not set.
+     * @cc.treturn number The seconds since the epoch depending on the selected locale.
```
SquidDev, line 386: See comments at the top about `[opt]`. I don't _think_ you need the explicit `cc.treturn` either here though, as the function explicitly returns a `long`.

src/main/java/dan200/computercraft/shared/peripheral/diskdrive/DiskDrivePeripheral.java
```diff
@@ -76,18 +99,35 @@ public final void setDiskLabel( Optional<String> labelA ) throws LuaException
...
     }
 
+    /**
+     * Returns the mount path for the inserted disk.
+     *
+     * @param computer The computer object
+     * @return The mount path for the disk, or {@code nil} if no data disk is inserted.
+     */
     @LuaFunction
     public final String getMountPath( IComputerAccess computer )
```
SquidDev, line 121:
```suggestion
    @Nullable
    public final String getMountPath( IComputerAccess computer )
```

Can you chuck a `@Nullable` here? Should mean the inferred return type is correct.

src/main/java/dan200/computercraft/shared/peripheral/diskdrive/DiskDrivePeripheral.java
```diff
@@ -96,6 +136,11 @@ public final boolean hasAudio()
         return media != null && media.getAudio( stack ) != null;
     }
 
+    /**
+     * Returns the title of the inserted audio disk.
+     *
+     * @return The title of the audio, or {@code nil} if no audio disk is inserted.
+     */
     @LuaFunction
     public final Object getAudioTitle()
```
SquidDev, line 145: Again, a `@Nullable` here. Sorry, these should really have been here originally.

SquidDev:
> The `term` API also needs documenting, but the `TermAPI.java` file does not have the actual functions inside it, so they'll need to go elsewhere.

Just as an addendum, These all live in `TermMethods` right now, which is basically the methods exposed by a terminal redirect. There's probably a little bit more work which needs to be done on to the tooling for this to work correctly - same with file handles.

src/main/java/dan200/computercraft/core/apis/FSAPI.java
MCJack123: I added this, but I'm not really sure what causes `"unlimited"` to be returned. I think I've seen it before but I can't remember what caused it. Could you explain what would cause this? I see it's returned if the underlying FS's `getRemainingSpace` returns -1, but there doesn't seem to be an example of this in any implementations of `IWritableMount`.

SquidDev: So this code path is never hit using CC's builtin mounts, but I suppose it exists in the unlikely case that an external mod adds a mount without any file size restrictions? I'm afraid this is Dan code, so I can't really comment on the thinking behind it :).

SquidDev:
❤️ for fixing that, looks fantastic to me. Thank you very much!

PR #491, opened 2020-07-04
Add port to Host http header if necessary

neumond:
Host header must contain port if it is not default port for the protocol.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host

### DIFF
--- a/src/main/java/dan200/computercraft/core/apis/http/request/HttpRequestHandler.java
+++ b/src/main/java/dan200/computercraft/core/apis/http/request/HttpRequestHandler.java
@@ -86,7 +86,7 @@ public final class HttpRequestHandler extends SimpleChannelInboundHandler<HttpOb
         {
             request.headers().set( HttpHeaderNames.USER_AGENT, this.request.environment().getComputerEnvironment().getUserAgent() );
         }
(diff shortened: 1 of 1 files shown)


src/main/java/dan200/computercraft/core/apis/http/request/HttpRequestHandler.java
```diff
@@ -62,6 +62,28 @@
...
     }
 
+    private String getHostHeader()
+    {
+        String host = uri.getHost();
+
+        if( uri.getPort() >= 0 )
+        {
+            int defaultPort = -1;
+            switch ( uri.getScheme() )
```
SquidDev, line 72: I'd be inclined just to replace this with `uri.getScheme().equalsIgnoreCase("https") ? 443 : 80` - it's what we do in `HttpRequest`.

Alternatively, just drop the default port check. From what I can tell reading the spec, this _should_ be fine?? It'll only get triggered if people explicitly provide a port, so it doesn't seem like the end of the world if we include it.

neumond: Although okhttp and aiohttp.client do cut off default ports, I agree on dropping this chek.

SquidDev:
Thanks

PR #506, opened 2020-07-23
Add module for reading tarfiles

JakobDev:
I add a module for handling with tarfiles.

**Why we need that module?**
CC had currently no archive format that allows putting multiple files into once. This is use because you have to handle (copy,download etc) only one file instead of multiple files. tar is a very simple standard format that can be extracted and created by almost every archive program.

**What is planned?**
- Writing a archive
- Add a Commandline program
- Add support for reading strings directly without saving as file first

**What is not planned?**
- Support for compression

**Example code**
```lua
local tarfile = require("cc.tarfile")
local tar = tarfile.open("/test.tar","r")
tar:extractAll("/unpack")
```


Lemmmy:
This program might be better suited to publish on [the forums](https://forums.computercraft.cc/) rather than adding more bloat to the ROM

SquidDev:
Yeah, I'm afraid I agree with lem here.

SquidDev:
Having chatted with some other people, [I've set up a small repository](https://github.com/SquidDev-CC/metis/) for modules like this, which aren't really suited for CraftOS but are still helpful. Would you be OK for me to add it there?

JakobDev:
As long as you name me as author, I'm OK with it. All my programs are liensed under BSD 2-Clause. But I really think, CC should have a lib for reading and writing a package format. You can find the lib [here](https://gitlab.com/JakobDev/Computercraft/-/blob/master/Modules/tarlib.lua). There is also a [documentation for ghv](https://gitlab.com/JakobDev/Computercraft/-/blob/master/Help/tarlib.ghv) and a [commandline program](https://gitlab.com/JakobDev/Computercraft/-/blob/master/Tools/jtar.lua). Everything can also be installed with packman.

PR #508, opened 2020-07-24
Fix JSON serialization of strings

neumond:
`textutils.serializeJSON` was making invalid jsons if we include control characters into strings. It used `string.format("%q")` which is great for lua representation of strings, but not that great for json.


src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -335,6 +335,42 @@ empty_json_array = mk_tbl("[]", "empty_json_array")
...
+    local function control_replace(c)
+        local r = map[c]
+        if r ~= nil then return r end
+        return c
+    end
+
+    serializeJSONString = function(s)
+        return string.format(
+            '"%s"',
+            s:gsub("[%c\"\\]", control_replace):gsub(hpat, hexify)
```
SquidDev, line 369: I'm fairly sure we can do `:gsub("[%c\"\\]", map)` and drop the `control_replace` function.

src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -335,6 +335,42 @@ empty_json_array = mk_tbl("[]", "empty_json_array")
 -- @see textutils.unserialiseJSON
 json_null = mk_tbl("null", "json_null")
 
+local serializeJSONString
+do
+    local function hexify(c)
+        return "\\u00" .. string.format("%02X", string.byte(c))
+    end
+
+    local hpat = string.format("[%c-%c]", 0x7F, 0xFF)
```
SquidDev, line 344: I'd just replace this with `[\x7f-\xff]`, and inline it into the pattern below.

src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -335,6 +335,42 @@ empty_json_array = mk_tbl("[]", "empty_json_array")
 -- @see textutils.unserialiseJSON
 json_null = mk_tbl("null", "json_null")
 
+local serializeJSONString
+do
+    local function hexify(c)
+        return "\\u00" .. string.format("%02X", string.byte(c))
```
SquidDev, line 341:
```suggestion
        return string.format("\\u00%02X", string.byte(c))
```

src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
SquidDev: Actually, reading through the spec, it's not clear if we need this or not. Anything in the `\x20` to `\x10ffff` range is valid, which definitely includes all of CC's characters.

neumond: Yep, it's valid, but it's unclear what encoding we use to send jsons and what encoding receiving side expects. Spec recommends to encode in UTF-8 as the best guess, but we can safely encode everything using only ASCII. Python json encoder escapes everything from 0x7f as well, unless you explicitly disable this.

SquidDev:
Fantastic, thank you!

Issue #264, opened 2019-07-03
Ability to connect peripherals to a modem via lua


fatboychummy:
So I was making a thing today with turtles which builds some furnaces and etc on top of modems.  I, however, ran into a problem which cannot be solved via any available APIs.

I cannot connect said furnaces to the network, unless I walk my character to the location manually and do it myself.

I am requesting a simple addon to the modem API to be able to do this.

It would look something like the following:

```lua
local modem = peripheral.wrap("modem_bla")
modem.connect()
```

this would connect all peripherals attached to the modem to the network, and (potentially) return names of the newly added peripherals.  This would be useful not only for plethora, but also when using a turtle to build a computer-cluster attached to modems in 'vanilla' (just cc:tweaked).

fatboychummy:
This could also be addressed by #125, having the ability to right-click a modem would help as well.

fatboychummy:
Just adding another idea to this because random thoughts are random:

Having `modem_[dis]connected` events for when a modem is manually connected or disconnected would be helpful as well for a mainframe to ensure security in a wired network.

xAnavrins:
@fatboychummy `peripheral/peripheral_detach` events should work on remote peripherals if I'm not mistaken.

Lupus590:
Do modems fire peripheral events?

SquidDev:
Yes.

https://github.com/SquidDev-CC/CC-Tweaked/blob/fff8353451451be5ae31e0f63d8e529b127fd186/src/main/java/dan200/computercraft/shared/peripheral/modem/wired/WiredModemPeripheral.java#L324-L334

fatboychummy:
> @fatboychummy `peripheral/peripheral_detach` events should work on remote peripherals if I'm not mistaken.

While yes they do exist, if we get the ability to 'enable or disable' modems via Lua commands, it would be nice to know if the modem has been enabled so we can disable it immediately if need be, rather than looping through each modem to see what modem the peripheral returned by `peripheral_attach` is attached to.

Iluab21:
Hi @SquidDev. I was doing some testing with the "Turtles ♥ Modems" datapack and looked into `TurtlePlaceCommand.java`. I noticed why the current workaround feels a bit clunky for pure automation.

Right now, `execute()` fails early if the slot is empty:
[...]

PR #510, opened 2020-07-27
WIP: Modem peripherals toggling

neumond:
I would like to toggle modems by turtles to automate building.

Regarding the code: it works. Only with fullsized wired modems.
There's nothing else which could be called _good_, since I'm complete newbie in modding, in CC codebase and in Java to complete the picture. Therefore I'm ready to be kicked for bad code/patterns/structure.

I'm planning to add similar thing to usual wired modems and to think more about Lua API. And of course I'm open for discussion.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/peripheral/modem/wired/WiredModemElement.java
+++ b/src/main/java/dan200/computercraft/shared/peripheral/modem/wired/WiredModemElement.java
@@ -19,6 +19,7 @@ public abstract class WiredModemElement implements IWiredElement
 {
     private final IWiredNode node = new WiredNode( this );
     private final Map<String, IPeripheral> remotePeripherals = new HashMap<>();
+    private boolean sharingPeripherals = false;
 
     @Nonnull
     @Override
@@ -61,4 +62,37 @@ public abstract class WiredModemElement implements IWiredElement
     protected abstract void attachPeripheral( String name, IPeripheral peripheral );
 
     protected abstract void detachPeripheral( String name );
+
+    public boolean isSharingPeripherals()
+    {
+        return sharingPeripherals;
+    }
+
+    public void forceSetSharingPeripherals( boolean enabled )
+    {
+        // doesn't update Tile
+        sharingPeripherals = enabled;
+    }
+
+    public boolean trySetSharingPeripherals( boolean enabled )
+    {
+        // updates Tile if necessary
+        if( sharingPeripherals == enabled ) return true;
+        sharingPeripherals = enabled;
+
+        if( sharingPeripherals )
+        {
+            sharingPeripherals = enableSharing();
+        }
+        else
+        {
+            disableSharing();
+        }
+
+        return sharingPeripherals == enabled;
+    }
+
+    protected abstract boolean enableSharing();
+
+    protected abstract void disableSharing();
 }
--- a/src/main/java/dan200/computercraft/shared/peripheral/modem/wired/TileCable.java
+++ b/src/main/java/dan200/computercraft/shared/peripheral/modem/wired/TileCable.java
@@ -75,9 +75,25 @@ public class TileCable extends TileGeneric
         {
             m_modem.detachPeripheral( name );
         }
+
+        protected boolean enableSharing()
+        {
+            m_peripheral.attach( world, getPos(), getDirection() );
+            if( !m_peripheral.hasPeripheral() ) return false;
+
+            m_node.updatePeripherals( m_peripheral.toMap() );
+            updateBlockState();
+            return true;
+        }
+
+        protected void disableSharing()
+        {
+            m_peripheral.detach();
+            m_node.updatePeripherals( Collections.emptyMap() );
+            updateBlockState();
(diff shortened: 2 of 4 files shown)


neumond:
Looks like #264 is related.

neumond:
Regarding API.
Obviously I don't like current `modem.togglePeripherals()` method. This should be `is...` + `set...` pair. But I can't pick good name for this.
Another issue is that peripherals can't be detected from coroutine thread (when turning on) and we can't simply return whether it will be enabled or not. I looked at how modem.open updates the block, but it seems to be one-way unconditional passing of boolean with subsequent TickScheduler.schedule, then the server thread takes new value and updates the block.

SquidDev (changes requested):
`ModemPeripheral`/`ModemState` does weird things with `TickScheduler` because we don't actually care about what's done on the main thread - it's only there to update the visual state.

In our case, we do probably care about the result of `togglePeripheralAccess` - it'd be nice if we returned whether peripherals were enabled or not (or something).  In this case, we don't need to do anything with `AtomicBoolean`s or anything - just add a `@LuaFunction( mainThread = true )` on `togglePeripherals` and you should be safe to call the TE method directly.

As far as naming goes, I'm not too sure. Definitely agree that either `is`/`set` or `is`/`enable`/`disable` is the way to go. Not really sure what the best name is though - "Active" would be one candidate I guess, but it's not super clear.

neumond:
Interesting thing happens. With full block modems everything works as expected, but with cable modems it's only possible to share a computer directly attached to the modem. Turtles don't see cable modems as peripherals at all. Therefore they can't call any methods. Probably some `modem_control` lightweight peripheral can solve the issue, but I would like to discuss this.

SquidDev:
> Turtles don't see cable modems as peripherals at all

Wired modems shouldn't be able to attach to turtles at all, as turtles aren't a full block, so I don't really think this is an issue.

neumond:
I agree, turtles shouldn't use cable modems as modems. But they should be able to activate/deactivate neighbor modem as part of construction process. This could be a new type of peripheral, or may be possibility to `place()` (right click) at the side of modem. Probably better alternative exists.

src/main/java/dan200/computercraft/shared/peripheral/modem/wired/WiredModemElement.java
```diff
@@ -61,4 +62,31 @@ public void networkChanged( @Nonnull IWiredNetworkChange change )
...
+    public boolean isSharingPeripherals()
+    {
+        return sharingPeripherals;
+    }
+
+    public boolean setSharingPeripherals( boolean enabled, boolean updateTile )
+    {
+        if( sharingPeripherals == enabled ) return sharingPeripherals;
+        sharingPeripherals = enabled;
+        if( updateTile )
```
SquidDev, line 75: Doing this as a boolean flag feels super janky. It's probably worth having two methods, where one of them is a `trySetSharingPeripherals` or something.

SquidDev (changes requested):
Ahh, that makes sense. I'd not worry about that right now - adding another peripheral is kinda gross, and if people really want it they can use full block modems.

src/main/java/dan200/computercraft/shared/peripheral/modem/wired/WiredModemElement.java
neumond: Done. Now it's two methods.

Ocawesome101:
`isModemActive` and `setModemActive` could be good names.

Alternatively, `isPeripheralActive` and `setPeripheralActive`, or `isActive` and `setActive`, work if you don't want to tie this directly to the modem peripheral.

merlinlikethewizard:
Thinking of picking this up, am I right in thinking this issue in the mod was never resolved?

Ocawesome101:
correct. no such feature exists in the main mod.

khankul:
In #1469, 
> While I understand the use-case, it feels strange to me to add a method to the turtle API for interacting with a single block.

I suppose this PR can be closed too.

SquidDev:
This PR adds peripheral methods to the modem themselves, not the turtle. This is mostly stalled as I'm still not 100% convinced on the idea, but not against it either.

powerboat9:
Maybe turtles could toggle modems by using ```turtle.place```?

sugoidogo:
The inability of a turtle to turn on a modem feels more like something is missing than anything else. I actually went searching in the modem API to find the functions that this PR would add because I assumed it would be present, given that this is a computercraft peripheral and not a random block interaction. But I also feel like the arbitrary limiting of interactions is not great in the first place. Rather than hard-coding what a turtle can and can't interact with, why not make it a configurable option? I'd much rather be able to go into my server's config folder and simply enable the feature, or as a modpack dev include a default configuration for it based of the balance that I'm trying to achieve, much like how you can now with fuel usage.

khankul:
Looks like a datapack with modems added to `turtle_can_use` does the job now; you can get it [here](https://datapacks.madefor.cc/).

PR #512, opened 2020-07-28
Make the key for mtime "modified" in fs.attributes

### LINKED ISSUE #511: "Modification" key in fs.attributes is inconsistent with docs
Docs say: [modified](https://github.com/SquidDev-CC/CC-Tweaked/blob/mc-1.15.x/src/main/java/dan200/computercraft/core/apis/FSAPI.java#L470)
Code does: [modification](https://github.com/SquidDev-CC/CC-Tweaked/blob/mc-1.15.x/src/main/java/dan200/computercraft/core/apis/FSAPI.java#L481)

plt-hokusai:
Just like the docs say!

Fixes #511. See also: #513

### DIFF
--- a/src/main/java/dan200/computercraft/core/apis/FSAPI.java
+++ b/src/main/java/dan200/computercraft/core/apis/FSAPI.java
@@ -478,6 +478,7 @@ public class FSAPI implements ILuaAPI
         {
             BasicFileAttributes attributes = fileSystem.getAttributes( path );
             Map<String, Object> result = new HashMap<>();
+            result.put( "modification", getFileTime( attributes.lastModifiedTime() ) );
             result.put( "modified", getFileTime( attributes.lastModifiedTime() ) );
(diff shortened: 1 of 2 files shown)


JakobDev:
fs.attributes() is in CCT for a while now and some programs are already using it. Chancing the key will break compatibility without any benefit. It's way better to just change the documentation.

plt-hokusai:
Changing the documentation doesn't fix the issue of consistency, now, does it? One is `created`, while the other is `modified`. The options should be `modified` and `created` or `modification` and `creation`, _not_ a mix!

dmarcuse:
Clearly the best solution is to instead have four keys, `created`, `creation`, `modified`, and `modification`

SquidDev:
I think we should add an `__index` meta method which does Levenshtein distance to find the correct key. 

Or just add `modified` and `modification`.

PR #513, opened 2020-07-28
Change docs for fs.attribute() to say modification

### LINKED ISSUE #511: "Modification" key in fs.attributes is inconsistent with docs

plt-hokusai:
Just like the code does!

Fixes #511. See also #512.


SquidDev:
Closing in favour of #512.

Issue #487, opened 2020-07-01
textutils.pagedTabulate is not paged


neumond:
## Useful information to include:
 - Minecraft version: 1.15.2
 - CC: Tweaked version: 1.89.2 precisely, artifact [of this build](https://github.com/SquidDev-CC/CC-Tweaked/runs/822210259)

It works like tabulate, dumps whole table immediately, even if it doesn't fit the screen.
I wouldn't say I really need this bug fixed, I'm simply testing everything.

SquidDev:
Reminder for myself. This line, needs to be replaced with a call to `newLine()`.

https://github.com/SquidDev-CC/CC-Tweaked/blob/36bb8b67c927084c68d04823e91d98d95426af7e/src/main/resources/data/computercraft/lua/rom/apis/textutils.lua#L199

SquidDev:
Having another look at this, and there's a lot more bugs in the whole paging system than I was expecting. I'm honestly not quite sure how to best fix this.

PR #517, opened 2020-08-04
Fix textutils.pagedTabulate and tabulate

wieselkatze:
This addresses Issue #487 and is a first _proof of concept_ for fixing textutils.pagedTabulate and textutils.tabulate.

Shots of some tests:
ccpagedtabulate
ccpagedtabulate2
ccpagedtabulate3

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
+++ b/src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
@@ -152,61 +152,219 @@ function pagedPrint(_sText, _nFreeLines)
     return result
 end
 
-local function tabulateCommon(bPaged, ...)
-    local tAll = table.pack(...)
-    for i = 1, tAll.n do
-        expect(i, tAll[i], "number", "table")
+local function adjustColumnWidths(nWidth, tWidths)
+    --[[
+    Tries to fit all columns in tWidths on the screen width (nWidth)
+    Shrinks biggest columns down to either
+    1. Match the next smaller column
+    2. by 1 if there are equally sized columns to alternate between the two
+    3. By the amount the entire table is too wide if the column doesn't get too small
+    ]]
+    nWidth = nWidth - (#tWidths - 1) -- Spaces in between columns
+
+    local nSum = 0
+    -- Figure out the sum once; this well get adjusted accordingly later
+    -- and thus just needs to be done once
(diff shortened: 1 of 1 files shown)


JakobDev:
Cool, we are really needing this. There are a lot of cases were the content is longer than a line. Especially for showing arguments in a help.

But I think we shouldn't change the behavior of a function that is in CC for so long. I think this is better in a own module.

wieselkatze:
@JakobDev What do you suggest it would be moved to? As far as I understand, this is how textutils.tabulate and pagedTabulate were supposed to work, but never did.

SquidDev:
Would you be able to rebase this back onto the 1.15.2 branch? That's currently the trunk, so it makes the most sense to keep it there.

One problem I've found is that `pagedTabulate` won't correctly handle scrolling. It'll print until you reach the bottom of the terminal before printing "press any key" - if you're at the bottom of the terminal, this is immediately! Ideally we'd print the height of the terminal before prompting.

image

Though I guess ideally ideally, we'd have a proper "pager" function which generalises all of this :).

@JakobDev I don't think this changes anything major about the function? At least, not in a breaking way. I don't know, haven't played with it too much, so I may be missing something.

wieselkatze:
@SquidDev What exactly do you mean by it printing at the bottom of the terminal? I've tested that case and when I'm at the bottom of the screen (or one line before), it looks like this:
grafik

Is there maybe some double keystroke going on that already triggers the first key event?

SquidDev:
> What exactly do you mean by it printing at the bottom of the terminal? I've tested that case and when I'm at the bottom of the screen (or one line before), it looks like this:

Sorry, I wasn't especially clear. Basically it _shouldn't_ do this - I'd expect it to fill the screen immediately. So like so:

image

> Is there maybe some double keystroke going on that already triggers the first key event?

Yeah, it's picking up my screenshot key. Sorry for the confusion!

> There also is a "pager function" (`makePagedScroll`).

There is, though it is a little odd. It might be worth looking to switch to it though - genuinely not sure how easy it'd be.

wieselkatze:
> I'd expect it to fill the screen immediately. So like so:
I just assumed it shouldn't fill the whole screen, as that's how textutils.pagedPrint works - unless you _specifically_ make it print _nFreeLines with that argument.
grafik


I'll have a look as to how I can best adjust it.

Issue #525, opened 2020-08-17
Wierd time format


R93950X:
## Useful information to include:
game version: up to 1.15.2 at least
mod version: up to 1.90.2 at least

steps to reproduce: type "time" in the terminal and hit enter at 12:00 AM in game

result 
![image](https://user-images.githubusercontent.com/39017175/90438924-8bded780-e089-11ea-9c9b-9a31da918b97.png)

Proposed solution
since it's a result of the modulo function (12%12 = 0)
an easy fix is to calculate "(TIME-1)%12+1" instead of "TIME%12"
this is like modulo but it instead makes 13 sort of overflow back to 1 instead of 12 to 0

R93950X:
Looking at it more carefully, it's an issue with how the data for os.time() is processed in textutils
since os.time() gives a a number x where 0 ≤ x < 24 hour 0 of one day is actually hour 24 of the previous

two possible solutions, make os.time() a value x where 1≤ x < 25 (not a fan of this one) as it would break existing os.time based programs, as well as skip the number 0 in counting
or instead of using the raw os.time() you can use "(os.time()-1)%24+1" to turn the value 0 into 24 when it is found
which works in this case since every time value besides 0 is in the correct place.

PR #527, opened 2020-08-20
Update textutils.lua

R93950X:
Fix for #525 
Tested, by changing the hour value last second we avoid 12PM existing twice.


SquidDev (changes requested):
Thank you for the fix! Would it be possible to add a test for this to `textutils_spec.lua`?

R93950X:
To be honest I don't fully understand what's going on in there...

PR #533, opened 2020-08-29
Add date-specific MOTDs (like Minecraft)

MCJack123:
This PR adds a few MOTDs from Minecraft that will only be displayed on specific dates. Like Minecraft, they will also be the only MOTDs displayed for that date. The list includes: 
* "Happy new year!" - January 1
* "Merry X-mas!" - December 24
* "OOoooOOOoooo! Spooky!" - October 31

These may be tuned as necessary, since they come from splash text which is shorter than the MOTDs usually are.

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/programs/motd.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/motd.lua
@@ -1,15 +1,24 @@
-local tMotd = {}
+local date = os.date("*t")
+if date.month == 1 and date.day == 1 then
+    print("Happy new year!")
+elseif date.month == 12 and date.day == 24 then
+    print("Merry X-mas!")
+elseif date.month == 10 and date.day == 31 then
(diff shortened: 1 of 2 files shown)


plt-hokusai (approved):
LGTM :+1:

SquidDev (approved):
Thank you for the contribution - sorry for the delay in getting to it.

Wasn't sure if that `stub` call would work, but the tests seem to pass... Well, locally at least - goodness knows what the build will do.

It distresses me that this uses Swedish Christmas dates, but I guess it's only appropriate.

Lemmmy:
Bit late to the party, but re: Swedish Christmas dates - perhaps make it show on both the 24th and 25th?  Best of both worlds, given that the mod is English

SquidDev:
If we want to be obnoxiously British, "Happy Boxing Day" on the 26th instead?

PR #546, opened 2020-09-27
Extend swedish translation

Granddave:
Hi, I've updated the swedish translations based on the english translation file. :slightly_smiling_face:


SquidDev:
Hey, thank you for your contribution! 

Translations are managed through [Weblate](https://i18n.tweaked.cc/) instead, which is then synced to GitHub. This makes it easier to keep track of translations, and see what's outdated and missing.  Would you be able to submit them there instead?

Granddave:
Alright, of course! I should have read the guidelines more clearly, my bad.

Issue #548, opened 2020-10-03
Computer relays Redstone


KriKri98:
## Useful information to include:
 - Minecraft version: 1.16.1
 - CC: Tweaked version: 1.90.1
Problem: Computers relay a already existing redstone signal. 
![CC-tweaked](https://user-images.githubusercontent.com/46320871/94998356-990c4300-05b1-11eb-9c2a-001da2c02e0d.jpg)
-Comparator sends signal strenght of one into the computer, computer relays the signal to the left and right side. Left side is powered by a programm. If i flip the lever ontop signal strenght goes to 15.

PR #549, opened 2020-10-03
Don't propagate adjacent redstone signals for computers

SquidDev:
Minecraft propagates "strong" redstone signals (those from comparators and repeaters) through solid blocks, allowing them to be picked up on the other side. For example:

image

This also happens with computers, which may not always be desired (see #548). It definitely makes working with redstone a little more awkward. This PR changes computers to not propagate redstone - in the above screenshot the two wires would no longer be lit.

This is _technically_ a breaking change<sup>1</sup>, and does mean computers no longer behave consistently with vanilla redstone components. I think it's worth it, so would appreciate people's thoughts.

<sup>1</sup>: In the sense that redstone mechanics change. Not that Mojang don't do that every few Minecraft versions anyway.


Lupus590:
If I want my computer to have a redstone value pass through it I can have the computer do that via a program. In contrast to that, If it happens anyways when I didn't want it then there is very little that I can do to prevent it.

PR #553, opened 2020-10-11
Add color table to docs

Lemmmy:
One of many small improvements to the docs. Adds the colour table from the wiki with all the useful numbers. In the future I'll look into adding some kind of doc/HTML embedding to Illuaminate, so that this massive table doesn't clutter the source files (would also be useful for reusable components). 

This PR also fixes a few tiny things:
- Summary for `pink` was not showing up due to incorrect punctuation
- Hex color for black was wrong
- Mentions bundled cables in .combine, .subtract and .test

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/apis/colors.lua
+++ b/src/main/resources/data/computercraft/lua/rom/apis/colors.lua
@@ -1,136 +1,137 @@
---- The Colors API allows you to manipulate sets of colors.
---
--- This is useful in conjunction with Bundled Cables from the RedPower mod,
--- RedNet Cables from the MineFactory Reloaded mod, and colors on Advanced
--- Computers and Advanced Monitors.
---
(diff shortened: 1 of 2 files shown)


src/main/resources/data/computercraft/lua/rom/apis/colors.lua
```diff
@@ -8,6 +8,127 @@
 -- and it will use the other API, colours which is exactly the same, except in
 -- British English (e.g. @{colors.gray} is spelt @{colours.grey}).
 --
+-- On basic terminals (such as the Computer and Monitor), all the colors are
```
SquidDev, line 11: Might be worth converting this entire thing to be a comment in the style of this?

```lua
--[-

]]
```

Really any non-trivial comment should be like this, but I've never been consistent.

SquidDev (changes requested):

SquidDev (approved):

PR #560, opened 2020-10-21
Fix TBO normalisation issues on old GPUs

Lemmmy:
This patch fixes https://github.com/Zundrel/cc-tweaked-fabric/issues/17 but will need to be tested on as many different cards as possible to ensure it works.


SquidDev:
For those who have a moment to test, [a build can be downloaded from CI](https://github.com/SquidDev-CC/CC-Tweaked/suites/1374796978/artifacts/22554262).

RubenHetKonijn:
The issue is back again :/

Lemmmy:
You were definitely on "texture buffers" renderer when you tested the 1.12 build?

RubenHetKonijn:
I think I had the wrong jar, im going to try again

RubenHetKonijn:
Yes works now.

PR #563, opened 2020-10-28
Add term.getLine()

JakobDev:
This PR adds term.getLine(). As the name says, this function returns the content of a line. It is already part of the window API, but now it's part of the term API too. A possible use case for this would be a program who restores the content of the terminal after closing.

### DIFF
--- a/src/main/java/dan200/computercraft/core/apis/TermMethods.java
+++ b/src/main/java/dan200/computercraft/core/apis/TermMethods.java
@@ -380,4 +380,31 @@ public abstract class TermMethods
         terminal.getPalette().setColour( colour, r, g, b );
         terminal.setChanged();
     }
+
+    /**
+     * Get the contents of a line.
+     *
+     * @param line The line.
+     * @return The contents of the line.
+     * @throws LuaException (hidden) If the terminal cannot be found.
+     * @cc.treturn number The text.
+     * @cc.treturn number The foreground colour.
+     * @cc.treturn number The background colour.
+     */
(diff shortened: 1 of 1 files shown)


SquidDev:
As mentioned in #272, I explicitly don't want this on arbitrary terminal redirects.

The Window API is very much a frame buffer. Other redirects, and (IMO) _term.native_ however shouldn't be treated as such. For instance, things like TRoR (terminal redirection over rednet) don't have an internal buffer - they just send messages on a modem.

I've mentioned this elsewhere, but apparently not on GitHub, but my view is that `getLine` should be used by people redirecting _to_ windows (for instance, to then copy that window to another buffer). It's not designed for people to see what they've already written to the screen.

JakobDev:
Any other suggestions what function we could implement to get the content of the screen?

SquidDev:
No, it's not something I want to support. If you need to get the contents of the screen, use a `window`.

JakobDev:
Why should we keep the content of the screen as a big secret?

Lupus590:
I assume because it better emulates how real computers work.

SquidDev:
@JakobDev As outlined above, it's not that the contents of a terminal are some spooky secret that must be preserved at all costs. The issue that is we have a method on `term.native()`, then it is assumed it appears on all terminal redirects, and _that_ is what is not desirable.

That said, I don't see the need for it. Is there a specific example/need you've got in mind where using a framebuffer won't suffice?

PR #570, opened 2020-11-01
Use blit to draw boxes, add colors.toBlit

### LINKED ISSUE #561: Cleaning up paint(utils)
Just having a flick through paint and paintutils, and a couple of things we could definitely improve:

## paintutils
 - [ ] Redundant type checks: https://github.com/SquidDev-CC/CC-Tweaked/blob/6734af6e4a7ec411e447e0b4fb6066101eb2da8a/src/main/resources/data/computercraft/lua/rom/apis/paintutils.lua#L74-L76
 - [ ] Box drawing should use `term.blit`. We should still probably keep the `term.setBackgroundColour` for "backwards compatibility".
 - [ ] Experiment with using `term.blit` for drawing images. Sadly images are stored using tables of numbers, so we can't blit super efficiently, but worth seeing if it's any faster anyway.

   It occurs to me that this won't work with transparent images, so may not be possible at all.

## paint
 - [ ] Drawing the initial image is slow. We should replace `drawCanvas` with a `term.blit` based version and see if it is any faster.

Lemmmy:
This PR adds tests to some of the paintutils functions and switches to using term.blit() in drawBox and drawFilledBox (closes #561). I haven't benchmarked the performance of the blit version yet.

The updated paintutils function make use of a new function in `colors`, `colors.toBlit`. This converts the given colour to a paint/blit hex code. It does this with math, so it's definitely not as fast as a lookup table, but it will work more gracefully with invalid, non-po2 colours. It could probably be changed to use a lookup table instead if that's preferred, but the window API already has one in there with its own blit conversion stuff (which looks micro-optimised).

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/apis/colors.lua
+++ b/src/main/resources/data/computercraft/lua/rom/apis/colors.lua
@@ -332,3 +332,14 @@ function rgb8(r, g, b)
         return packRGB(r, g, b)
     end
 end
+
+--- Converts the given color to a paint/blit hex character (0-9a-f).
+--
+-- This is equivalent to converting floor(log_2(color)) to hexadecimal.
+--
+-- @tparam number color The color to convert.
+-- @treturn string The blit hex code of the color.
+function toBlit(color)
+    expect(1, color, "number")
+    return string.format("%x", math.floor(math.log(color) / math.log(2)))
+end
--- a/src/main/resources/data/computercraft/lua/rom/apis/colors.lua
+++ b/src/main/resources/data/computercraft/lua/rom/apis/colors.lua
@@ -333,6 +333,12 @@ function rgb8(r, g, b)
     end
 end
 
+-- Colour to hex lookup table for toBlit
+local color_hex_lookup = {}
+for i = 0, 15 do
+    color_hex_lookup[2 ^ i] = string.format("%x", i)
+end
+
 --- Converts the given color to a paint/blit hex character (0-9a-f).
 --
 -- This is equivalent to converting floor(log_2(color)) to hexadecimal.
@@ -341,5 +347,6 @@ end
 -- @treturn string The blit hex code of the color.
 function toBlit(color)
     expect(1, color, "number")
-    return string.format("%x", math.floor(math.log(color) / math.log(2)))
+    return color_hex_lookup[color] or
+        string.format("%x", math.floor(math.log(color) / math.log(2)))
 end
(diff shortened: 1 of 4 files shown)


src/main/resources/data/computercraft/lua/rom/apis/colors.lua
```diff
@@ -332,3 +332,14 @@ function rgb8(r, g, b)
...
+
+--- Converts the given color to a paint/blit hex character (0-9a-f).
+--
+-- This is equivalent to converting floor(log_2(color)) to hexadecimal.
+--
+-- @tparam number color The color to convert.
+-- @treturn string The blit hex code of the color.
+function toBlit(color)
+    expect(1, color, "number")
+    return string.format("%x", math.floor(math.log(color) / math.log(2)))
```
SquidDev, line 344: There is a bit of me which is tempted to do what I've done elsewhere. First declare a look up table (like window.lua) and then do `colour_lookup[color] or string.format(...)`. Not sure if it's worth it, but (should) be much faster in the common case. I should probably benchmark this first.

src/test/resources/test-rom/spec/apis/colors_spec.lua
```diff
@@ -73,4 +73,39 @@ describe("The colors library", function()
...
         expect({ colors.rgb8(0x4c7f99) }):same { 0x4c / 0xFF, 0x7f / 0xFF, 0.6 }
     end)
+
+    describe("colors.toBlit", function()
+        it("validates arguments", function()
+            expect.error(colors.toBlit, nil):eq("bad argument #1 (expected number, got nil)")
+        end)
+
+        it("converts all colors", function()
+            local colorsHex = {
```
SquidDev, line 83: This version is a little shorter 😛.

```lua
local colorsHex = {}
for i = 0, 15 do colorsHex[2 ^ i] = string.format("%x", i) end
```

I guess it's less explicit, but I'm fine with that

SquidDev (approved):
Thank you very much for doing this!

Looks good to me. I've left a couple of comments, but feel free to ignore them.

src/main/resources/data/computercraft/lua/rom/apis/colors.lua
Lemmmy: Probably better to fall back to this, yeah. Will just copy the window table and leave window.lua untouched for the time being (since it's a bit more optimised)

src/test/resources/test-rom/spec/apis/colors_spec.lua
Lemmmy: May as well remove the table entirely at that point

SquidDev (approved):
I'm SquidDev, and I approve of this PR.

PR #571, opened 2020-11-01
use arg[0] in all usage printouts

### LINKED ISSUE #439: Delete referes to itself as rm in its usage
I can easily make a pull request to make it and the other programs use `shell.getRunningProgram`, I'm just wondering if it's worth it

 - Minecraft version: whatever master branch builds
 - CC: Tweaked version: commit 17b7727 on master branch
 - Detailed reproduction steps: on any computer run the command `delete` with no args, note that the printed usage is `rm <paths>` not `delete <paths>`

Lupus590:
closes #439 

I wonder if any of the tests will break, do we have any that check for programs printing out their usage? Guess I'll find out

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/programs/alias.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/alias.lua
@@ -1,7 +1,7 @@
 local tArgs = { ... }
 if #tArgs > 2 then
     local programName = arg[0] or fs.getName(shell.getRunningProgram())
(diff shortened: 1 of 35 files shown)


Lupus590:
I'm not sure how to fix it complaining about `arg` being an unknown global. I also thought that I got all of the whitespace complaints but I'll have another look.

SquidDev:
Oh, that's probably an oversight on my part. I'll get that fixed later today.

Issue #579, opened 2020-11-13
os.epoch is inconsistent with documentation


Gibbo3771:
## Versions
 - Minecraft version - 1.16.3
 - CC: Tweaked version - 1.93.1

## Reproduce

 Pretty straight forward. Run this:

```lua
while true do
    sleep(0.05)
    print(os.epoch("local"))
end
```

Clearly counts up in milliseconds but docs state that it `os.epoch()` returns in seconds. I am not sure if the docs are wrong, or if something else is broken.

[Docs](https://tweaked.cc/module/os.html#v:epoch)

PR #580, opened 2020-11-13
epoch not represented in seconds as per the docs

Gibbo3771:
Not sure if you want to update the docs to reflect milliseconds or update the code to reflect seconds. Regardless, here is a PR changing the milliseconds to seconds.

**Issue**
https://github.com/SquidDev-CC/CC-Tweaked/issues/579

### DIFF
--- a/src/main/java/dan200/computercraft/core/apis/OSAPI.java
+++ b/src/main/java/dan200/computercraft/core/apis/OSAPI.java
@@ -150,7 +150,7 @@ public class OSAPI implements ILuaAPI
 
     private static long getEpochForCalendar( Calendar c )
     {
-        return c.getTime().getTime();
+        return c.getTime().getTime() / 1000;
     }
 
     /**
--- a/src/main/java/dan200/computercraft/core/apis/OSAPI.java
+++ b/src/main/java/dan200/computercraft/core/apis/OSAPI.java
@@ -150,7 +150,7 @@ public class OSAPI implements ILuaAPI
 
     private static long getEpochForCalendar( Calendar c )
     {
-        return c.getTime().getTime() / 1000;
+        return c.getTime().getTime();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Lemmmy:
Changing the behaviour of the API just because the docs are wrong would be a breaking change for no good reason

Gibbo3771:
> Changing the behaviour of the API just because the docs are wrong would be a breaking change for no good reason. Other documentation sources say 'milliseconds', so this is probably just a typo in the docs

I agree. I suspect _a lot_ of people have code that uses this and they are aware it uses milliseconds :D.

src/main/java/dan200/computercraft/core/apis/OSAPI.java
```diff
@@ -380,15 +380,15 @@ public final int day( Optional<String> args ) throws LuaException
     /**
      * Returns the number of seconds since an epoch depending on the locale.
```
SquidDev, line 381:
```suggestion
     * Returns the number of milliseconds since an epoch depending on the locale.
```

src/main/java/dan200/computercraft/core/apis/OSAPI.java
```diff
@@ -380,15 +380,15 @@ public final int day( Optional<String> args ) throws LuaException
...
+     * * If called with {@code ingame}, returns the number of milliseconds since the
      * world was created. This is the default.
-     * * If called with {@code utc}, returns the number of seconds since 1
+     * * If called with {@code utc}, returns the number of milliseconds since 1
      * January 1970 in the UTC timezone.
-     * * If called with {@code local}, returns the number of seconds since 1
+     * * If called with {@code local}, returns the number of milliseconds since 1
      * January 1970 in the server's local timezone.
      *
      * @param args The locale to get the seconds for. Defaults to {@code ingame} if not set.
```
SquidDev, line 390:
```suggestion
     * @param args The locale to get the milliseconds for. Defaults to {@code ingame} if not set.
```

SquidDev (changes requested):
Thank you for doing this! Just spotted a couple more stray "seconds":

Gibbo3771:
> 
> 
> Thank you for doing this! Just spotted a couple more stray "seconds":

Good spot, shoulda done a search and replace in a range of line numbers!

SquidDev (approved):

PR #582, opened 2020-11-17
Clear gets an option to reset the palette

### LINKED ISSUE #555: clear should reset the palette colours
After a program modified the palette colours it can be quite difficult to interact with the console, having to go into the Lua interpreter and resetting the colour palette can be difficult.

I suggest that clear should reset the palette colours in addition to clearing the screen.
If adding that behaviour to the existing clear program is too big of a modification I'd suggest adding a "reset" program that clears the screen and resets the palette. (That would be similar naming to the respective Linux/Unix programs).

This would be especially useful for monitors where the only option to reset the palette colour without doing it via Lua is to break and replace the monitor.

I'd welcome feedback on this idea. If this suggestion is accepted I can create a PR to implement either option.

Luca0208:
Fix #555

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/help/clear.txt
+++ b/src/main/resources/data/computercraft/lua/rom/help/clear.txt
@@ -1,1 +1,6 @@
-clear clears the screen.
+clear clears the screen and/or resets the palette.
+ex:
+"clear" clears the screen, but keeps the palette.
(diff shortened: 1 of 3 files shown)


SquidDev (approved):

SquidDev:
Looks good to me, thank you!

PR #583, opened 2020-11-17
Strict Globals

### LINKED ISSUE #233: Prevent defining environment-level globals in programs
@lupus590 recently added a [strict mode](http://lua-users.org/wiki/DetectingUndefinedVariables) to mbs's shell (SquidDev-CC/mbs@031d3c766354415e4b6bb647691c88af187e7aaa), which raises an error when a program tries to define a global within its environment table.

It'd be nice to integrate something similar into the CraftOS shell. Most of the time defining globals is a bug in the program's code (due to someone missing a `local`). That said, in the interest of backwards compatibility, it's probably worth leaving off for the time being.

I'm also not 100% happy with the error message in the attached PR (sorry Lupus :p), so suggestions are welcome!

Lupus590:
Defaults to off to preserve old behaviour.

Closes #233

### DIFF
--- a/src/main/resources/data/computercraft/lua/bios.lua
+++ b/src/main/resources/data/computercraft/lua/bios.lua
@@ -523,6 +523,13 @@ function os.run(_tEnv, _sPath, ...)
 
     local tEnv = _tEnv
     setmetatable(tEnv, { __index = _G })
+
+    if settings.get("bios.strict_globals", false)  then
+        getmetatable(tEnv).__newindex = function(_, name)
+          error("Attempt to create global " .. tostring(name), 2)
+        end
+    end
+
     local fnFile, err = loadfile(_sPath, nil, tEnv)
     if fnFile then
         local ok, err = pcall(fnFile, ...)
@@ -900,11 +907,6 @@ settings.define("shell.autocomplete", {
     description = "Autocomplete program and arguments in the shell.",
     type = "boolean",
 })
-settings.define("shell.strict_globals", {
-    default = false,
-    description = "Prevents assigning variables into a program's enviroment. Make sure you use the local keyword or assign to _G explicitly.",
-        type = "boolean",
-})
 settings.define("edit.autocomplete", {
     default = true,
     description = "Autocomplete API and function names in the editor.",
@@ -957,6 +959,11 @@ settings.define("lua.function_args", {
 settings.define("lua.function_source", {
     default = false,
     description = "Show where a function was defined when printing functions.",
+    type = "boolean",
+})
+settings.define("bios.strict_globals", {
+    default = false,
+    description = "Prevents assigning variables into a program's enviroment. Make sure you use the local keyword or assign to _G explicitly.",
     type = "boolean",
 })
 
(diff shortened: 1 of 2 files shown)


Lemmmy:
Should default on :japanese_goblin:

src/main/resources/data/computercraft/lua/bios.lua
```diff
@@ -900,6 +900,11 @@ settings.define("shell.autocomplete", {
     description = "Autocomplete program and arguments in the shell.",
     type = "boolean",
 })
+settings.define("shell.strict_globals", {
+    default = false,
+    description = "Prevents assigning variables into a program's enviroment. Make sure you use the local keyword or assign to _G explicitly.",
+        type = "boolean",
```
SquidDev, line 906: The indentation is a little odd here.

SquidDev (changes requested):
I've only [tested this quickly on copy-cat](https://copy-cat.squiddev.cc/gh/pull/583/), but I from what I can tell, it doesn't appear to work.

The trouble is that `os.run` clobbers the metatable, which means that `__newindex` is never used.

https://github.com/SquidDev-CC/CC-Tweaked/blob/d2a1a00dc43e5b65f6b64111ce76dd3db16c919f/src/main/resources/data/computercraft/lua/bios.lua#L520-L525

I think there's two options here:
 - Move this check into `os.run`.
 - Make `os.run` not overwrite the metatable.

Of course, the two are not mutually exclusive, though I am inclined to just go for the first of those. _Technically_ `os.run` is the "correct" way to launch a program (it's what `multishell` etc... do), so I think it makes sense to add it there.

Maybe worth calling it `bios.strict_globals` or something then though?

Lupus590:
Now I'm wondering if I could write a program/API that can toggle the behaviour for the current shell.  While I fix the issues what do you think of having such a program before I start implementing it? It could be useful for people who want strict globals but also depend on programs which are incompatible with it.

Lupus590:
PR is ready to merge, I'm trying to write a program which will toggle the behaviour for the current shell, but it's increasingly looking like I will have to either make a custom shell to disable it in or make code spaghetti. I did manage to make on which would work without multishell though.

src/main/resources/data/computercraft/lua/bios.lua
```diff
@@ -954,6 +964,11 @@ settings.define("lua.function_source", {
     description = "Show where a function was defined when printing functions.",
     type = "boolean",
 })
+settings.define("bios.strict_globals", {
+    default = false,
+    description = "Prevents assigning variables into a program's enviroment. Make sure you use the local keyword or assign to _G explicitly.",
```
SquidDev, line 969:
```suggestion
    description = "Prevents assigning variables into a program's environment. Make sure you use the local keyword or assign to _G explicitly.",
```

Sorry, should have spotted this before.

SquidDev:
Thank you very much!

> TODO: find a spell checker for VSCode that does Lua strings and comments

Emacs has `flyspell-mode` :p.

PR #586, opened 2020-11-21
Remove extra space

Lupus590:
Sorry, I noticed it after you merged.


SquidDev:
Am I going to need to get illuaminate to yell about this too.

Hrmr, actually that's not a bad idea.

PR #588, opened 2020-11-24
Use table.pack so that programs can use arg.n

Lupus590:


SquidDev:
`arg` is present to be compatible with the version present [in standard Lua](https://www.lua.org/manual/5.3/manual.html#7). This does not contain an `n` field, and so there's no reason to add it to the shell.

```lua
~ ᐅ lua -e 'for k, v in pairs(arg) do print(k, v) end'
1       -e
2       for k, v in pairs(arg) do print(k, v) end
0       lua
```

Worth noting that all values in the argument list are non-nil (and in fact are strings), so `#arg` should work as expected.

PR #602, opened 2020-12-10
Roll our own wrapping function for help

SquidDev:
This saves us writing to a buffer multiple times, and so makes things much, much faster.

As in, it's actually usable on copy-cat now!

---

At this point, it might be worth the actual paginator into a separate utility module. Then we could deprecated `textutils.pagedPrint` and point towards that instead - sadly I don't think it's really possible to replicate the behaviour of `_nFreeLines`.

I'm not 100% sure how I feel about exposing entire UI interfaces from a module though. We'll have to see.


Lupus590:
I'm literally writing a word wrapper. I wasn't sure if I would make it a textutils thing or a cc.pretty thing or a new mondule.

Lupus590:
So I copied the code from here and started writing tests. This one fails with `...got "test       "`

`expect(textutils.wrap("test string is long", 11)[1]):eq("test string")`

I would expect it to see the end of the word string with `g` as the 11th character and wrap it there.

src/main/resources/data/computercraft/lua/rom/programs/help.lua
```diff
@@ -13,6 +13,61 @@ if sTopic == "index" then
...
+                        push_line()
+                        space_remaining = width
+                    end
+
+                    current_line = current_line .. sub(word, 1, space_remaining)
+                    word = sub(word, space_remaining + 1)
+                end
+            else
+                -- Print a word normally
+                if width - #current_line - 1 < #word then push_line() end
```
Lupus590, line 52: Remove the -1 and then text wraps how I expect it to in the main discussion.

PR #603, opened 2020-12-11
word wrap utility

Lupus590:
~~I'm also making a cc.pretty version, it will be a separate commit on this PR.~~


Lupus590:
Tests on the pretty version are not passing on my machine, I can't figure out why.

src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -822,3 +822,74 @@ function complete(sSearchText, tSearchTable)
...
+        end
+    end
+
+    push_line()
+
+    -- Normalise the strings suitable for use with blit. We could skip this and
+    -- just use term.write, but saves us a clearLine call.
+    for k, line in pairs(lines) do
+        line = line:sub(1, width)
+        if #line < width then line = line .. (" "):rep(width - #line) end
```
SquidDev, line 890: I think we should drop this line. If people want to pad it afterwards that's fine, but the original code was written for a very specific purpose, and I don't think this is as general.

SquidDev (changes requested):
Well, the build failures due to trunk are moderately annoying. Even forcing a rebuild didn't fix it.

Rather than chucking it in textutils (eww, globals) or `cc.pretty` (not really related to this at all), I wonder if it's worth chucking it into a new module. Not sure about names - `cc.textutils` is probably confusing, so maybe `cc.strings`?

src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
Lupus590: I moved this to a new function in the new module.

src/test/resources/test-rom/spec/modules/cc/strings_spec.lua
```diff
@@ -0,0 +1,42 @@
...
+        it("wraps lines", function()
+            expect(str.wrap("test string is long")[1]):eq("test string is long")
+
+            expect(str.wrap("test string is long", 15)[1]):eq("test string is ")
+            expect(str.wrap("test string is long", 15)[2]):eq("long")
+
+            expect(str.wrap("test string is long", 12)[1]):eq("test string ")
+            expect(str.wrap("test string is long", 12)[2]):eq("is long")
+
+            expect(str.wrap("test string is long", 11)[1]):eq("test string")
```
Lupus590, line 23: This test fails, I can't figure out how to remove the trailing space nicely.

SquidDev: We only take the length of the line into account when appending new words, not spaces. So `test string` is fine, then we see a space (`test string `) and _then_ we add a new line.

The normalisation code above (`:sub(1, width)`) should fix that problem, hence me suggesting keeping that but not the padding with additional spaces :p.

src/main/resources/data/computercraft/lua/rom/modules/main/cc/strings.lua
```diff
@@ -0,0 +1,105 @@
...
+-- This is done by adding spaces or truncating the string.
+--
+-- @tparam string line The string to normalise.
+--
+-- @tparam[opt] number width The width to contrain to, defaults to the width of
+-- the terminal.
+--
+-- @treturn { string... } The wrapped input string.
+-- @usage strings.normalise( "long string", 5 )
+-- @usage strings.normalize( "long string", 5 )
```
SquidDev, line 87: These two examples are the same!

SquidDev (changes requested):
Sorry, I'm very much nitpicking here!

src/main/resources/data/computercraft/lua/rom/modules/main/cc/strings.lua
```diff
@@ -0,0 +1,105 @@
...
+local function normalise(line, width)
```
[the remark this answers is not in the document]
Lupus590: It doesn't just pad though, it also truncates.

src/main/resources/data/computercraft/lua/rom/modules/main/cc/strings.lua
Lupus590: normali**s**e vs normali**z**e

Lupus590: wow, even bolding the s and z doesn't make that ovious

src/main/resources/data/computercraft/lua/rom/modules/main/cc/strings.lua
```diff
@@ -0,0 +1,105 @@
+--- Varous string utilities that might go into textutils if we weren't avoiding
+-- globals.
+
+local expect = require "cc.expect".expect
+
+--- Converts a long string into a table of strings of length shorter than a
+-- given width.
+--
+-- It attemps to split on whitespace and converts \n into the next table entry.
```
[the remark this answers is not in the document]
Lupus590: Is the new version better?

src/main/resources/data/computercraft/lua/rom/modules/main/cc/strings.lua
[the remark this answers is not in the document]
SquidDev: Yeah, gah. `fix_width`, `ensure_width` maybe?

src/main/resources/data/computercraft/lua/rom/modules/main/cc/strings.lua
SquidDev: Fine, under alpha-renaming the two examples are the same! :p. 

I guess my point is that the second doesn't divulge any new information about where the function can be used, and so is that sense isn't very useful.

SquidDev:
Merged manually in 16d74dd2e8709b4df35a826ca053b4a3c332f393. I thought I could push to PR branches, but either something is wrong with my setup or GH removed that feature.

PR #606, opened 2020-12-12
Added Numpad Enter Support - bios.lua

TheWireLord:
Add the ability to use Numpad Enter and have it act just like normal Enter.
(Just like the web-based emulator on the Tweaked.cc wiki)


SquidDev:
Thank you! 

IIRC Javascript doesn't distinguish numpad enter and normal enter, hence the slightly different behaviour on the website.

PR #623, opened 2020-12-25
Add a resolution subcommand to monitor.lua

MCJack123:
This PR adds a special `resolution` command through `monitor.lua` that allows you to set the text scale without going into the Lua REPL. Its syntax is `monitor <side> resolution <n>`, where `<n>` is the text scale from 0.5 to 5.

It doesn't have completion yet, though, as I'm not sure how to implement it in conjunction with the `completion.program` function.


SquidDev:
It's Christmas, go have fun :p (merry Christmas though) . Just a couple of thoughts:

 - Wonder if it's worth calling it `scale` or something. My worry with `resolution` is that people will expect it to take a resolution as an argument instead. 
 - There's an ambiguity here where `monitor left resolution` could either resize the monitor or run the `resolution` program (though obviously in always does the former). Maybe `monitor resolution left` instead?

MCJack123:
I've gone and made the changes you requested, so now it's run with `monitor scale <side> <scale>`.

Lupus590:
> 
> 
> I've gone and made the changes you requested, so now it's run with `monitor scale <side> <scale>`.

I would expect it to be `monitor scale <scale> <side>`.

SquidDev:
> I would expect it to be monitor scale <scale> <side>.

I think `side value` is pretty consistent with other commands. Both `redstone set` and `label set` do that. Only one I can think of which doesn't is `equip`, and I _always_ get that wrong.

SquidDev:
> It doesn't have completion yet, though, as I'm not sure how to implement it in conjunction with the completion.program function.

Only took me 5 months!

Thank you very much, sorry for taking so long to get to this.

PR #632, opened 2020-12-27
Fixed missing argument names in file handle docs

MCJack123:
This PR fixes some instances of argument names that are missing from various `@cc.param` docs for writable text file handles.


SquidDev (changes requested):
Good catch - I really need to get illuaminate to complain about these, it's happened enough times. 

I think in each case the function only accepts one value, it's just we can accept any value (which our argument system doesn't handle very well). So maybe worth using `val` or `str` instead of `...`?

MCJack123:
Ah, I just saw `IArguments` and figured it was a vararg function (I didn't remember it being one before anyway). Fixed.

SquidDev:
Sorry, but could you also revert the `value` -> `value(s)` change?

MCJack123:
Oops, missed that. Fixed.

SquidDev (approved):

SquidDev:
Ta very much!

PR #637, opened 2020-12-30
term.redirect: redirect colo(u)r variants to the other one by default

LoganDark:
This allows redirect objects to provide their preferred spelling only, whether it's `colour` or `color`. This also improves compatibility with redirect objects which already do that (where one spelling will work and the other one will error). See https://github.com/MCJack123/craftos2-rom/pull/8

Since the delegate key is calculated once and stored as an upvalue, this offers only a very minimal performance hit at runtime (only the overhead of a single field access and function call when calling redirected functions, plus the one-time hit when `term.redirect` is first called)

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/apis/term.lua
+++ b/src/main/resources/data/computercraft/lua/rom/apis/term.lua
@@ -42,8 +42,22 @@ term.redirect = function(target)
     for k, v in pairs(native) do
         if type(k) == "string" and type(v) == "function" then
             if type(target[k]) ~= "function" then
-                target[k] = function()
-                    error("Redirect object is missing method " .. k .. ".", 2)
+                if k:sub(-6, -1) == 'Colour' then
+                    local delegate = target[k:sub(1, -7) .. 'Color']
+
+                    target[k] = function(...)
+                        target[delegate](...)
+                    end
+                elseif k:sub(-5, -1) == 'Color' then
+                    local delegate = target[k:sub(1, -6) .. 'Colour']
+
+                    target[k] = function(...)
+                        target[delegate](...)
+                    end
+                else
+                    target[k] = function()
+                        error("Redirect object is missing method " .. k .. ".", 2)
+                    end
                 end
             end
         end
--- a/src/main/resources/data/computercraft/lua/rom/apis/term.lua
+++ b/src/main/resources/data/computercraft/lua/rom/apis/term.lua
@@ -42,21 +42,29 @@ term.redirect = function(target)
     for k, v in pairs(native) do
         if type(k) == "string" and type(v) == "function" then
             if type(target[k]) ~= "function" then
+                local delegate
+                
                 if k:sub(-6, -1) == 'Colour' then
-                    local delegate = target[k:sub(1, -7) .. 'Color']
-
-                    target[k] = function(...)
-                        return target[delegate](...)
-                    end
+                    delegate = target[k:sub(1, -7) .. 'Color']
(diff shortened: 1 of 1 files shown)


LoganDark:
This could potentially cause infinite tail recursion if both variants are missing. Give me a little bit to come up with a solution

Wojbie:
I get the idea behind but implementation really feels weird to me for some reason. I wonder if there is way to implement it in simpler way.

LoganDark:
> I get the idea behind but implementation really feels weird to me for some reason. I wonder if there is way to implement it in simpler way.

Originally I was going to have a map of all color variants to colour variants, but this way it works out automatically.

SquidDev:
I think this is a little too magic for my tastes. While I appreciate it's a little ugly, I don't think the complexity within `term.redirect` is worth it to avoid the additional 4? lines within a redirect buffer's definition.

LoganDark:
> I think this is a little too magic for my tastes. While I appreciate it's a little ugly, I don't think the complexity within `term.redirect` is worth it to avoid the additional 4? lines within a redirect buffer's definition.

You either have to duplicate code entirely, which is prone to divergent behavior, or reassign afterwards. Having to do this is annoying:

image

These are 7 nearly identical lines, and would be even more (21 perhaps) if actually included in the definition.

Just so you know, not everyone is as tolerant as you are, and some people still prefer to keep the alternative spellings out of their codebase. I am one of those people. Whether or not this patch exists, there are still those who will stick to their preference and be blatantly incompatible with programs of "the other side".

An opinionated solution like redirecting all `Colour` variants to `Color` variants (or vice versa) would be less complex, but it would cause a lot of backlash. This is something I would love to get merged because of the aforementioned reasons, so I'm open to other solutions to this problem.

Lemmmy:
The solution is simply not being an ass. First the Unbritishier, and now this? Sounds like somebody is just xenophobic.

LoganDark:
> The solution is simply not being an ass. First the Unbritishier, and now this? Sounds like somebody is just xenophobic.

That... is extremely rude and uncalled for. Please do not use GitHub for personal attacks.

Lupus590:
Would you be open to an alternative implementation of this?
Perhaps two new functions, maybe in a new module?

`britishifyTerm(termObject) : termObject`
`americanifyTerm(termObject : termObject`

LoganDark:
> @SquidDev
> Would you be open to an alternative implementation of this?
> Perhaps two new functions, maybe in a new module?
> 
> `britishifyTerm(termObject) : termObject`
> `americanifyTerm(termObject) : termObject`

At the very least, the method names should be a bit more neutral. British English is in fact not only used by Great Britain, but also... like every other country except for the US. Perhaps something like `colorToColour` and `colourToColor` would be better?

Lupus590:
My first choice was to have them be englishify and americanify, but that implies that Americans don't speak English.

LoganDark:
> My first choice was to have them be englishify and americanify, but that implies that Americans don't speak English.

Hah. The TerraTech approach is to have "English" and "US English". Still, more neutral is better.

Lupus590:
Straying into off-topic:
I could have gone this route instead.
https://static.boredpanda.com/blog/wp-content/uploads/2017/05/mocking-americans-stereotypes-fb.png

LoganDark:
> Straying into off-topic:
> I could have gone this route instead.
> https://static.boredpanda.com/blog/wp-content/uploads/2017/05/mocking-americans-stereotypes-fb.png

British Chinese and American Chinese.

Issue #634, opened 2020-12-28
Add a getType method for the DiskDrive peripheral


GamesArsenal96:
I think adding a getType method for the DiskDrive would be a good addition so we can only allow certain devices through code like: 
Disc
Disk
Computer
Turtle
Pocket
Maybe in the isDiskPresent method it could return the type of disk or make a method on its own

Lupus590:
I assume that disc is for the vanilla and vanilla-like music discs personally I would call those records in a getType method for less ambiguity.

GamesArsenal96:
> I assume that disc is for the vanilla and vanilla-like music discs personally, I would call those records in a getType method for less ambiguity.

Yes definitely

SquidDev:
Curious as to the use-case of "only allow certain devices"?

I don't have any objections, but I am worried that people are going to start relying on `getType() == "record"` or something instead of `disk.hasAudio`, and that's going to blow up once you have non-default `IMedia` implementations.

GamesArsenal96:
> Curious as to the use-case of "only allow certain devices"?
> 
> I don't have any objections, but I am worried that people are going to start relying on `getType() == "record"` or something instead of `disk.hasAudio`, and that's going to blow up once you have non-default `IMedia` implementations.

Well because there is no way to differentiate if a computer, pocket computer, or turtle is in the drive you can only check if the item has data

SquidDev:
Well, I got that far :p. Just curious as to what you're _actually_ trying to do here.

Wojbie:
A good solution would be to just do `getType() == "audio"` instead of `record` and have it be that if `disk.hasAudio` instead of if its record. Solves issue in future new audio media compatibility.

In fact that is exactly what #638 does under the hood it just calls all playable audio sources `record`

PR #638, opened 2020-12-30
Added disk.getType (SquidDev-CC/CC-Tweaked#634)

MCJack123:
This PR implements `disk.getType` as described in #634, but with a few changes:
* Music disks are reported as `"record"` to avoid ambiguity
* When the drive is empty, it reports `"empty"`
* If the type is unable to be determined (can't really think of a reason why this may be, but mods may add something?), it returns `nil`
* The drive method is currently named `getDiskType` to avoid a collision with `IPeripheral.getType()` - maybe this could be changed? (I'm not super experienced with Cobalt's APIs.)

screenshot

Not sure how well-done this is, but it works enough for me in testing.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/peripheral/diskdrive/DiskDrivePeripheral.java
+++ b/src/main/java/dan200/computercraft/shared/peripheral/diskdrive/DiskDrivePeripheral.java
@@ -11,7 +11,11 @@ import dan200.computercraft.api.media.IMedia;
 import dan200.computercraft.api.peripheral.IComputerAccess;
 import dan200.computercraft.api.peripheral.IPeripheral;
 import dan200.computercraft.shared.MediaProviders;
+import dan200.computercraft.shared.computer.items.ItemComputer;
+import dan200.computercraft.shared.computer.items.ItemComputerBase;
 import dan200.computercraft.shared.media.items.ItemDisk;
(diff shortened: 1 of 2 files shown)


src/main/java/dan200/computercraft/shared/peripheral/diskdrive/DiskDrivePeripheral.java
```diff
@@ -194,6 +197,46 @@ public final void ejectDisk()
...
+            return "pocket";
+        }
+        else if ( MediaProviders.get( stack ) != null && MediaProviders.get( stack ).getAudio( stack ) != null )
+        {
+            return "record";
+        }
+        else
+        {
+            return null;
+        }
```
SquidDev, line 237: This should really be a (default) method on `IMedia` (and then overridden on the various `IMedia` implementations) rather than an if-else block here. Probably worth doing this the proper OOP (and more importantly, extensible) way.

SquidDev (changes requested):
I'm leaving this as a review request, as I'd like to change it if/when this gets merged.

However, I'd really like to sort out what #634 is going to be used for first. It's not clear how we (or rather, programs consuming this API) should handle non-ComputerCraft `IMedia` implementations (such as Plethora's neural interface or OC's disks).

PR #639, opened 2020-12-30
add isReadOnly to fs.attributes

Lupus590:


SquidDev:
Thanks Lupus! 

Would you be able to update the documentation too? I guess we should mention it on line 474 and in the return type.

src/main/java/dan200/computercraft/core/apis/FSAPI.java
```diff
@@ -471,16 +471,16 @@ public final Object getCapacity( String path ) throws LuaException
     /**
      * Get attributes about a specific file or folder.
      *
-     * The returned attributes table contains information about the size of the file, whether it is a directory, and
-     * when it was created and last modified.
+     * The returned attributes table contains information about the size of the file, whether it is a directory,
+     * when it was created and last modified, and  whether it is read only.
```
SquidDev, line 475:
```suggestion
     * when it was created and last modified, and whether it is read only.
```

SquidDev (changes requested):

SquidDev:
Thank you!

Issue #504, opened 2020-07-20
Turtles and computers no longer stack due to ids being kept when they are broken


ralphgod3:
I could not find a method of clearing the turtle/computer id from them, this leads to unstackable computers/turtles
So could a way to do this be added into the mod

original feature request that caused this issue: https://github.com/SquidDev-CC/CC-Tweaked/pull/437

 - Explanation of how the feature/change should work:
config option to use the old way of labeling turtles or an ingame command to clear ids of a computer
 - Some rationale/use case for a feature. My general approach to designing new features is to ask yourself
quite a few of my more advanced programs place turtles themselves this leads to some issues when the turtles dont stack anymore.

edit: i need to learn how to spell.

SquidDev:
I guess the cleanest thing todo would be add  recipe which clears a computer of all additional data. So you place a single computer in a crafting table, and get out a computer with no id or label. Fairly sure this is similar to what other mods do.

Lemmmy:
It might be too easy to accidentally wipe a turtle with the crafting method. Obviously it wouldn't happen when crafting with peripherals, because you could only put one in anyway, however its defintiely very easy to do if you have some sort of `turtle.craft()` system going - if you end up missing an item, the turtle is gonna wipe itself as a valid reicpe. It could also feasibly happen with a few misplaced shift-clicks.

Another proposal was letting the computer lose its ID if the drive is empty when being broken, but I can see this causing more complexity. The computer will have to perform a filesystem check on the main thread every time it's broken, so unless there's a way round that, there doesn't seem to be a very clean solution. This one could also lead to somebody losing an ID they care about.

Naheulf:
> I don't think doing "does this computer have files" is a great way of doing it. It makes sense, but I don't think it's the most intuative.
It's not the most intuitive but you can't do that by mistake.

And how about a message in the advanced tool-tip that explicitly said how to do that ? Example: "Remove all files and tags to make this turtle stackable"

ralphgod3:
> I don't think doing "does this computer have files" is a great way of doing it. It makes sense, but I don't think it's the most intuative. You'll have some computers which'll drop with their ID and some which won't, and it may take a while for people to figure out why.
> 
> @ralphgod3 can you explain exactly what you're doing? I'm trying to envision a situation where you'd automatically place turtles and then not want them to keep their files.
> 
> **Edit:** Oh, are you booting them off an adjacent disk? That'd make sense I guess.

[...]

PR #653, opened 2021-01-05
Computers no longer keep IDs 

99-lives:
See #504 for discussion and justification. Happy to continue the discussion in that issue if you aren't convinced.

1. Unlabeled computers and turtles no longer keep their IDs when broken, reverting the behavior introduced in #437.
2. CraftOS programs `edit`, `wget` and `pastebin get` now print a warning to the user after completion, if they are ran on an unlabeled computer.
3. Added a line to the label help doc explaining that unlabeled computers will lose their files.

My warning message is pretty wordy, so I'm happy to hear any other suggestions to slim it down.

Note that the code that displays computer IDs in inventory tooltips is still present, but as far as I'm aware, is now unused. I'm afraid I'm not familiar enough with this project to find and delete that code.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/computer/items/ComputerItemFactory.java
+++ b/src/main/java/dan200/computercraft/shared/computer/items/ComputerItemFactory.java
@@ -19,7 +19,9 @@ public final class ComputerItemFactory
     @Nonnull
     public static ItemStack create( TileComputer tile )
     {
-        return create( tile.getComputerID(), tile.getLabel(), tile.getFamily() );
+        String label = tile.getLabel();
+        int id = label != null ? tile.getComputerID() : -1;
+        return create( id, label, tile.getFamily() );
     }
 
     @Nonnull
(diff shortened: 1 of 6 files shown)


Lupus590:
While I like the changes to the lua programs and think that that is a good idea along with extra docs about making sure that users don't accidentally lose data, I would prefer that there was a config option for preserving IDs.

99-lives:
Are you talking about a server wide config option in `computercraft-server.toml`? If so, the option could become a point of contention on multiplayer servers. But I supposed the same could be said about this whole patch. 

Would it be possible to make it a per-user setting? A command like `/computercraft keepid true` perhaps? This would affect every computer placed by the player, while leaving computers placed by turtles and other players unaffected. Unfortunately I lack the Forge or Java knowledge to implement this (I only created the Java patches using my experience with other languages, I don't actually know Java :laughing: )

Also is there something I can do to fix the license check in gradel? Like I said, I don't know Java.

MCJack123:
> Also is there something I can do to fix the license check in gradel? Like I said, I don't know Java.

Looks like you just need to merge in `mc-1.15.x` into your branch. The CI checks have been updated to check for 2021, but your files were forked while they still were in 2020.

99-lives:
Thanks @MCJack123, that worked.

SquidDev:
Thank you for your PR!

I'm afraid I'm not going to merge this. I really don't want to revert the changes to turtle drops, at least not wholesale. I've put a bit more explanation/rationale for the whole change in https://github.com/SquidDev-CC/CC-Tweaked/issues/504#issuecomment-756459791.

PR #656, opened 2021-01-07
add / to dir in list

Lupus590:
now monochromatic displays are less ambiguous when listing folders and files

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/programs/list.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/list.lua
@@ -21,7 +21,7 @@ for _, sItem in pairs(tAll) do
     if bShowHidden or string.sub(sItem, 1, 1) ~= "." then
         local sPath = fs.combine(sDir, sItem)
         if fs.isDir(sPath) then
-            table.insert(tDirs, sItem.."/")
+            table.insert(tDirs, sItem .. "/")
         else
             table.insert(tFiles, sItem)
         end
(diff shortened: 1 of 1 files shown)


Lupus590:
Not managed to test this yet and got to have food.

SquidDev:
Hrrrmrm. I'm not 100% sure how I feel about this. I don't know if this is just because I'm not used to it, but it does look a little odd.

The new `ls` command

Would appreciate other people's thoughts on this - you can try it out at https://copy-cat.squiddev.cc/gh/pull/656/.

I guess the alternative solution here would be just to use `colours.grey` (or even keep green) on non-advanced terminals. This way we still have some visual indicator. Thoughts?

Lupus590:
Monochromatic green looked too similar to the white on my 1.16 play world, grey could work though.

It's probably unnecessary complication, but I could put it behind a setting.
`list.dir_disabiguator` with valid settings as `none`, `slash`, `colour` and `both`.

Lupus590:
Removing the `term.color` check works and makes it obvious. I guess that things like that should have been changed when pallets were introduced.

Seniorendi:
> Would appreciate other people's thoughts on this
Does not look bad, but i like it more without the /

SquidDev:
Sorry for taking so long to reply to this (yay my terrible mental health and even worse procrastination).

I've pushed a change which makes folders light grey on normal computers. I think that's a happy compromise between this and doing nothing - still not really happy with the trailing slash.

PR #662, opened 2021-01-08
Update edit.lua

Pechu317:
made it so you can change how many spaces a tab will do, default 2. im new to this so i might have missed something plz tell me if i did


Pechu317:
turns out i missed something B) gonna close this and fix it

PR #663, opened 2021-01-08
Allowed changing how many spaces a tab is in edit.lua

Pechu317:
Added a setting to change how many spaces will be created when you hit tab in the edit program. Default is 2 spaces, as per the Lua style guide.

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/programs/edit.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/edit.lua
@@ -532,7 +532,7 @@ while bRunning do
                 else
                     -- Indent line
                     local sLine = tLines[y]
-                    local indent = settings.get("edit.indent")
+                    local indent = settings.get("edit.indentWidth")
                     tLines[y] = string.sub(sLine, 1, x - 1) .. (" "):rep(indent) .. string.sub(sLine, x)
                     setCursor(x + indent, y)
                 end
--- a/src/main/resources/data/computercraft/lua/rom/programs/edit.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/edit.lua
@@ -532,8 +532,9 @@ while bRunning do
                 else
                     -- Indent line
                     local sLine = tLines[y]
-                    tLines[y] = string.sub(sLine, 1, x - 1) .. (" "):rep(settings.get("edit.indent")) .. string.sub(sLine, x)
-                    setCursor(x + 4, y)
+                    local indent = settings.get("edit.indent")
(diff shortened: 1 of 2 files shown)


Pechu317:
im really bad at this

SquidDev (changes requested):
Thank you for your PR!

I'd really recommend making and testing these changes locally before submitting a PR. It helps avoid obvious build problems (syntax errors) as well as giving me some confidence there aren't any bugs in the changes.

In this case, I can spot a couple of small issues:
 - The cursor offset when pressing tab hasn't been changed with the indent, so will _always_ increase by 4.
 - The backspace logic hasn't been touched at all, so that still uses an indent of 4.

LoganDark:
> Default is 2 spaces, as per the Lua style guide.

No. This has already been changed in CC:T from 2 to 4 spaces. The default should stay 4.

Pechu317:
> > Default is 2 spaces, as per the Lua style guide.
> 
> No. This has already been changed in CC:T from 2 to 4 spaces. The default should stay 4.
sorry, i forgot to update that. the default is 4.

Pechu317:
ok im gonna start using local things now so that everything will build right B)

Pechu317:
oh i forgot the backspace logic lemme go do that

Pechu317:
cool everything should be done now

LoganDark:
Also - I'd recommend caching the indent stuff somewhere, like this:

image

I haven't opened another PR because I don't want to compete with this one. However, calculating string.rep for a constant that never changes throughout the execution of the program is a little wasteful.

Pechu317:
...ok

src/main/resources/data/computercraft/lua/rom/programs/edit.lua
```diff
@@ -650,9 +651,10 @@ while bRunning do
                 if x > 1 then
                     -- Remove character
                     local sLine = tLines[y]
-                    if x > 4 and string.sub(sLine, x - 4, x - 1) == "    " and not string.sub(sLine, 1, x - 1):find("%S") then
-                        tLines[y] = string.sub(sLine, 1, x - 5) .. string.sub(sLine, x)
-                        setCursor(x - 4, y)
+                    local indent = settings.get("edit.indentWidth")
+                    if x > indent and string.sub(sLine, x - indent, x - 1) == "    " and not string.sub(sLine, 1, x - 1):find("%S") then
```
SquidDev, line 655: The `"    "` should also be `(" "):rep(indent)`. As @LoganDark says, probably worth hoisting this to the top of the file.

SquidDev (changes requested):
Thank you for updating. Sorry, just one other nitpick.

Issue #636, opened 2020-12-29
Improvement of serializeImpl-function


magiczocker10:
Change the beginning of the serializeImpl-function to the following
![grafik](https://user-images.githubusercontent.com/59317952/103306468-eb70ca80-4a0d-11eb-93e0-0b5f96910f9b.png)

This would allow the user to use the same table within a table without triggering the recursive table errormessage. But a table within it self would still trigger the error.

Without change:
![grafik](https://user-images.githubusercontent.com/59317952/103306591-34288380-4a0e-11eb-8544-22645e5573c9.png)

With change:
![grafik](https://user-images.githubusercontent.com/59317952/103306728-984b4780-4a0e-11eb-8951-c9f1f5e3f2b4.png)

SquidDev:
I'm afraid I'm probably not going to change this. I think my stance is that unserialization should return a semantically identical table<sup>1</sup> to the original one.

That said, I guess `serialize` could accept an option table like `unserializeJSON` does (possibly also an option to specify whether we emit whitespace or not). I'd definitely accept a PR for that.

[1] Well, ignoring metatables. IMO this was a mistake, but can't change it now :).

PR #664, opened 2021-01-08
Add some options for textutils.serialize

magiczocker10:
This pull request adds some small options to change the behavior and output of textutils.serialize.

I have first issued this in #636 and was asked to create a pull request with some requested changes to it.


src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -262,9 +262,16 @@ local g_tLuaKeywords = {
     ["while"] = true,
 }
 
-local function serializeImpl(t, tTracking, sIndent)
-    local sType = type(t)
+local function serializeImpl(t, tTracking2, sIndent, opts)
+    local sType, tTracking = type(t), {}
     if sType == "table" then
+				if opts.advancedComparing then
```
SquidDev, line 268: Instead of copying the table, could we just do `tTracking[t] = nil` at the end of this block (and probably move `tTracking[t] = nil` inside the `else` branch).

src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -275,20 +282,20 @@ local function serializeImpl(t, tTracking, sIndent)
             return "{}"
         else
             -- Other tables take more work
-            local sResult = "{\n"
-            local sSubIndent = sIndent .. "  "
+            local sResult = "{" .. (opts.noWrap and "" or "\n")
```
SquidDev, line 285: I wonder if it's worth hoisting the "separator string" logic into one place. So something like:

```lua
local open, close, equal, comma = "{\n", "\n}", " = ", ",\n"
-- I'm sure there's more ("[ ", " ] = ", etc...), but hopefully gives the right idea
if ops.noWhiteSpace then
  open, close, equal, comma = "{", "}", "=", ","
end
```

I'd personally also merge the `noWhiteSpace` and `noWrap` options. I guess I'm seeing two use cases for `textutils.serialize`:
 - Want the output to be easy to read (e.g. printing, configuration files).
 - Want the output to be compact (e.g. saving large amounts of data to a file).

src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -645,9 +652,17 @@ end
 -- @throws If the object contains a value which cannot be
 -- serialised. This includes functions and tables which appear multiple
 -- times.
-function serialize(t)
+function serialize(t, opts)
     local tTracking = {}
-    return serializeImpl(t, tTracking, "")
+		expect(2, opts, "table", "nil")
+		if opts then
+			field(opts, "noWrap", "boolean", "nil")
```
SquidDev, line 659: Possibly worth switching to snake_case - that's what we do with `unserializeJSON`.

I'd also make all the terms positive, so you'd have `{ wrap = false }` instead of `{ noWrap = true }`. It gets a little confusing if you're specifying options explicitly - the double negative of `{ noWrap = false }` is a little hard to parse :).

SquidDev (changes requested):
Thank you for submitting this, really sorry for the slow review! Just a couple of comments:

src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
magiczocker10: Done

SquidDev (approved):

SquidDev:
Sorry for taking so long to get to this. Thank you very much! 

Doc examples on the website won't work quite yet as copy-cat hasn't updated yet.

PR #668, opened 2021-01-11
term.redirect attempts localisation

Lupus590:
An alternative to https://github.com/SquidDev-CC/CC-Tweaked/pull/637


Lupus590:
Probably not the best title for the PR but I wanted to differentiate it from #637

Lupus590:
There could probably still be some changes to make this look nicer, but I'm not familiar enough with sting manipulation to do them.

SquidDev:
I think this is a little too magic for my tastes. While I appreciate it's a little ugly, I don't think the complexity within `term.redirect` is worth it to avoid the additional 7 lines within a redirect buffer's definition.

PR #672, opened 2021-01-11
Make rightAlt only close menu, never open it.

### LINKED ISSUE #669: Pressing Right Alt enters edit.lua's menu, which makes it difficult to program with a french AZERTY keyboard layout.
## Background
I use a french AZERTY keyboard layout.
For all of the other layouts users out there, some characters that are widely used in programming (such as `{`,`[`, their closing counterparts, and `\`) need a specific key combination to get them written onto the screen (see the layout image below).
This combination uses the `Right Alt` key, to access these special chars (eg. `R-Alt + 5` to get an opening square bracket).

## Issue
However, as seen [in edit.lua's source code](https://github.com/SquidDev-CC/CC-Tweaked/blob/113d5d982fbaa8fe5a940689d4a65a4c7d1dd923/src/main/resources/data/computercraft/lua/rom/programs/edit.lua#L689), the Right Alt key is used to toggle the menu where you can save, run, exit.

So, to write one of those special chars, you need to enter the menu first, then type the combination (which exits the menu and registers the special char in the file you're editing).
This makes editing in-game really annoying, especially when dealing with tables and indexes.

## Repro Steps
Install a French AZERTY layout,
Run the `edit.lua` program on any computer,
Try to write a special char that uses a Right Alt combination.

Below is an example of an AZERTY layout, so you know where to type.
The R-Alt key is labeled "Alt Gr" in this example, and the chars that can be accessed with it are in blue.
image

## Useful information
This issue is present on each branch I could find on this repo (1.13+), and I reproduced it in MC 1.14.4 (CC:T 1.85.1)
I didn't make a PR because I don't have any experience with downloading, building & testing Java projects.

Wojbie:
Fixes #669

In Edit rightAlt was added to handle fact that on some keyboard layout right alt sends both leftCtrl and rightAlt key codes.
obraz
So leftCtrl was opening the menu and rightAlt closed it.
Sadly on keyboards layouts that only send rightAlt this causes it to open the menu instead when user presses it to write special symbol.
obraz
This changes the code to rightAlt acts like paste event and only closes menu when received.
This preserves old functionality while removing unwanted action.

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/programs/edit.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/edit.lua
@@ -687,7 +687,7 @@ while bRunning do
 
             end
 
-        elseif param == keys.leftCtrl or param == keys.rightCtrl or param == keys.rightAlt then
+        elseif param == keys.leftCtrl or param == keys.rightCtrl then
             -- Menu toggle
             bMenu = not bMenu
             if bMenu then
@@ -696,7 +696,12 @@ while bRunning do
                 term.setCursorBlink(true)
             end
             redrawMenu()
-
+        elseif param == keys.rightAlt then
+            if bMenu then
+                bMenu = false
+                term.setCursorBlink(true)
+                redrawMenu()
+            end
         end
 
     elseif sEvent == "char" then


SquidDev:
This seems a reasonable enough solution, thank you!

Though it is probably worth asking if we know what the LGJGL3 change did WRT the original bug. Don't know if @Lignum or anyone else with a German keyboard can check?

PR #675, opened 2021-01-12
optional transience

Lupus590:
I've not quite got it working yet and am a little stumped. I'm going to sleep on it but for now, here's what I have,

closes: https://github.com/SquidDev-CC/CC-Tweaked/issues/504

### DIFF
--- a/src/main/java/dan200/computercraft/core/apis/IAPIEnvironment.java
+++ b/src/main/java/dan200/computercraft/core/apis/IAPIEnvironment.java
@@ -66,6 +66,10 @@ public interface IAPIEnvironment
 
     void setLabel( @Nullable String label );
 
+    Boolean isTransitory();
+
+    void setTransitory( @Nullable Boolean isTransitory );
+
     int startTimer( long ticks );
 
     void cancelTimer( int id );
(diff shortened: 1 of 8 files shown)


99-lives:
When I wrote #653 I put the changes in `src/main/java/dan200/computercraft/shared/computer/items/ComputerItemFactory.java` and `src/main/java/dan200/computercraft/shared/turtle/items/TurtleItemFactory.java`. That's where the changes were made in #437 and I was just reversing them (more or less). See [my commit here](https://github.com/SquidDev-CC/CC-Tweaked/pull/653/commits/97ff934c5c9776a0b857917f7d30b231923fe8bc), it might help.

SquidDev:
Copying from our disccusion on Discord:

> I don't really know what to do about #653, but am kinda avoiding it right now because I'm really not happy with any of the proposed solutions. I think it has to be something people opt in to, rather than some heuristic. Any heuristic is going to be wrong and/or confusing in some cases, and so best avoided.
>
> The trouble is that I personally see the proposed opt-in systems as kinda ugly. It's exposing a bit of the mod that I don't really want to expose. It's just that some people find that bit useful, and so it needs to be done :/.

PR #677, opened 2021-01-13
Fix/Update id.lua

Wojbie:
Resolves concatenate nil error with disk.getID(sDrive) returning nil for non-disk medium. Adds identification for audio and non-disk medium.

Putting it up as draft cause i am unable to capture proper wording for "non-disk medium" and other messages. They feel off to me. Would appreciate someone with better handle on english chipping in.

I tried to keep file in same style as old one despite the fact it could/should be written more cleanly.
Also no tests cause no simple way to simulate peripherals in tests.

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/programs/id.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/id.lua
@@ -32,7 +32,7 @@ else
 
     local id = disk.getID(sDrive)
     if id then
-        print("The disk is #" .. disk.getID(sDrive))
+        print("The disk is #" .. id)
     else
         print("Non-disk data medium")
     end
--- a/src/main/resources/data/computercraft/lua/rom/programs/id.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/id.lua
@@ -13,19 +13,17 @@ if sDrive == nil then
     end
 
 else
-    local bAudio = disk.hasAudio(sDrive)
-    if bAudio then
-        print("Audio medium")
-
+    if disk.hasAudio(sDrive) then
         local title = disk.getAudioTitle(sDrive)
         if title then
-            print("Audio is titled \"" .. title .. "\"")
+            print("Has audio track \"" .. title .. "\"")
+        else
+            print("Has untitled audio")
         end
         return
     end
 
-    local bData = disk.hasData(sDrive)
-    if not bData then
+    if not disk.hasData(sDrive) then
         print("No disk in drive " .. sDrive)
         return
     end
@@ -34,7 +32,7 @@ else
     if id then
         print("The disk is #" .. id)
     else
-        print("Non-disk data medium")
+        print("Non-disk data source")
     end
 
     local label = disk.getLabel(sDrive)
--- a/src/main/resources/data/computercraft/lua/rom/programs/id.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/id.lua
@@ -13,16 +13,32 @@ if sDrive == nil then
     end
 
 else
+    local bAudio = disk.hasAudio(sDrive)
+    if bAudio then
+        print("Audio medium")
+
+        local title = disk.getAudioTitle(sDrive)
+        if title then
+            print("Audio is titled \"" .. title .. "\"")
+        end
+        return
+    end
+
     local bData = disk.hasData(sDrive)
     if not bData then
         print("No disk in drive " .. sDrive)
         return
     end
 
-    print("The disk is #" .. disk.getID(sDrive))
+    local id = disk.getID(sDrive)
+    if id then
+        print("The disk is #" .. disk.getID(sDrive))
+    else
+        print("Non-disk data medium")
+    end
 
     local label = disk.getLabel(sDrive)
     if label then
-        print("The disk is labelled \"" .. label .. "\"")
+        print("Labelled \"" .. label .. "\"")
     end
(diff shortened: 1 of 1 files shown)


SquidDev:
> Also no tests cause no simple way to simulate peripherals in tests.

[Oh how wrong you are](https://github.com/SquidDev-CC/CC-Tweaked/commit/331031be450a2eab062ea7b71d3a98758f57f230). Entirely in document right now though, so probably better to leave that to me :p.

src/main/resources/data/computercraft/lua/rom/programs/id.lua
```diff
@@ -13,16 +13,32 @@ if sDrive == nil then
...
     local bData = disk.hasData(sDrive)
     if not bData then
         print("No disk in drive " .. sDrive)
         return
     end
 
-    print("The disk is #" .. disk.getID(sDrive))
+    local id = disk.getID(sDrive)
+    if id then
+        print("The disk is #" .. disk.getID(sDrive))
```
Lupus590, line 35: Why fetch the id again when you can just get it out of the var `id`?

src/main/resources/data/computercraft/lua/rom/programs/id.lua
```diff
@@ -13,16 +13,32 @@ if sDrive == nil then
     end
 
 else
+    local bAudio = disk.hasAudio(sDrive)
+    if bAudio then
```
SquidDev, line 17:
```suggestion
    if disk.hasAudio(sDrive) then
```

src/main/resources/data/computercraft/lua/rom/programs/id.lua
```diff
@@ -13,16 +13,32 @@ if sDrive == nil then
...
+        print("Audio medium")
```
SquidDev, line 18: Maybe instead of doing "Audio medium", we either display `Has audio track "the title"` or `Unknown audio track`?

Dunno - "Audio medium" is correct, but also feels a bit of a clunky descriptor?

src/main/resources/data/computercraft/lua/rom/programs/id.lua
```diff
@@ -13,16 +13,32 @@ if sDrive == nil then
...
         print("No disk in drive " .. sDrive)
         return
     end
 
-    print("The disk is #" .. disk.getID(sDrive))
+    local id = disk.getID(sDrive)
+    if id then
+        print("The disk is #" .. id)
+    else
+        print("Non-disk data medium")
```
SquidDev, line 37:
```suggestion
        print("Disk has no id")
```

SquidDev (changes requested):

SquidDev:
> Also no tests cause no simple way to simulate peripherals in tests.

Oh, one thing you could do is just stub out the disk API. That's what we've done in other tests.

src/main/resources/data/computercraft/lua/rom/programs/id.lua
Wojbie: Can be other things then disk. Like computer, or pocket, or turtle. Should we really call it a "Disk" then?

src/main/resources/data/computercraft/lua/rom/programs/id.lua
Wojbie: This was exactly my problem with this program... Hmm `Has audio track "the title"` or `Has untitled audio track` ?

src/main/resources/data/computercraft/lua/rom/programs/id.lua
Wojbie: Should i do that with all locals? Cause i will gladly get rid of all of them if we are dropping idea of "educational and stupidly long programs".

src/main/resources/data/computercraft/lua/rom/programs/id.lua
SquidDev: Oh, true. Maybe "Drive does not contain a disk"? I'd say wait for #634, but that's not going to give you a localisable string.

Honestly, not entirely opposed to just doing something like the following in the if statement above. `id` isn't meant to be a diagnostic tool for the whole adjacent drive.

```lua
if not disk.hasData(sDrive) or not disk.getID(sDrive) then
  print("No disk in drive " .. sDrive)
  return
end
```

(don't feel like you have to, I'm fine with doing what you're suggesting here).

src/main/resources/data/computercraft/lua/rom/programs/id.lua
SquidDev: Well, ones we don't reuse. But yes :).

SquidDev: Personal take is that the best way CraftOS programs can be educational is if they are readable and demonstrate good(ish) practice. They don't need to be overly explicit/verbose, as that's what examples are for!

Wojbie:
Honestly it sounds more productive to wait for peripheral testing to get finalized and documented and then roll it out en mass to all tests in separate PR then stubbing one test at time.

Wojbie:
Thanks for help with phrasing. This sounds way better then my first version did.

Probably `id` should be updated again with new methods if #638 gets merged or someone implements #634 in some other way but i will mark it as ready for current state of mod.

SquidDev (approved):

SquidDev:
Thank you!

Issue #689, opened 2021-01-20
New Texture for Crafty Upgrade?


Flexico:
![turtle_crafty_face_16x](https://user-images.githubusercontent.com/5751770/105167499-cb04cf00-5ae6-11eb-8222-3a353c83508a.png)
It's always kinda bugged me that the crafting table upgrade is gray when a crafting table is tan/brown! So I made a little replacement texture that I've been using for a while, and I wanted to see if you would want to include it in the official mod~

Lupus590:
You could make this a Pull Request by editing the texture in the repo, it will save SquidDev some time of having to implement the change themself.

Flexico:
Oh, I've never done that before! Lemme try that~

Flexico:
Oh -- it says I don't have push permission?

Lupus590:
You need to clone the repo, make an edit in the clone and then create a pull request.

Lupus590:
I should have said fork the repo not clone. And make the edit in the fork.

Flexico:
Ok I'll try that~

SquidDev:
[GitHub already have a pretty good guide](https://guides.github.com/activities/forking/) which I'd recommend reading, but basically to give a quick set of instructions specific to CC: Tweaked:

 - Fork the repository and clone _your forked repository_ (using GitHub desktop or the like).
 - Replace the texture at `src/main/resources/assets/computercraft/textures/block/turtle_crafty_face.png` with your texture.
 - Make a commit, and push it up to GitHub.
 - Make a PR!

In this case, I wouldn't worry too much about doing though - I'm happy to apply this manually.

Flexico:
Awesome, thank you! I'll be sure to read up before trying again~

Flexico:
> In this case, I wouldn't worry too much about doing though - I'm happy to apply this manually.
Have you added it in yet? I don't see the new texture in v1.95.3!

Jummit:
I made a quick mock up texturepack that just replaces the texture with the crafting table top:

![image](https://user-images.githubusercontent.com/28286961/112754732-b0b30a00-8fdd-11eb-8557-e3480ff0b65b.png)

Flexico:
Oh heck, that's better than my idea! Awesome~

Flexico:
I notice this hasn't been included in the 1.96.0 release! Is there a reason for that?

Lupus590:
[...]

PR #690, opened 2021-01-20
Mc 1.16.x

Flexico:
[CC-Tweaked-mc-1.16.x.zip](https://github.com/SquidDev-CC/CC-Tweaked/files/5844033/CC-Tweaked-mc-1.16.x.zip)
I think I'm doing this right...

[Edit] WAIT NO NOT 1.15 *facedesk*


Lupus590:
1.15 is the right branch to make the pr against unless it's a 1.16 specific thing.

Also you should say what the PR does.

BTW, it looks like this PR is malformed, I assume that you are trying to do the crafting upgrade retexture.

Flexico:
Yeah, that's what I'm trying to do! Please forgive my ignorance, I've never done this before. X3

SquidDev:
Hey! Thank you for the PR, I'm going to close this one as I'm afraid it's pretty unsalvageable and leave a comment on #689 with some instructions.

Flexico:
Ok, thanks~

Issue #552, opened 2020-10-09
Edit launches scripts in the wrong directory


SquidDev:
When using the "Run" option from `edit`, we save to a file called `/.temp` and run it:

https://github.com/SquidDev-CC/CC-Tweaked/blob/30d35883b83831900b34040f0131c7e06f5c3e52/src/main/resources/data/computercraft/lua/rom/programs/edit.lua#L392-L397

Historically this was fine, but is more problematic now that `require` exists. As the package path relies on the directory the program is stored in, packages may not be found.

There's a couple of possible solutions to this:
 - Save the file in the script's directory iff the directory isn't read only.
 - Write a bootstrap script which then loads the program and uses `cc.require` to bootstrap a new package environment.

Neither are going to be perfect - ideally we'd even be able to fake `getRunningProgram`, but they might work. Ideas welcome!

Luca0208:
> Neither are going to be perfect - ideally we'd even be able to fake getRunningProgram, but they might work. Ideas welcome!

It might be worth taking a look at how [trace](http://www.computercraft.info/forums2/index.php?/topic/27844-trace-simple-stack-traces-for-errors/) deals with this stuff, in [L.47f](https://pastebin.com/CYQp63kb) it modifies the shell in the environment to return the correct value for `shell.getRunningProgram` I imagine something similar would work for this as well.

Although, since multishell is unable to launch a function directly, we'd still need to write to a .temp file with some sort of wrapper around it.

SquidDev:
> In L.47f it modifies the shell in the environment to return the correct value for `shell.getRunningProgram` I imagine something similar would work for this as well.

OK, that's horrific.

SquidDev:
I think the above commit fixes this as much as is really possible for now.

PR #723, opened 2021-02-22
Modify edit.lua run code to display errors/results of execution and handle require.

### LINKED ISSUE #674: Multishell closes tab too early
when I tell the program to get detailed information with the turtle.getItemDetail(1,true) command it crashes with no return or error

Wojbie:
This change in `edit.lua` program code makes it so that if its possible edit will create `.temp` file in directory currently edited file is. This will make require work consistently with how it will work in edited file. If file is in read only location it will be created in root same as it is currently.

Additionally it adds a handler code to code that is run that will display errors and hold tab open until any key is pressed to indicate user seen the results/errors generated by run action.

Touches #552 
Resolves #674 

Note1: I didn't include code to reset palette because it could have distorted results if someone is messing with palette. If they are unable to read error message it is expected they will realize their pallette changes made reading it impossible.
Note2: `]] .. [[` breaking up that literal string at lines 51 and 52 is there to ensure edit is capable to run edit. It is side effect of run handler code that any code that itself contains a level 33 long bracket literal string will not be run properly. But this should be fairly unlikely situation. If anyone got idea for reasonable code that would deal with this issue it would be helpful.

Example:

obraz
obraz

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/programs/edit.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/edit.lua
@@ -63,7 +63,7 @@ end
 if ok and y >= h then
     term.scroll(1)
 end
-term.setCursorPos(1,h)
+term.setCursorPos(1, h)
 if ok then
     write("Program finished. ")
 end
--- a/src/main/resources/data/computercraft/lua/rom/programs/edit.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/edit.lua
@@ -47,30 +47,29 @@ else
     stringColour = colours.white
 end
 
-local runHeader = [[local current = term.current()
-local ok,err = load([===============]] .. [[==================[]]
-local runHandler = [[]===============]] .. [[==================], "@.temp", nil, _ENV)
-if ok then ok,err = pcall(ok,...) end
+local runHandler = [[multishell.setTitle(multishell.getCurrent(), %q)
+local current = term.current()
+local ok, err = load(%q, %q, nil, _ENV)
+if ok then ok, err = pcall(ok, ...) end
 term.redirect(current)
-term.setTextColor(%i)
-term.setBackgroundColor(%i)
+term.setTextColor(term.isColour() and colours.yellow or colours.white)
+term.setBackgroundColor(colours.black)
 term.setCursorBlink(false)
-local _,y = term.getCursorPos()
-local _,h = term.getSize()
-if err then
+local _, y = term.getCursorPos()
+local _, h = term.getSize()
+if not ok then
     printError(err)
 end
-if not err and y >= h then
+if ok and y >= h then
     term.scroll(1)
 end
 term.setCursorPos(1,h)
-if not err then
-    write("Execution finished. ")
+if ok then
+    write("Program finished. ")
 end
 write("Press any key to continue")
 os.pullEvent('key')
 ]]
-runHandler = string.format(runHandler, highlightColour, bgColour)
 
 -- Menus
 local bMenu = false
@@ -114,7 +113,7 @@ local function load(_sPath)
     end
 end
 
-local function save(_sPath, bRun)
+local function save(_sPath, fWrite)
     -- Create intervening folder
     local sDir = _sPath:sub(1, _sPath:len() - fs.getName(_sPath):len())
     if not fs.exists(sDir) then
@@ -126,11 +125,9 @@ local function save(_sPath, bRun)
     local function innerSave()
         file, fileerr = fs.open(_sPath, "w")
         if file then
-            if bRun then file.write(runHeader) end
-            for _, sLine in ipairs(tLines) do
-                file.write(sLine .. "\n")
+            if file then
+                fWrite(file)
             end
-            if bRun then file.write(runHandler) end
         else
             error("Failed to open " .. _sPath)
         end
@@ -320,7 +317,11 @@ local tMenuFuncs = {
         if bReadOnly then
             sStatus = "Access denied"
         else
-            local ok, _, fileerr  = save(sPath)
+            local ok, _, fileerr  = save(sPath, function(file)
+                for _, sLine in ipairs(tLines) do
+                    file.write(sLine .. "\n")
+                end
+            end)
             if ok then
                 sStatus = "Saved to " .. sPath
             else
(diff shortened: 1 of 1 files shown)


Wojbie:
Thanks to @Lupus590 for looking over the prototype and commenting on it.

src/main/resources/data/computercraft/lua/rom/programs/edit.lua
```diff
@@ -47,6 +47,31 @@ else
     stringColour = colours.white
 end
 
+local runHeader = [[local current = term.current()
+local ok,err = load([===============]] .. [[==================[]]
+local runHandler = [[]===============]] .. [[==================], "@.temp", nil, _ENV)
```
SquidDev, line 52: Instead of doing this horror story, could we do something like the following:

```lua
local runHandler = [[
local current = term.current()
local ok, err = load(%q, %q, nil, _ENV)
...
]]
```


And then:

```lua
local str = runHandler:format(table.concat(tLines, "\n"), "@" .. fs.getName(sPath))
```

Probably also worth dropping the `highlightColour`/`bgColour` - I'd just make those white and black respectively.

src/main/resources/data/computercraft/lua/rom/programs/edit.lua
```diff
@@ -47,6 +47,31 @@ else
...
+if ok then ok,err = pcall(ok,...) end
+term.redirect(current)
+term.setTextColor(%i)
+term.setBackgroundColor(%i)
+term.setCursorBlink(false)
+local _,y = term.getCursorPos()
+local _,h = term.getSize()
+if err then
```
SquidDev, line 60:
```suggestion
if not ok then
```

Otherwise this'll trigger if someone returns a value from the program.

src/main/resources/data/computercraft/lua/rom/programs/edit.lua
```diff
@@ -47,6 +47,31 @@ else
     stringColour = colours.white
 end
 
+local runHeader = [[local current = term.current()
+local ok,err = load([===============]] .. [[==================[]]
+local runHandler = [[]===============]] .. [[==================], "@.temp", nil, _ENV)
+if ok then ok,err = pcall(ok,...) end
```
SquidDev, line 53: Stupid nitpick, sorry, but can you add spaces after all the commas. It's consistent with the rest of the code - it's just the linter won't pick up on this.

src/main/resources/data/computercraft/lua/rom/programs/edit.lua
```diff
@@ -47,6 +47,31 @@ else
...
+local _,h = term.getSize()
+if err then
+    printError(err)
+end
+if not err and y >= h then
+    term.scroll(1)
+end
+term.setCursorPos(1,h)
+if not err then
+    write("Execution finished. ")
```
SquidDev, line 68:
```suggestion
    write("Program finished. ")
```

Up to you - IMO it's a little friendlier?

src/main/resources/data/computercraft/lua/rom/programs/edit.lua
```diff
@@ -101,9 +126,11 @@ local function save(_sPath)
     local function innerSave()
         file, fileerr = fs.open(_sPath, "w")
         if file then
+            if bRun then file.write(runHeader) end
```
SquidDev, line 129: Using a boolean here feels a little bit weird. Maybe instead we should accept a callback, and do something like:

```lua
if file then
  write(file)
end
```

Then when calling it do this for the "Save" command:

```lua
save(sPath, function(x)
  for _, sLine in ipairs(tLines) do
    file.write(sLine .. "\n")
  end
end)
```

And this for the "Run" command:

```lua
save(sTempPath, function(file)
  file.write(runHandler:format(table.concat(tLines, "\n"), "@" .. fs.getName(sPath)))
end)

File: src/main/resources/data/computercraft/lua/rom/programs/edit.lua [hunk 6]
```diff
@@ -47,6 +47,31 @@ else
...
+local _,y = term.getCursorPos()
+local _,h = term.getSize()
+if err then
+    printError(err)
+end
+if not err and y >= h then
+    term.scroll(1)
+end
+term.setCursorPos(1,h)
+if not err then
```
SquidDev, line 67: Similarly here

```suggestion
if ok then
```

SquidDev (commented):
> This change in edit.lua program code makes it so that if its possible edit will create .temp file in directory currently edited file is. [...] If file is in read only location it will be created in root same as it is currently.

Ooooh, that is nice. Thank you!

SquidDev (changes requested):
That was meant to be a "Request changes". Woops.

Anyway, most of these are nitpicks - the string format one is probably the most important, as yikes the current thing is a little hard to read :p.

src/main/resources/data/computercraft/lua/rom/programs/edit.lua
Wojbie: I will admit i completely forgot that `%q` option exists in lua.

Wojbie:
Added changes as requested.
Additionally added code to set correct multishell tittle for the running file and tweaked the name/logic for temp file. It was pointed to me that I should prevent possibility for edit to overwrite someone existing code now that in creates temp files in directories other then root.

src/main/resources/data/computercraft/lua/rom/programs/edit.lua
```diff
@@ -390,8 +418,18 @@ local tMenuFuncs = {
...
     end,
     Run = function()
-        local sTempPath = "/.temp"
-        local ok = save(sTempPath)
+        local sTitle = fs.getName(sPath)
+        if sTitle:sub(-4) == ".lua" then
+            sTitle = sTitle:sub(1, -5)
+        end
+        local sTempPath = bReadOnly and sTitle .. ".temp" or fs.combine(fs.getDir(sPath), sTitle .. ".temp")
+        if fs.exists(sTempPath) then
```
SquidDev, line 426: Can we call this `".temp." .. sTitle` instead? Just so the file is hidden.

Alternatively go full `(".temp.%s-%08x"):format(sTitle, math.random(0x7FFFFFFF))`, then then don't worry about overwriting files.

SquidDev (changes requested):

src/main/resources/data/computercraft/lua/rom/programs/edit.lua
Wojbie: I don't this there is actually point it going that deep in this case. First filename you suggested will be way more rare then what i used ;p

SquidDev (approved):

SquidDev:
Thanks!

Issue #658, opened 2021-01-07
A grand-ol' cleanup 


SquidDev:
A continuation of #91 for the modern age. Can't believe we're 550 issues later...

 - [x] Go through all times we request a block update, and see if we can narrow the scope. We don't actually need to mark the chunk as dirty a lot of the times.

   This was, in part at least, responsible for #643 (turtles shouldn't be doing block updates when reading NBT!).
 -  [ ] Clean up "on destroyed" hooks for blocks (i.e. modems). I feel there's a couple of places we could get away with using `removed`. Likewise, could we rebuild the state when validated?
 - [x] Remove all the `m_` prefixes. So petty I know, and going to cause merge conflicts. But currently we're horrifically inconsistent.
 - [x] Investigate using Mojang mappings. Should make the job of Fabric people easier, but a) need to check legal terms and b) see how well FG copes with it.
 - [x] Reformat everything to use single-line braces and no spaces around parens. Undecided on this, but would mean IDEA can cope with switch expressions in a sensible way.
 - [x] Look into what network packets block updates send. I'm fairly sure most blocks (peripherals especially) don't need to send as much now.

Ronan-H:
Hey, would you accept PRs just refactoring random parts of the codebase against this issue? I want to contribute to this project somehow but I'm reluctant about learning all about forge modding because it looks like such a complicated ecosystem.

Just to pick a random example, in [TextBuffer.java](https://github.com/SquidDev-CC/CC-Tweaked/blob/mc-1.15.x/src/main/java/dan200/computercraft/core/terminal/TextBuffer.java), we have:
- lots of duplicated code in constructors and elsewhere. Instead we can use `this()` or create new functions maybe, assuming the constructor overloads are actually being used
- [redundant code](https://github.com/SquidDev-CC/CC-Tweaked/blob/66e42e0817484702d3be7ab880e60246097f7ecb/src/main/java/dan200/computercraft/core/terminal/TextBuffer.java#L94):

```Java
end = Math.min( end, pos + text.length() );
end = Math.min( end, this.text.length ); // overwrites the prev line with no side effects, right?
```

Hope I'm right about those, I'm quite rusty with Java at this point.

Ronan-H:
> I don't think this one is correct - end is reused within the next Math.min call, so we do use the intermediate value. It looks like this ensures end is less than the buffer's width and the text's length.

Oops. That one just looked weird to me at a glance.

[...]

PR #738, opened 2021-03-14
Refactor and add tests for TextBuffer

Ronan-H:
Discussed in #658, I just wanted to dip my toes in and see if I could make a simple contribution to the project.
- Added tests for TextBuffer (~~close to~~ edit: full coverage now)
- Refactored TextBuffer
  - Removed unused methods
  - Some minor simplifications here and there, tried to keep changes unopinionated, e.g. wanted to use `System.arrayCopy` but can't really say that it would actually be more efficient

Feel free to nitpick or discard or whatever. I went a bit wild with the tests and some of them might be kind of redundant (:

### DIFF
--- a/src/main/java/dan200/computercraft/core/terminal/TextBuffer.java
+++ b/src/main/java/dan200/computercraft/core/terminal/TextBuffer.java
@@ -42,11 +42,10 @@ public class TextBuffer
 
     public void write( TextBuffer text )
     {
-        int start = 0;
-        int end = Math.min( start + text.length(), this.text.length );
-        for( int i = start; i < end; i++ )
+        int end = Math.min( text.length(), this.text.length );
+        for( int i = 0; i < end; i++ )
         {
-            this.text[i] = text.charAt( i - start );
+            this.text[i] = text.charAt( i );
         }
     }
 
--- a/src/main/java/dan200/computercraft/core/terminal/TextBuffer.java
+++ b/src/main/java/dan200/computercraft/core/terminal/TextBuffer.java
@@ -12,46 +12,12 @@ public class TextBuffer
     public TextBuffer( char c, int length )
     {
         text = new char[length];
-        for( int i = 0; i < length; i++ )
-        {
-            text[i] = c;
-        }
+        this.fill( c );
     }
 
     public TextBuffer( String text )
     {
-        this( text, 1 );
-    }
-
-    public TextBuffer( String text, int repetitions )
-    {
-        int textLength = text.length();
-        this.text = new char[textLength * repetitions];
-        for( int i = 0; i < repetitions; i++ )
-        {
-            for( int j = 0; j < textLength; j++ )
-            {
-                this.text[j + i * textLength] = text.charAt( j );
-            }
-        }
-    }
-
-    public TextBuffer( TextBuffer text )
-    {
-        this( text, 1 );
-    }
-
-    public TextBuffer( TextBuffer text, int repetitions )
-    {
-        int textLength = text.length();
-        this.text = new char[textLength * repetitions];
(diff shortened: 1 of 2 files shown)


src/main/java/dan200/computercraft/core/terminal/TextBuffer.java
```diff
@@ -61,76 +27,34 @@ public int length()
 
     public String read()
```
SquidDev, line 28: I think we can probably just delete `read()` entirely.

src/main/java/dan200/computercraft/core/terminal/TextBuffer.java
```diff
@@ -61,76 +27,34 @@ public int length()
...
-    {
-        int pos = start;
-        start = Math.max( start, 0 );
-        end = Math.min( end, pos + text.length() );
-        end = Math.min( end, this.text.length );
-        for( int i = start; i < end; i++ )
-        {
-            this.text[i] = text.charAt( i - pos );
-        }
+        this.write( text.toString() );
```
SquidDev, line 50: I think it's probably worth special-casing this one and doing it as an array copy. Just so we can avoid allocating an extra array.

SquidDev (changes requested):
Thank you for this :heart: ! Left a couple of comments, but otherwise looks good.

src/main/java/dan200/computercraft/core/terminal/TextBuffer.java
Ronan-H: Yeah. I think I was confusing usages in the new tests that I wrote for actual usages.. :D

src/main/java/dan200/computercraft/core/terminal/TextBuffer.java
Ronan-H: Updated to be similar to write( String text, int start )

src/main/java/dan200/computercraft/core/terminal/TextBuffer.java
```diff
@@ -12,112 +12,40 @@
...
-    {
-        int pos = start;
         start = Math.max( start, 0 );
-        end = Math.min( end, pos + text.length() );
-        end = Math.min( end, this.text.length );
+        int end = Math.min( start + text.length(), this.text.length );
         for( int i = start; i < end; i++ )
         {
-            this.text[i] = text.charAt( i - pos );
+            this.text[i] = text.charAt( i - start );
```
SquidDev, line 39: Sorry - should have spotted this earlier, but I don't believe this change is quite correct. Before, we had the following behaviour:

```java
buffer.write("Hello, world!", -3);
System.out.println(buffer);
// `lo, world!   '
```

Now we'd get:

```java
buffer.write("Hello, world!", -3);
System.out.println(buffer);
// `Hello, world!'
```

Ronan-H: [Admittedly I wasn't paying enough attention when updating tests](https://github.com/SquidDev-CC/CC-Tweaked/pull/738/commits/ddcc8ee854d68d24b0bd6a927bf0c49d57d6b75b#diff-f89c5b51d7e953968eb16197e4603fb6344906abb7459e0b16c40e27175e797bL107-R80), since I wrote the tests first to prevent such issues... will update later!

Ronan-H: 979b1a8593e85589585291823e4698b76bd0a4f6

Code is the same as before now, just substituting `start + text.length()` for the `end` argument.
+ extra test for the example you gave

SquidDev (approved):

SquidDev:
Thank you very much (and for coping with my nagging reviews)! :heart:

PR #740, opened 2021-03-17
Unit tests for Terminal

Ronan-H:
Let me know if you have any tips for me, I only know the basics of JUnit so I'm kind of making it up as I go along.

Also:
- Should I create an issue for increasing unit test coverage?
- Do you know if there is a plugin or command or anything to auto fix checkstyle errors? Either way, I'll worry about fixing up the formatting later


SquidDev:
Sorry, haven't had a proper look at the code yet, but just to answer your questions:

> Should I create an issue for increasing unit test coverage?

I think this is one of those tasks which doesn't really have a point where you say "we're done", so probably not worth it. Like any code-quality thing, it's more of a "if you're touching this code, try to leave it better".

> Do you know if there is a plugin or command or anything to auto fix checkstyle errors? Either way, I'll worry about fixing up the formatting later

If you're using IntelliJ IDEA, you can [import our code style settings](https://github.com/SquidDev-CC/CC-Tweaked/blob/mc-1.15.x/config/idea/codeStyleSettings.xml) and then run the reformatter - that _should_ fix the checkstyle issues. I'm afraid I'm not aware of a CLI tool to do that though :/.

SquidDev:
Sorry for taking so long to look at this. I've tweaked a couple of things to use Hamcrest, but otherwise looks good.

Thank you very much!

PR #742, opened 2021-03-19
Fix missing `term.setCursorBlink(true)` in edit.lua

Wojbie:
I noticed that it was not restoring cursor blink when mouse click canceled menu.
Turns out it was missing one line.


SquidDev:
Thank you!

Lennart4711:
This bug is still preasent in the advanced turtles edit mode, when clicking to cancele the "save, run exit" menu

Issue #660, opened 2021-01-08
Turtle don't discover peripheral on side after unequipping internal peripheral.


Wojbie:
Reproduction steps:
 - Place turtle next to disk drive.
![obraz](https://user-images.githubusercontent.com/5893536/104012320-e222f300-51af-11eb-9351-f1f1ea39466c.png)
- Turtle sees peripheral.
 - Equip internal peripheral (modem/tool) on side disk drive is.
 - Turtle sees internal peripheral.
 - Unequp internal peripheral.
 - Turtle now don't see any peripherals despite disk drive still being there.

Causing any kind of update will make turtle see peripheral again. Turning turtle / removing or adding disk to disk drive ect.

 - Minecraft version: 1.16.4
 - CC: Tweaked version: cc-tweaked-1.16.4-1.95.0

I was also able to reproduce it on Switchcraft CC:T version on 1.12

Ronan-H:
Able to reproduce in:
- MC 1.15.2
- CC 1.95.3

SquidDev:
Fixed in #743.

Issue #643, opened 2021-01-02
Mining turtle dupe


RoxasLamprunge:
mining turtles seem to duplicate when moved with the carry on mod in valhasia 3  after placing it the turtle is invisible right clicking it will cause a duped turtle to spawn on the block next to the invisible turtle right clicking fast causes a crash. oddly enough after reloading the save i was carrying a turtle instead of a mining turtle. also it dupes what ever is in its inventory when you pick it up

https://user-images.githubusercontent.com/67304114/103455056-6c45f580-4c9e-11eb-93b6-4498acef2a95.mp4

SquidDev:
I've filed a bug on carry on (see above), which should resolve this when fixed. Tracking the cleanup I need to do in #658.

PR #743, opened 2021-03-20
Fixed side input consistency issue when a turtle upgrade is removed

Ronan-H:
Bug report: #660

Turtle side inputs were not being updated when an upgrade was removed.


SquidDev:
Thanks for this PR!

We do have a test for this - it's probably worth updating it to remove the `(required = false)` (actually, none of the tests in there should have that bit, clearly a copy/paste error).

https://github.com/SquidDev-CC/CC-Tweaked/blob/669b6d2d56cb16f7c349c894be819d4ba8e6acbc/src/test/java/dan200/computercraft/ingame/TurtleTest.kt#L8-L9

This fix is correct, but I do have some concerns about how it'll interact with another bug. Currently, `setUpgrade` is called when loading the turtle's NBT. This is _fine_ right now - it's not _technically_ broken, but is not a good idea and has caused an issue in the past (see #643/#658).

However, adding a call to `updateInput` there is definitely going to make things worse. We need to interact with adjacent blocks, who may not have loaded from NBT yet - probably not a good idea!

I'm not proposing you fix that issue - I haven't found a solution yet whose safety/correctness I'm comfortable with - just wanted to explain why I'm not going to merge this quite yet!

Ronan-H:
Hmm. Sounds like too big of an issue for me to understand properly right now, feel free to close this, it would be counter-productive for me to try to fix something with such big scope. Thanks for the info though! I thought of another idea to fix this but if you want `setUpgrade` to be reworked altogether then maybe a small scale fix doesn't make sense.

...probably about time I read some more forge docs :D

SquidDev:
The appropriate bit of #643 is fixed, so this should not have any adverse side effects. The test appears to be passing, so looks good now.

Thank you!

Issue #747, opened 2021-03-23
drawLine sortCoords not working


lilyzeiset:
The following code should draw a line from the bottom-left to top-right
![code](https://user-images.githubusercontent.com/56274881/112103345-fd32bb80-8b7f-11eb-8949-444396027aa2.png)
![result](https://user-images.githubusercontent.com/56274881/112103352-002dac00-8b80-11eb-9c59-2895645603ef.png)

drawLine(1,1,w,h) [correctly] and drawLine(1,h,w,1) [incorrectly] draw the same line.

lilyzeiset:
@Lemmmy

Lemmmy:
This is rather embarassing 😓 I looked over the code a few times even now and I can't see what could've caused this, so guess I should try and wake up a little more first

SquidDev:
Lem, it's 7am. Stop looking at GitHub.

Lemmmy:
I remember when I originally wrote #570, I stared at this code for a VERY long time trying to figure out what it does:

![](https://i.lemmmy.pw/Rg5h.png)

I think that the confusion comes from the fact that the old code only actually attempts to swap the X coordinates, swapping the Y simultaneously. My code swaps both independently (which is the way I think it should work). The line drawing function probably relies on this behaviour implicitly, causing the line to draw upside down?

SquidDev:
Yeah that makes sense. The new logic is fine for boxes, where there isn't a direction, but doesn't work for lines. 

Woops, should have property tested the pre and post refactoring code.

PR #749, opened 2021-03-23
fixed sortCoords for draw functions

lilyzeiset:
https://github.com/SquidDev-CC/CC-Tweaked/issues/747

idk if this is the most clear/best way to solve this but the math is at least correct

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/apis/paintutils.lua
+++ b/src/main/resources/data/computercraft/lua/rom/apis/paintutils.lua
@@ -23,23 +23,15 @@ local function parseLine(tImageArg, sLine)
     table.insert(tImageArg, tLine)
 end
 
--- Sorts pairs of startX/startY/endX/endY such that the start is always the min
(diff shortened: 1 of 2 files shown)


Lemmmy:
Might be worth also adding another test, or modifying this one:

https://github.com/SquidDev-CC/CC-Tweaked/blob/669b6d2d56cb16f7c349c894be819d4ba8e6acbc/src/test/resources/test-rom/spec/apis/paintutils_spec.lua#L58-L69

Just to be 100% sure that something like this wouldn't happen again ;) I wrote the tests before changing the line drawing algorithm, but as we can see, they weren't comprehensive enough.

SquidDev:
Looking at this again, I wonder if a "better" solution is just to revert #570's changes to `drawLine`, and make `sortCoords` `draw(Filled)Box` specific. I think it'd be nicer than having a step within the box drawing functions.

SquidDev:
Thanks!

PR #755, opened 2021-03-28
Use vanilla crafting table as crafty turtle peripheral

### LINKED ISSUE #689: New Texture for Crafty Upgrade?

Jummit:
Fixes #689


SquidDev:
IIRC the Minecraft EULA doesn't allow us to redistribute assets and art from the game, so I'm not sure we can go this route.

It might be possible to update the model to point towards the built-in texture, but I think I want to fix how we handle texture references first.

PR #760, opened 2021-04-11
Just the redstone integrater from plethora.

Lupus590:
I have no idea what I'm doing.


SquidDev:
Definitely open to adding this, just I think this PR is a little dead :).

Lupus590:
Fair. You can always reopen it if I continue this again.

Issue #691, opened 2021-01-20
Server crashes when turtle places an advanced monitor to an already existing monitor array


spr2-dev:
## System info
MC version : 1.16.4
Forge version : 35.1.29
CC:T version : 1.95.2

## Description
In singleplayer, game crashes when a turtle is attempting to place an advanced monitor to an already existing monitor array.
This is the setup I used (turtle places monitor on the top-right hand corner of the array)
![image](https://user-images.githubusercontent.com/77290740/105233680-7a36ba00-5b6a-11eb-869a-075f2e2ab3ab.png)
According to the [crash dump](https://github.com/SquidDev-CC/CC-Tweaked/files/5845266/crash-2021-01-20_21.52.53-server.txt), the issue is server-side, happened on monitor expansion, and seems to be a race condition (oh no).

## Repro steps
Log into your singleplayer world,
Place the monitors as per the screenshot above,
**Make sure you place a block (type doesn't matter, solid by preference) 2 blocks in front of the turtle, so the monitor screen faces the right direction (faces the turtle).**
Make the turtle place the top-right monitor by adding one in its active slot, and calling `turtle.place()`.

## Additional info
The bug was originally discovered by [xXTurnerLP](https://github.com/xXTurnerLP), and I managed to reproduce it. Same MC and CC:T versions were used.

SquidDev:
Similar bugs happen on 1.15, and boy this is fun. The TLDR is that it's a vanilla bug, which I'm sure we'll have to work around in weird and wonderful ways.

The rough code path is as followed:
 - We start ticking the turtle, and prepare to place a block.
 - The monitor is placed.
 - When placed, we attempt to merge ourselves into other monitors. This involves updating the block state of the current block.

So far, so good. The issue is that we're doing all of this _while the turtle is being ticked_. 

As a bit of background, if you add/remove tile entities while another TE is ticking, they get added to the `pendingBlockEntities` list rather than being set in-world (in order to avoid CMEs on the main `blockEntityList` list). When you call `World.getBlockEntity(BlockPos)`, Minecraft is smart enough to check this list of tiles. Well, sometimes at least.

Monitors call `World.setBlock` to update their state with the new adjacent information. This then calls `Chunk.setBlockState` to actually set the block. `Chunk.setBlockState` sets the block, and then checks if this block needs a tile. If such a tile isn't in the world, it creates a new one and sets it.

[...]

PR #778, opened 2021-05-04
Made monitors queue updates when placed by turtles

baeuric:
Proposed fix for #691

### DIFF
--- a/src/main/java/dan200/computercraft/shared/peripheral/monitor/BlockMonitor.java
+++ b/src/main/java/dan200/computercraft/shared/peripheral/monitor/BlockMonitor.java
@@ -6,6 +6,7 @@
 package dan200.computercraft.shared.peripheral.monitor;
 
 import dan200.computercraft.shared.common.BlockGeneric;
+import dan200.computercraft.shared.turtle.core.TurtlePlayer;
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockState;
 import net.minecraft.entity.LivingEntity;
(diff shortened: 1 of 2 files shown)


src/main/java/dan200/computercraft/shared/peripheral/monitor/BlockMonitor.java
```diff
@@ -85,9 +86,12 @@ public void setPlacedBy( @Nonnull World world, @Nonnull BlockPos pos, @Nonnull B
         if( entity instanceof TileMonitor && !world.isClientSide )
         {
             TileMonitor monitor = (TileMonitor) entity;
-            monitor.contractNeighbours();
-            monitor.contract();
-            monitor.expand();
+            if (livingEntity instanceof TurtlePlayer) {
```
SquidDev, line 89:
```suggestion
            if (livingEntity == null || livingEntity instanceof FakePlayer) {
```

Or similar? This should hopefully capture other block placers, which will have the same issue.

It might also be worth adding a quick comment here explaining why we do this. Something along the lines of "Defer the block update if we're being placed by another TE. See #691".

src/main/java/dan200/computercraft/shared/peripheral/monitor/TileMonitor.java
```diff
@@ -526,6 +532,16 @@ private boolean mergeDown()
         return true;
     }
 
+    void updateNeighborsNextTick() {
+        needsUpdate = true;
```
SquidDev, line 536: Do we also need a call to `TickScheduler.schedule( this );`, or does that kinda magically happen due to `onLoad` or similar?

SquidDev (changes requested):
Thank you for looking into this. Just had a couple of questions, but otherwise looks good.

SquidDev:
Oh, would you also be able to retarget this to 1.15.x? I generally prefer to merge things from 1.15.x to 1.16.x than cherry-picking them across back to 1.15.x.

src/main/java/dan200/computercraft/shared/peripheral/monitor/TileMonitor.java
baeuric: I don’t think it’s necessary, seems to work without it.

PR #779, opened 2021-05-04
Defer monitor tile update when placed by another TE

### LINKED ISSUE #691: Server crashes when turtle places an advanced monitor to an already existing monitor array

baeuric:
Fixes #691 & changes requested in #778

### DIFF
--- a/src/main/java/dan200/computercraft/shared/peripheral/monitor/BlockMonitor.java
+++ b/src/main/java/dan200/computercraft/shared/peripheral/monitor/BlockMonitor.java
@@ -20,6 +20,7 @@ import net.minecraft.tileentity.TileEntityType;
 import net.minecraft.util.Direction;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.World;
+import net.minecraftforge.common.util.FakePlayer;
 import net.minecraftforge.fml.RegistryObject;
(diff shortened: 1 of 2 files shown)


SquidDev:
Thank you very much!

PR #782, opened 2021-05-06
retexture of turtles

chipsams:
makes them both cuter and more reminicent of their materials!


SquidDev:
I'm afraid I'm not going to merge this. While I do rather like these textures (the chest is very cute!), and goodness knows everything needs a redo post Jappa textures, turtles are probably the most recognisable thing from CC - I really don't want to make such major changes to it.

chipsams:
> 
> 
> I'm afraid I'm not going to merge this. While I do rather like these textures (the chest is very cute!), and goodness knows everything needs a redo post Jappa textures, turtles are probably the most recognisable thing from CC - I really don't want to make such major changes to it.


aww...
i understand though. can you at least break the line to give a ._. ? if the turtles were like that they'd be cuter, while keeping their distinctive appearance.

Seniorendi:
I really love the texture (:

PR #787, opened 2021-05-14
add trailing / to urls

Lupus590:
it works in the browser but doesn't in actual CC

see: https://discord.com/channels/477910221872824320/477911902152949771/842887808779419658


Lupus590:
Person was not using recent CC.

Which means that copy cat has a bug.

Issue #767, opened 2021-04-17
Port PUC Lua's test suite to CC


SquidDev:
@MCJack123 has been working hard at updating Cobalt (CC:T's Lua runtime) to Lua 5.2. However, we're going to have to be very careful that this update doesn't break any existing (and reasonable) programs. For instance, we'll have to preserve `getfenv` and `setfenv` in some form.

I think one useful step here would be to copy some of Cobalt's test suite into CC:T. They're currently a mix of home-grown things, and [PUC Lua's test suite][puc] - we'll need to extract out what's relevant from there and integrate it.

For instance, some useful things include:
 - `getfenv`/`setfenv` on functions and numbers (stack levels).
 - Coroutines (especially yielding within debug hooks and native functions).
 - Basic behaviour of the standard library
 - Lua 5.2/5.3 features (e.g. `__pairs`, `utf8`, new strings)
 - Behaviour of `os.time` (do we do this already??)

However, I think it's fair to ignore things like:
 - Cobalt specific regressions
 - `debug.getinfo`'s `activelines`
 - Basic language semantics (e.g. Lua's `constructs.lua` or `verybig.lua`).

[puc]: https://www.lua.org/tests/

sir-maniac:
I hope you don't mind my asking...  Want any help with this?

SquidDev:
Yeah, definitely. I'm kinda in a place where I don't really know what approach would be best for this all.

A bit of a prerequisite is to pin down what bits we care about (i.e. what do people use) and what we don't. For instance, ideally we could drop things like `table.maxn` - I don't think anything relies on it, but need to find out.

That said, most of this can probably be deferred for a while - doesn't impact most of the test suite.

sir-maniac:
Okay! :grin: Sorry for not getting back to you very soon.

I haven't spent much time looking at all the historic lua code, but am willing to try to look over it.  I guess a good start would be to review the forum helps sections and github issues.  Any tips for things to look for?

I noticed strftime might nor be up to snuff.  If you haven't already  I could try setting up a fuzzing test to make sure compatibility isn't broken.

SquidDev:
> I noticed time formatting might nor be completely compatible with strftime(). If you haven't already I could try setting up a fuzzing test to make sure compatibility is correct and doesn't regress unnoticed in the future.

Oh, that'd definitely be very cool. There's definitely issues with our time code, and probably also some in `string.format` and the string matching functions, though those may be harder to fuzz.

[...]

PR #790, opened 2021-05-16
Add "cc.expect".range

Lupus590:
Expect a number to be within a specific range


Lupus590:
The fail is not the new tests:

```
[02:13:34] [Server thread/ERROR] [minecraft/MinecraftServer]: java.lang.AssertionError: Unequipping gives a drive - Assertion failed:
Expected drive,
got nil
[02:13:42] [Server thread/WARN] [minecraft/TestLogger]: (optional) turtle_test.unequip_refreshes_peripheral failed. Didn't succeed or fail within 200 ticks```

File: src/test/resources/test-rom/spec/modules/cc/expect_spec.lua [hunk 1]
```diff
@@ -50,4 +50,22 @@ describe("cc.expect", function()
...
+            expect(e.range(1, 1, 1)):eq(1)
+            expect(e.range(2, 1, 3):eq(2)
+
+            expect.error(e.range, 2, 0, 1):eq("number outside of range (expected value to be within 0 and 1)")
+            expect.error(e.range, 0, 1, 2):eq("number outside of range (expected value to be within 1 and 2)")
+        end)
+                    
+        it("fills in min and max if they are nil", function()
+            expect(e.range(1, 1)):eq(1)
+            expect(e.range(2, nil, 3):eq(2)
```
SquidDev, line 65:
```suggestion
            expect(e.range(2, nil, 3)):eq(2)
```

src/main/resources/data/computercraft/lua/rom/modules/main/cc/expect.lua
```diff
@@ -88,7 +88,30 @@ local function field(tbl, index, ...)
...
+-- @throws If the value is outside of the allowed range.
+local function range(num, min, max)
+  expect(1, num, "number")
+  local min = expect(2, min, "number", "nil") or -math.huge
+  local max = expect(3, max, "number", "nil") or math.huge
+  if min > max then
+      error("min must be less than or equal to max)", 2)
+  end
+  
+  if num < min or num > max then
```
SquidDev, line 106:
```suggestion
  if num ~= num or num < min or num > max then
```

This way we also catch when `num` is NaN. Might be worth having a test for this too :/.

src/main/resources/data/computercraft/lua/rom/modules/main/cc/expect.lua
```diff
@@ -88,7 +88,30 @@ local function field(tbl, index, ...)
...
+      error(("number outside of range (expected value to be within %d and %d)"):format(min, max), 3)
```
SquidDev, line 107: So our Java code also has a similar assertion, but uses an error with the following format:

https://github.com/SquidDev-CC/CC-Tweaked/blob/e10e30f82b8f06f1d4efc7e749cbc890e1f94f6c/src/main/java/dan200/computercraft/shared/peripheral/generic/methods/ArgumentHelpers.java#L40

Which is then called with something like `assertBetween(slot, 1, size, "Slot out of range (%s)")`. I wonder if we could do something similar here? So accept an optional message to prefix it with.

It might also be worth including the original value in the error message (like `expected %d to be within %d and %d` - flexible on the format :p).

SquidDev (changes requested):
That error is expected. It basically tests #660, but obviously that fails right now, so is marked as `required = false`. This shouldn't cause the build to fail.

https://github.com/SquidDev-CC/CC-Tweaked/blob/e10e30f82b8f06f1d4efc7e749cbc890e1f94f6c/src/test/java/dan200/computercraft/ingame/TurtleTest.kt#L8

It looks like there's another error below this one:

```
> Task :test

ComputerTestDelegate > modules/cc/expect_spec.lua FAILED
    java.lang.IllegalStateException at ComputerTestDelegate.java:462
```

Looks like there's a syntax error in `expect_spec.lua`, which is why this fails.

src/test/resources/test-rom/spec/modules/cc/expect_spec.lua
```diff
@@ -50,4 +50,22 @@ describe("cc.expect", function()
                 :eq("bad field 'l' (expected string, table or number, got boolean)")
         end)
     end)
+        
+    describe("range", function()
+        it("checks a single type", function()
+            expect(e.range(1, 1, 1)):eq(1)
+            expect(e.range(2, 1, 3):eq(2)
```
SquidDev, line 57:
```suggestion
            expect(e.range(2, 1, 3)):eq(2)
```

Sorry, missed this one!

SquidDev (changes requested):

src/main/resources/data/computercraft/lua/rom/modules/main/cc/expect.lua
Lupus590: I feel like being a bit more explicit that we are testing for NaN would be better, which might be best done but adding a small function for that.

```lua
local function isNan(num)
  return num ~= num
end
```

SquidDev:
> `tostring(nan)` apparently different OSs running lua can't agree on how to write nan and I was hoping that we wouldn't have that problem as we are running on a Java VM

Hrmr, that's definitely a bug in Cobalt. Weird.

Lupus590:
> 
> 
> > `tostring(nan)` apparently different OSs running lua can't agree on how to write nan and I was hoping that we wouldn't have that problem as we are running on a Java VM
> 
> Hrmr, that's definitely a bug in Cobalt. Weird.

It might not be the issue. Is there a way to know which line in the test failed that I'm missing?

Lupus590:
If `doh` works then I'll revert `try strings` and maybe do some other cleanup as needed

SquidDev:
> It might not be the issue. Is there a way to know which line in the test failed that I'm missing?

If you run it locally, `build/reports/tests/test/index.html` includes the full Lua error message. I can see if it's possible to include this in CI builds.

On a side note, you don't need `tostring` when using `%s` - it does that automatically.

Lupus590:
> On a side note, you don't need `tostring` when using `%s` - it does that automatically.

At this point, I'm being paranoid at how `NaN` is being converted to a string, like I said, I'll try to do some cleanup once the build passes.

Lupus590:
> If you run it locally, `build/reports/tests/test/index.html` includes the full Lua error message. I can see if it's possible to include this in CI builds. I do miss Jenkins's test view :(.

Could you have the HTML report pushed to a gist and linked in the CI log?

Lupus590:
YAY! The tests passed! Now to argue with the linter.

SquidDev:
> Could you have the HTML report pushed to a gist and linked in the CI log?

I can either publish it as an artifact 

> Now to argue with the linter.

Sorry :p. In my defence, it's much easier when you run the build steps locally, but I realise getting a CC environment set up is a bit of a pain. Maybe I need to provide GitPod/Codepanel base images (or just make it easier to use our cursed tooling).

Lupus590:
> Sorry :p. In my defence, it's much easier when you run the build steps locally, but I realise getting a CC environment set up is a bit of a pain. Maybe I need to provide GitPod/Codepanel base images (or just make it easier to use our cursed tooling).

I was expecting this to be a much simpler thing than what it's turning out as.
What could be nice would be if the github editor could hock into the linter and run it before you commit. Maybe once the codespace feature gets out of beta.

sir-maniac:
> > `tostring(nan)` apparently different OSs running lua can't agree on how to write nan and I was hoping that we wouldn't have that problem as we are running on a Java VM
> 
> Hrmr, that's definitely a bug in Cobalt. Weird.

Added the following to a Cobalt test, and it seems to pass.  If it is a bug, I agree it's weird.
`
local nan = 0/0;
assert(tostring(nan) == "nan")
assert(("%s"):format(nan) == "nan")
`

Lupus590:
> 
> 
> > > `tostring(nan)` apparently different OSs running lua can't agree on how to write nan and I was hoping that we wouldn't have that problem as we are running on a Java VM
> > 
> > 
> > Hrmr, that's definitely a bug in Cobalt. Weird.
> 
> Added the following to a Cobalt test, and it seems to pass. If it is a bug, I agree it's weird.
> `local nan = 0/0; assert(tostring(nan) == "nan") assert(("%s"):format(nan) == "nan")`

it was the `%d` I think.

sir-maniac:
> it was the `%d` I think.

Okay, that prints '0', because the C printf "%d" only converts from an integer.

Lupus590:
That's another thing that Cobalt needs to change to be more PUC compliment then.

sir-maniac:
> That's another thing that Cobalt needs to change to be more PUC compliment then.

I just tried it in lua5.3
```
> print(string.format("%d", 0/0))
stdin:1: bad argument #2 to 'format' (number has no integer representation)
stack traceback:
	[C]: in function 'string.format'
	stdin:1: in main chunk
	[C]: in ?
```

@SquidDev have a preference in how to handle this?

SquidDev:
For reference:

```
$ lua5.1
Lua 5.1.5  Copyright (C) 1994-2012 Lua.org, PUC-Rio
> =("%d"):format(0/0)
-9223372036854775808

$ lua5.3
Lua 5.3.6  Copyright (C) 1994-2020 Lua.org, PUC-Rio
> =("%d"):format(0/0)
stdin:1: bad argument #1 to 'format' (number has no integer representation)
```

Meanwhile Cobalt returns 0.

sir-maniac:
> Meanwhile  Cobalt returns 0.
Yeah, the main reason for that is the difference in casting. 

`(long)doubleValue` in c just converts the bits to a long, in java, it tries to do the integer equivilent, which isn't possible, so it silently returns 0.

SquidDev:
Yeah, this should be very easy to fix, I think we need to work out what the desired behaviour here is. Probably a discussion for #767.

SquidDev (approved):

SquidDev:
Thanks for bearing with my annoying comments and more annoying CI setup!

> I can see if it's possible to include this in CI builds. 

Right, have added some python scripts which will format the test reports in a nicer way within the build logs, and attempt to attach them to commits where possible. Someone will have to break the build to see if this works in the wild though!

PR #800, opened 2021-05-31
Add __eq metamethod and equals method to the vector api.

fatboychummy:
Nothing much else to say, just adding it here since we were talking about it.

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/apis/vector.lua
+++ b/src/main/resources/data/computercraft/lua/rom/apis/vector.lua
@@ -151,6 +151,15 @@ local vector = {
     tostring = function(self)
         return self.x .. "," .. self.y .. "," .. self.z
     end,
+    
+    --- Check for equality between two vectors.
+    --
+    -- @tparam Vector self The first vector.
+    -- @tparam Vector other The vector to compare to.
+    -- @treturn boolean Whether or not the vectors are equal or not.
+    equals = function(self, other)
+        return self.x == other.x and self.y == other.y and self.z == other.z
+    end,
 }
 
 local vmetatable = {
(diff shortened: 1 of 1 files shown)


src/main/resources/data/computercraft/lua/rom/apis/vector.lua
```diff
@@ -151,6 +151,15 @@ local vector = {
...
         return self.x .. "," .. self.y .. "," .. self.z
     end,
+
+    --- Check for equality between two vectors.
+    --
+    -- @tparam Vector self The first vector to compare.
+    -- @tparam Vector other The second vector to compare to.
+    -- @treturn boolean Whether or not the vectors are equal.
+    equals = function(self, other)
+        return self.x == other.x and self.y == other.y and self.z == other.z
```
SquidDev, line 161: I think it's probably worth doing a type check here too. So something like `getmetatable(other) == vmetatable and ...`. Obviously you'll need to pre-declare `vmetatable`, alas.

SquidDev (changes requested):

src/main/resources/data/computercraft/lua/rom/apis/vector.lua
Wojbie: Issue i would see is that it would then not show a vector transmitted thru modem as equal if it was checking for vmetatable.

SquidDev: Oh, I'm an idiot. We don't need this check as the Lua VM enforces it for us.

SquidDev (approved):

src/main/resources/data/computercraft/lua/rom/apis/vector.lua
Wojbie: Oh right forgot that VM enforces that both have to have same metatable for equality... So no matter what this won't work on vectors transmitted via modem using metatable magic and will have to be called explicitly in that case. Still this is a really nice addition.

PR #804, opened 2021-06-02
Add `http.pastebin` module

SkyTheCodeMaster:
## A quick checklist
 - If you're working on CraftOS, try to write a few test cases so we can ensure everything continues to work in the future. Tests live in `src/test/resources/test-rom/spec` and can be run with `./gradlew check`. - Ran successfully.

This adds a simple pastebin module, so that people won't have to use `shell.run("pastebin get ID file")`, they can use `local paste = pastebin.get("ID")` instead.
It's [documented already](https://skydocs.madefor.cc/misc/pastebin.html), so no need to worry about that.
If there is any changes needed, let me know please!


SkyTheCodeMaster:
Alright, it's finally successful. Here is a little example of the module:
image
https://pastebin.com/9y5DpvAW
https://pastebin.com/8nQTN8bG

SquidDev:
Sorry for taking so long to look at this. It's taken me a while to work out what I really want to do about this and pastebin in general.

At this point, I really don't want to be doing any more than we are already with pastebin. It's _fine_, and the anonymous uploads are a nice feature for people just starting out, but there's far better alternatives out there (gists).

Sorry.

PR #807, opened 2021-06-05
Make craft without arguments craft as many items as possible.

### LINKED ISSUE #806: Craft without arguments doesn't craft as many items as possible
The help says that craft without any arguments will craft as many items as possible, but it just shows the command usage.

fatboychummy:
Fixes #806

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/programs/turtle/craft.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/turtle/craft.lua
@@ -11,9 +11,9 @@ end
 local tArgs = { ... }
 local nLimit = tonumber(tArgs[1])
 
-if not nLimit and tArgs[1] then
+if not nLimit and tArgs[1] ~= "all" then
     local programName = arg[0] or fs.getName(shell.getRunningProgram())
-    print("Usage: " .. programName .. " [number]")
+    print("Usage: " .. programName .. " [all/number]")
     return
 end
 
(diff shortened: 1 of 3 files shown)


fatboychummy:
I'm unable to test the tests currently -- It's quite late, I don't have gradle on this pc yet, and I need to do the sleep thing.

I'm assuming the tests will run in the pr build though, if anything fails I can fix it when I wake up.

SquidDev (changes requested):
Sorry, should have left a comment on the original issue about this: I think a better solution would be that `craft` still displays its usage, but `craft all` crafts as many items as possible. Obviously (and ideally auto-complete) would have to update the docs to reflect this too.

The rationale here being that `craft` can be a destructive action at times, and I'd rather than shell commands didn't do too much damage when run without args.

src/main/resources/data/computercraft/lua/rom/programs/turtle/craft.lua
```diff
@@ -11,9 +11,9 @@ end
 local tArgs = { ... }
 local nLimit = tonumber(tArgs[1])
 
-if not nLimit and tArgs[1] then
+if not nLimit and tArgs[1] ~= "all" then
     local programName = arg[0] or fs.getName(shell.getRunningProgram())
-    print("Usage: " .. programName .. " [number]")
+    print("Usage: " .. programName .. " [all/number]")
```
SquidDev, line 16: As the amount is required, it should be `<number>` now - I think `all|<number>` makes the most sense, or have two separate usage lines.

SquidDev (changes requested):
Thank you! Just one minor nitpick, but otherwise looks good!

src/main/resources/data/computercraft/lua/rom/programs/turtle/craft.lua
fatboychummy: Alright, it's now using `all|<number>`

SquidDev (approved):

SquidDev:
Thank you!

Issue #51, opened 2018-07-14
Some sort of indicator lights in the bottom of the GUI


Lemmmy:
Perhaps in the bottom left of the terminal border? The purpose of it should be to indicate if the computer is on/off, as well as a brief flash when the computer is restarting (as it is currently impossible to tell if a computer rebooted when a startup program draws nothing)

Here's a very crude mockup of what I'm thinking of:
![](https://i.imgur.com/friE68H.png)

It could be green when on, dark grey when off, and ensure a brief flash (0.25s?) when rebooting.

vpontin:
@rolandoislas But would a gradient like this be compatible with the retro-like design of CC?

SquidDev:
There's a bit of me which would like to add more of a GUI to computers (buttons to shutdown (and reboot?) as well as indicator light), but I'm really not sure if it's possible to add it while looking aesthetically pleasing. 

I suspect you'd have to make the bottom fatter, or add some additional sidebar or something. If people have mockups or ideas, I'd love to see them

Lemmmy:
![](https://you.wouldnot.download/a-selective-dory.png)

SquidDev:
I do quite like that. I'm not sure about the gradient on the buttons, though that's definitely me nitpicking.

There's a bit of me which wonders if it'd be better to flip that around, so the power button is on the RHS. Computer monitors _generally_ have their buttons there, so it'd kinda make sense.

I'd almost want to put the label at the top (as that's where it goes in other GUIs), but that's not going to look especially nice. Alas.

Lemmmy:
![](https://you.wouldnot.download/a-fellow-confirmation.png)

Lemmmy:
Here is one with turtles, with smaller buttons etc:

![](https://you.wouldnot.download/an-accused-smoke.png)

It looked very bizarre with the large buttons from before, this looked a little better. This is just slightly smaller than the chest header.

Jummit:
I think the design in the OP looks the best, because its just a little detail instead of the whole bottom bar of the computer being buttons.

1vannn:
I'm sorry for bringing up an old topic, but what about instead of just a power light, perhaps an indicator light, like when it's waiting for something or gets stuck in a loop (also could be waiting). So if a computer is waiting for a user to enter a password, or an event to happen, it's yellow, indicating that the computer is busy and isn't just "available" and when it is "available", it turns green again. So it's multipurpose and an indicator.

[...]

PR #809, opened 2021-06-05
Additional terminal buttons

SquidDev:
Following on from my comments in #51, I had a look at adding a sidebar to computer GUIs. There's a couple of reasons I think some form of this is worth adding:

 - Better discoverability of features. It's not obvious that <kbd>Ctrl+T</kbd> terminates a program, and <kbd>Ctrl+R</kbd> can be used to reboot. Having buttons to do so makes it a little more friendly.
 - More flexibility in the future. This is a more distant plan, but I'd quite like to add a button to upload/download a file (circumventing pastebin) and potentially an in-game docs browser, and we need an easy way to get to them.

While this PR is still a work in progress, the UI is in a state where I'd quite like some feedback on it. Any and all feedback is welcome - UX is not my strong point, so would definitely like this to a) look good and b) be intuitive.

The computer GUI with a new sidebar


Lemmmy:
Assuming the shortcuts Ctrl+T and Ctrl+S/R are hardcoded (if I remember correcrtly, they are), perhaps it's worth adding them to the tooltips for those buttons to increase the discoverability of features.

Something like `Stop the currently running code (hold Ctrl+T)`:
https://github.com/SquidDev-CC/CC-Tweaked/pull/809/files#diff-503dd38b9e07366621b83c3054b514f5c5af590d6c1b43ffb35f676c703ebb74R114-R116

Merith-TK:
This looks great! one suggestion I have, is allow peripherals to define a button to be added, which then gets appended below the power buttons as its own tab? 

also im trying to figure out how complicated this would be to implement in fabric... might take a bit

SquidDev:
> Allow peripherals to define a button to be added, which then gets appended below the power buttons as its own tab?

This would be neat, but possibly not something I want quite yet. Not sure what peripherals would want it for.

> Trying to figure out how complicated this would be to implement in fabric...

It _should_ be pretty simple - everything we're using here is core Minecraft, so the only changes needed should be mappings ones.

Merith-TK:
> It _should_ be pretty simple - everything we're using here is core Minecraft, so the only changes needed should be mappings ones.

keyword, *should*, would love for it to be a simple import,

PR #815, opened 2021-06-07
Add cc.shell.completion `programWithArgs` completion.

Wojbie:
Gives ability to define a way to complete arguments of another program specified via parameter or arguments itself. This should make using `monitor` program simpler as well as simplify adding auto completion to similar programs.
obraz

Also fixes wrong argument definitions in cc.shell.completion file.

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/modules/main/cc/shell/completion.lua
+++ b/src/main/resources/data/computercraft/lua/rom/modules/main/cc/shell/completion.lua
@@ -31,7 +31,6 @@ local completion = require "cc.completion"
 --
 -- @tparam table shell The shell we're completing in.
 -- @tparam string text Current text to complete.
--- @tparam { string... } previous The shell arguments before this one.
 -- @treturn { string... } A list of suffixes of matching files.
 local function file(shell, text)
     return fs.complete(text, shell.dir(), true, false)
@@ -41,7 +40,6 @@ end
 --
 -- @tparam table shell The shell we're completing in.
 -- @tparam string text Current text to complete.
--- @tparam { string... } previous The shell arguments before this one.
 -- @treturn { string... } A list of suffixes of matching directories.
 local function dir(shell, text)
     return fs.complete(text, shell.dir(), false, true)
@@ -78,22 +76,10 @@ end
 --
 -- @tparam table shell The shell we're completing in.
 -- @tparam string text Current text to complete.
--- @tparam { string... } previous The shell arguments before this one.
--- @tparam[opt] boolean completion_space Whether to add a space after the completed item if there is completion for this program.
 -- @treturn { string... } A list of suffixes of matching programs.
 -- @see shell.completeProgram
-local function program(shell, text, previous, completion_space)
-    local results = shell.completeProgram(text)
-    if completion_space then
-        local tCompletionInfo = shell.getCompletionInfo()
-        for n = 1, #results do
-            local sResult = results[n]
-            if sResult:sub(-1) ~= "/" and tCompletionInfo[shell.resolveProgram(text .. sResult)] then
-                results[n] = sResult .. " "
-            end
-        end
-    end
-    return results
+local function program(shell, text)
+    return shell.completeProgram(text)
 end
 
 --- Complete arguments of a program.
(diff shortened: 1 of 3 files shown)


Wojbie:
o.O I didn't touch THAT? Kinda confused about fail.

src/main/resources/data/computercraft/lua/rom/modules/main/cc/shell/completion.lua
```diff
@@ -74,12 +76,41 @@ end
...
+-- @tparam table shell The shell we're completing in.
+-- @tparam string text Current text to complete.
+-- @tparam { string... } previous The shell arguments before this one.
+-- @tparam string|number program Path of program to complete args for, or a number of argument index which specified program.
+-- @tparam number starting Which argument index this program args started at.
 -- @treturn { string... } A list of suffixes of matching programs.
 -- @see shell.completeProgram
-local function program(shell, text)
-    return shell.completeProgram(text)
+local function programArgs(shell, text, previous, program, starting)
```
SquidDev, line 108: I think it might be nicer here if we had `programWithArgs`, which only accepts a "starting" parameter. It's unlikely you're ever going to have a point where you have programArgs which _isn't_ preceded with a program.

src/main/resources/data/computercraft/lua/rom/modules/main/cc/shell/completion.lua
```diff
@@ -74,12 +76,41 @@ end
...
+    local resolved = type(program) == "string" and program or shell.resolveProgram(previous[program])
+    if not resolved then return end
+    local tCompletionInfo = shell.getCompletionInfo()
```
SquidDev, line 111: Nitpick, but `shell.getCompletionInfo()[resolved]` instead? Saves the additional lookup.

SquidDev (changes requested):
Thank you!  This is something I've wanted for a while. Left a couple of comments, but otherwise looks good.

If you merge in/rebase onto mc-1.15.x, the test failure should be fixed.

src/main/resources/data/computercraft/lua/rom/modules/main/cc/shell/completion.lua
Wojbie: Are you thinking `programWithArgs` that completes program initially then args on next argument ones and replaces both calls or keep current `complete.program` then `complete.programWithArgs` that is just provided with starting point pointing at `complete.program` to read the program name from?

Wojbie: It is certainly doable and i can make it work that way no problem.
It would just break current theme of having each function do only one thing. (completes programs, completes directories, files, both etc). 
It would be a case of function that completes a program on first call and then arguments on next `many` calls. If that is way you think it should be working its no more awkward as current monitor completion.

SquidDev: > It would just break current theme of having each function do only one thing. 

Hrrm. I rather view "program and arguments" as a single unit rather than two. Now, I realise that sounds absurd, but conversely "arguments" on its own doesn't make any sense! 

Obviously this whole API is made worse by `build`'s weird quirks (or even, build's `many` quirks). In worked _fine_ at the time, but is proving increasingly inflexible

src/main/resources/data/computercraft/lua/rom/startup.lua
```diff
@@ -98,11 +106,10 @@ shell.setCompletionFunction("rom/programs/rename.lua", completion.build(
     { completion.dirOrFile, true },
     completion.dirOrFile
 ))
-shell.setCompletionFunction("rom/programs/shell.lua", completion.build(completion.program))
```
SquidDev, line 101: An accident, I think?

SquidDev (changes requested):

src/main/resources/data/computercraft/lua/rom/startup.lua
```diff
@@ -81,9 +81,17 @@ shell.setCompletionFunction("rom/programs/monitor.lua", completion.build(
         if previous[2] == "scale" then
             return completion.peripheral(shell, text, previous, true)
         else
-            return completion.program(shell, text, previous)
+            return completion.programWithArgs(shell, text, previous, 3)
         end
-    end
+    end,
+    {
+        function(shell, text, previous)
```
SquidDev, line 88: Yikes, OK. This is ugly D:. Maybe we need a `buildv2` function which does things right :tm:

src/main/resources/data/computercraft/lua/rom/startup.lua
Wojbie: Yep.

SquidDev (approved):
Will let CI run, but looks good. Thank you!

src/main/resources/data/computercraft/lua/rom/startup.lua
Wojbie: There is a risk it would end up getting overbuild for its uses.

PR #816, opened 2021-06-07
Fixed error handling in scale subcommand

MCJack123:
@Wojbie notified me that running `monitor scale` without a scale value (or with an invalid value) will error out due to concatenating `nil`. This PR fixes that.


SquidDev:
With this and #815, today is a day I'm going to be haunted by CC's argument handling :(.

Thank you very much for fixing this though!

PR #817, opened 2021-06-07
Gitpod Support

Merith-TK:
This is nothing major, just would allow more people to contribute should they be unable to install the dev tools on their computer, or even help develop on their tablets or chromebooks (which is viable with the pojav launcher now)

Feel free to suggest edits


SquidDev:
Going to give this a whirl and see if it works. Thank you!

PR #819, opened 2021-06-10
Add very simple markdown support to the help viewer

### LINKED ISSUE #569: Improved help viewer for CraftOS
This is an old issue (dan200/ComputerCraft#212), but the in-game help viewer is ... somewhat lacklustre. I realise it doesn't get much use, but it'd be nice to have something with a few basic features:

 - Pagination/scrolling
 - Section support (for instance, see [ghv](http://www.computercraft.info/forums2/index.php?/topic/27898-ghv-glorified-help-viewer/)).
 - Very basic markdown support (probably only for files with an `.md` extension). Ideally we'd just be able to style sections, bulleted lists (replace `-`/`*` with `•`) and style bold/italics/code slightly differently.

SquidDev:
- Allow help files to use the ".md" suffix, and move changelog/whatsnew
   to use them.

 - When files end with ".md", the "help" program attempts to highlight
   them. This involves:
   - Colour code blocks with a lightGrey background.
   - Replace lists to use bullet points instead of "-"/"*".
   - Colours headings yellow.
   The implementation of this is a bit janky because a) I wrote this and
   b) we need to run this step before text wrapping, but preserve
   colours and section positions over wrapping (thanks to Jack for
   getting this working).

 - Add section navigation to the help viewer, with left/right to move to
   the next/previous section.

Closes #569

## Example:
The new help viewer previewing the changelog


MCJack123:
From testing this in CraftOS-PC using its changelog (which uses more Markdown syntax), code highlighting seems to be a little broken:

image

It gets worse lower in the file:

image

SquidDev:
Oh, good spot. Hrmr, that's very odd - the section parsing code should mean those sorts of runaways don't happen. Will look into it.

MCJack123:
Oh, whoops, forgot that I had left out a closing backtick in that file. (I'd fixed it locally when testing my initial version, but it never got fixed upstream.) Still a good idea to not flip out if there is an error.

SquidDev:
I feel there was something else I wanted to add to this, but can't for the life of me remember what it was. It'll do for now!

Wojbie:
> 
> 
> I feel there was something else I wanted to add to this, but can't for the life of me remember what it was. It'll do for now!

Was it clickable links? Or i am remembering different conversation?

SquidDev:
CC's docs don't really have links, so not right now. But might be nice to have in the future!

PR #829, opened 2021-06-16
-

ghost:


Wojbie:
Hmm.. I wonder if that is something we need as another argument. Could it not just be inferred from headers user sets? Like having it infer it from `http.websocket("ws://some.server", {["Sec-WebSocket-Protocol"]="subprotocol"})` header.

MDMCK10:
> 
> 
> Hmm.. I wonder if that is something we need as another argument. Could it not just be inferred from headers user sets? Like having it infer it from `http.websocket("ws://some.server", {["Sec-WebSocket-Protocol"]="subprotocol"})` header.

maybe, but in my opinion this approach of making it a separate argument looks cleaner than inferring from the headers, but if @SquidDev thinks it should be changed, I could do that.

SquidDev:
Sorry for taking so long to look at this.

Having had a look at what other implementations do, definitely agree with you that it should be a separate argument that extracted from the header. That said, I do think it'd be better if we start accepting arguments as tables as well. I'm happy to look into that if you'd prefer - I know getting argument parsing code is awkward.

MDMCK10:
Sorry for the late response - it seems that i forgot to enable notifications for this, oops.
Anyway, based on the feedback I got, I decided to scrap the idea of having subprotocol be another argument, and now it is instead just set based on if the user provides the Sec-WebSocket-Protocol header in the `headers` argument or not (default is still null like it used to be, but if the user specifies it, it gets set to whatever the user specified).

also hopefully i merged everything correctly - i hate working with the git command line sometimes

MDMCK10:
As for accepting arguments as tables - I don't really have much of a clue how I would go about doing that, so feel free to look into it if you want to, but it's not something I'll be doing in this PR.

SquidDev:
Thanks! I've applied this to mc-1.15.x manually, as I do want it available for both versions.

PR #833, opened 2021-06-18
fix #832

### LINKED ISSUE #832: wget fails when downloading empty files
When downloading an empty file, wget prints "Success" but fails to actually save the file. This is because `readAll()` returns `nil` when reading empty files:

```lua
local contents = http.get("https://raw.githubusercontent.com/xiondisc/CC-Test/5c4648b5bba5630953e3d09f08bc199fd15bf69c/TurtleTest/turtleTest.lua", nil, true)
print(contents.readAll()) -- nil!
contents.close()
```

This is correct behaviour (judging by what PUC Lua does), so we should fix this on the `wget` side.

Lupus590:


SquidDev:
> What would you suggest to print out if the responce.readAll() is nil?
>
> It should print out Downloaded as xxx, and write the file :p.

Please :)

PR #840, opened 2021-06-19
File uploading

SquidDev:
This implements very basic file uploading for computers. One may drop files onto the Minecraft window while a computer/turtle GUI is open to upload files to the root of the computer's file system.

 - [x] Basic file uploading
 - [x] Confirmation flow for files which would be overwritten
 - [x] Improve user interface

UX and UI for this is pretty terrible right now. I had a fiddle with designing a GUI for this and gave up, meaning we're just using Minecraft's default alert boxes. Any help in improving these would be appreciated:

A dialogue shown after a file is successfully uploaded
The dialogue shown if a file fails to upload

We should probably provide more helpful feedback messages too, saying what files are uploaded.


SquidDev:
Will possibly flesh this issue out a bit more tomorrow - heading off right now!

PR #849, opened 2021-07-05
Add os.nanoTime

### LINKED ISSUE #818: Supplying high-quality entropy to programs
Many programs often need to generate high-quality random numbers for use in dealing with secret data. Programs made for CC:T, which currently doesn't provide entropy, end up either naively calling `math.random` ([even after having it seeded with the system clock](https://github.com/BTCTaras/kristwallet/blob/master/kristwallet#L375-L385)) or making use of some dubious unpredictability in low-resolution timers, events, or low-quality PRNGs.

I see 3 possible approaches to solving this:

- Providing a high-level random bytes and numbers API. This minimizes problems of incorrect usage universally.
- Providing a low-level random bytes API that samples from places like `/dev/urandom`. This would require user-made libraries for converting bytes to numbers and take into account the subtle biases in the conversion.
- Providing a high-resolution timer. Clock jitter RNGs [are a thing](http://www.chronox.de/jent.html), but have also been regarded as [possibly riskier](https://github.com/rust-random/rand/issues/699) than raw OS randomness. This would be the least ad hoc solution and is even already implemented in CCEmuX. It would also require user-made libraries that deal with the even subtler biases.

Finally, one must also question whether the low-ish security environment of a Minecraft Computer even warrants such action. There's already precedent for changes like these in CC addons as well as OC modules. Personally, I think that even if the environment isn't that safe (from side channels to hackers and malicious server admins), solving this would shrink the attack surface for future software, as well as reduce uncertainty in the lives of some security library authors and users.

migeyel:
This *should* be enough to close #818, though this is definitely the "interesting" way of doing it.

I'm very on the fence between this and adding an interface into a `SecureRandom` instance. While this feels like the natural-ish function to have in `os`, it still has some small details that I feel are inelegant for a standard library function. Meanwhile adding `SecureRandom` just feels wrong even though it's the "boring" (and thus desirable) way of doing it.

This adds a new function `os.nanoTime` which behaves exactly like the Java `System.nanoTime`, except that it overflows on 2^53 instead of 2^64 to maintain full precision as a double. This inherits all the Java edge cases as well as needing a modular reduction for properly compute a duration across timestamp overflows (Java's docs uses the `long`'s overflow as a hidden modular reduction).


SquidDev:
Sorry for the slow response - moved house this weekend, and so that's rather been dominating my life of late.

While I understand the motivation, I'm a bit iffy about exposing this sort of information to CC right now. It's a bit fourth-wall breaky and does feel like a bit of an awkward solution, so am not going to merge this.

PR #857, opened 2021-07-10
Fix & enhance some docs

SkyTheCodeMaster:
This pull request fixes some grammar, examples, and adds the colours to blit table to the colours API.
Fixed:
* Added `boolean` to list of transmittable types in `modem.transmit`.
* Closed the file handle in second example of `fs.open()`.
* Added the colours to blit table in the Colours API.
* Fixed a typo (`Read an string`) in the `read` documentation.


SquidDev:
Sorry for the slow response - moved house this weekend, and so that's rather been dominating my life of late.

Thank you very much - good spots here!

Would you be able to revert the changes to colours for now? - I think I don't want to duplicate this table right now until we've got a better approach for handling the whole API. The page basically just redirects you to `colors`, so I don't think it's the end of the world.

SquidDev (approved):

PR #859, opened 2021-07-11
Fix typo in turtle.craft documentation

emiliskiskis:
Description wrote "to craft sticks, slots should contain sticks", whereas it should say "planks" in the second part.

## A quick checklist

### DIFF
--- a/doc/stub/turtle.lua
+++ b/doc/stub/turtle.lua
@@ -1,7 +1,7 @@
 --[[- Craft a recipe based on the turtle's inventory.
 
 The turtle's inventory should set up like a crafting grid. For instance, to
-craft sticks, slots 1 and 5 should contain sticks. _All_ other slots should be
+craft sticks, slots 1 and 5 should contain planks. _All_ other slots should be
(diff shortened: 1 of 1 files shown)


SquidDev:
Thanks!

SquidDev:
`turtle.craft` lives in `rom/turtle/turtle.lua`:

https://github.com/SquidDev-CC/CC-Tweaked/blob/347affcc5c2ac35e531afec0a1fe8765ed2c652c/src/main/resources/data/computercraft/lua/rom/apis/turtle/turtle.lua#L18-L20

Which then just calls the method in `CraftingTablePeripheral`:

https://github.com/SquidDev-CC/CC-Tweaked/blob/347affcc5c2ac35e531afec0a1fe8765ed2c652c/src/main/java/dan200/computercraft/shared/turtle/upgrades/CraftingTablePeripheral.java#L42-L47

emiliskiskis:
Many thanks for the fast response. I was just checking if I needed to update the Javadoc.

PR #862, opened 2021-07-18
Fix typo in paintutils documentation

Angalexik:
The documentation says the y position argument for `drawImage` is called `xPos`, but it should say `yPos`.

## A quick checklist


SquidDev:
Thanks!

PR #863, opened 2021-07-18
Fix `parallel.*` hanging when called with no args

MCJack123:
This PR fixes an issue where calling either `parallel` API function with no arguments causes a hang. Here's a demonstration of this issue:

Note: The current solution errors if no function is specified; perhaps it would be better to silently return?


MCJack123:
I added a bunch of tests for `parallel` since I realized there weren't any there before. It's erroring in `ComputerTestDelegate` for some reason, though, and I'm not sure what that's about.

SquidDev:
> Note: The current solution errors if no function is specified; perhaps it would be better to silently return?

I think so, yes. This is the sort of behaviour you see in other libraries (e.g.  `asyncio.gather()` in Python or `Promise.all([]) in JS`).

```lua
expect.error(parallel.waitForAny, function(), 2):eq("bad argument #2 (expected function, got number)")
```

On line 7. `function(), `s is most definitely not valid Lua :p. Unfortunate that my test report post-processor doesn't correctly parse those errors, will have to see what I can do.

MCJack123:
> ```lua
> expect.error(parallel.waitForAny, function(), 2):eq("bad argument #2 (expected function, got number)")
> ```
> 
> On line 7. `function(), `s is most definitely not valid Lua :p. Unfortunate that my test report post-processor doesn't correctly parse those errors, will have to see what I can do.

Definitely did not just wait for Gradle to run for 10 minutes before catching this :P. The test report parser did catch this though:

> ```
> ## src/main/java/dan200/computercraft/core/lua/BasicFunction.java:48: dan200.computercraft.core.ComputerTestDelegate.apis/parallel_spec.lua failed: java.lang.IllegalStateException: /test-rom/spec/apis/parallel_spec.lua:7: unexpected symbol near ','
> ```

SquidDev:
> The test report parser did catch this though:

It did pick it up, but put the annotation in the wrong place. Just pushed 74752c561c1242a016f95fdf1c2d68bae4e4a757, which should fix that.

SquidDev:
Probably worth swapping out to `table.insert` - what's happening here is that `entries[#entries + 1]` is evaluated before yielding, and so the table hasn't grown at this point.

SquidDev (changes requested):
Sorry, think this got missed in the other stuff:

> Note: The current solution errors if no function is specified; perhaps it would be better to silently return?

I think so, yes. This is the sort of behaviour you see in other libraries (e.g. `asyncio.gather()` in Python or `Promise.all([])` in JS).

SquidDev (approved):

Issue #847, opened 2021-07-04
Turtle NBT upgrade data are loss when you break turtle


SirEdvin:
### Minecraft Version

1.16.x

### Version

1.97.0

### Details

So, I am trying to store some Upgrade data in NBT, and when I break and place turtle back, everything is vanish. Not sure that this is how it should be.

SquidDev:
> Not sure that this is how it should be.

No, neither.

SquidDev:
I think I'm going to consider this as "working as intended". However, we should look into an alternative API for storing data about the upgrade which _will_ be persisted across breaking/placing.

SirEdvin:
Thanks. If we speak about `use cases`, current usecase is warp point storing. 

My turtle upgrade provide ability to store position points with labels and then warp to them. I cannot save this points to computer file system, so I should store it on nbt data in upgrade.

PR #873, opened 2021-07-28
Allows to add new capabilities to turtle by upgrade

SirEdvin:
You can see usage example here: https://github.com/SirEdvin/ProgressivePeripherals/pull/17

### DIFF
--- a/src/main/java/dan200/computercraft/api/turtle/ITurtleAccess.java
+++ b/src/main/java/dan200/computercraft/api/turtle/ITurtleAccess.java
@@ -15,6 +15,8 @@ import net.minecraft.util.Direction;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.vector.Vector3d;
 import net.minecraft.world.World;
+import net.minecraftforge.common.capabilities.Capability;
+import net.minecraftforge.common.util.LazyOptional;
 import net.minecraftforge.items.IItemHandlerModifiable;
 
 import javax.annotation.Nonnull;
@@ -297,4 +299,13 @@ public interface ITurtleAccess
      * @see #updateUpgradeNBTData(TurtleSide)
      */
     void updateUpgradeNBTData( @Nonnull TurtleSide side );
+
+    /**
+     * Tiers to get capability from turtle
+     *
+     * @param capability required capability
+     * @return LazyOptional capability, according to forge logic
+     */
+    @Nonnull
+    <T> LazyOptional<T> getCapability( @Nonnull Capability<T> capability);
 }
--- a/src/main/java/dan200/computercraft/api/turtle/ITurtleUpgrade.java
+++ b/src/main/java/dan200/computercraft/api/turtle/ITurtleUpgrade.java
@@ -84,28 +84,20 @@ public interface ITurtleUpgrade extends IUpgradeBase
     }
 
     /**
-     * Called when something inside turtle (mostly, this or another turtle upgrade) tries to get turtle capabilities
+     * Called when something inside turtle (mostly, this or another turtle upgrade) tries to get turtle capabilities for first time.
      *
      * @param turtle     Access to the turtle that the tool resides on.
      * @param side       Which side of the turtle (left or right) the tool resides on.
      * @param capability Required capability for turtle
      * @return Whether the turtle upgrade can provide this capability, instance of this capability should be returned.
      * Take a note, that storing information about capability internal data should be performed on upgrade side,
-     * turtle logic will not handle it.
+     * turtle logic will not handle it. But turtle itself will handle invalidation logic and call this method again in case of invalidation.
      */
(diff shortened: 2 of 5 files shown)


SirEdvin:
oh, current version contians ideological gaps, don`t review, please)

SirEdvin:
Okey. now everything should be just fine

SquidDev:
Apologies for taking so long to get to this. I have a lot of thoughts, and it's taken a bit of time to get them in order. I think there's a couple of things we can solve in one go here:

 - Capabilities on turtles. Aside, do we want capabilities on turtle item stacks too? It might be cool if turtles could, for instance, act as tanks in item form.
 - Persisting data when turtles are broken/placed (#847).
 - Handling additional NBT data on the upgrade `ItemStack` (we also need to handle doing this in a crafting grid).

I think the first two of these are relatively easy to solve in a cohesive manner by adding the following:

 - `interface TurtleUpgradeData extends ICapabilityProvider, INBTSerializable<Tag> { }`
 - `public TurtleUpgradeData ITurtleUpgrade.createData(ITurtleAccess, TurtleSide)`.
 - `public TurtleUpgradeData ITurtleAccess.getUpgradeData(TurtleSide)`.

I think we probably also need a generic `ITurtleAccess.setChanged()` method which just calls `TileEntity.setChanged()`.

Basically upgrades can define a class which implements `TurtleUpgradeData` and stores data about their upgrade (such as a fluid tank's contents or a list of waypoints). Those can then be persisted via `INBTSerializable<Tag>` on both the turtle block entity and item stack.

`createData` will be called before `createPeripheral`, so you can then access the constructed object when constructing your peripheral, rather than accessing it every method call.

`ICapabilityProvider` also allows this interface to expose capabilities if it wants, though those are not required.


As far as saving/reading to the upgrade item, perhaps we add two additional methods to `TurtleUpgradeData` which are basically `serializeToStack(ItemStack)` (or `ItemStack serializeToStack()`) and `deserializeFromStack(ItemStack)`?

## Example

```java
class FluidData implements TurtleUpgradeData {
  private final ITurtleAccess turtle;
  private final FluidTank current = new FluidTank();

  @Override
  public deserializeFromStack(ItemStack stack) {
    current.readFromNBT(stack.getCompoundTag().get(FLUID_TANK_TAG)); // TODO: more error checking please
  }

  // TODO: Storage and capabilities
}

class FluidUpgrade implements ITurtleUpgrade {
  @Override
  public TurtleUpgradeData createUpgradeData(ITurtleAccess turtle, TurtleSide side) {
    return new FluidData(turtle);
  }

  @Override
  public IPeripheral createPeripheral(ITurtleAccess turtle, TurtleSide side) {
    return new FluidPeripheral((FluidData)turtle.getUpgradeData(side));
  }
}
```

---

I don't know if this sounds even remotely sane?

SirEdvin:
Sound interesting for me. If we talk about upgrade data usage, one additional problem is data syncing between client and server and as I understand, `setChanged` are not perform sync. For example, you can look [here](https://github.com/Seniorendi/AdvancedPeripherals/blob/1.16/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/base/TurtlePeripheralOwner.java#L97) when for sync I used `playAnimation`. Maybe in additional to `setChanged` it would be nice to have something like `pushToClient`

SquidDev:
I guess there's a few options here:

 - Add `toClientTag`/`fromClientTag` from methods to the data object, and add a `pushToClient` method as you mention. I think it makes sense to avoid sending the whole store to the client when the turtle is a BlockEntity.

 - Use `getUpgradeNBTData`/`updateUpgradeNBTData` for now (as this is synced to the client). Then when 1.17 drops rename them to something client specific and change the semantics so it's not persisted by the turtle.

SirEdvin:
I am not sure, but now `getUpgradeNBTData` and `updateUpgradeNBTData`, if you speak about 1.16 are not automatically synced between client-server

SquidDev:
> If you speak about 1.16 are not automatically synced between client-server

They should be - that's how modems upgrades show which state they're in.

https://github.com/SquidDev-CC/CC-Tweaked/blob/62172c60492010c392ec1b359ac67ce8c8a0cda2/src/main/java/dan200/computercraft/shared/turtle/core/TurtleBrain.java#L178-L198

SirEdvin:
Well, from my view this methods doesn't call immediately sync, so I need to perform some sort of dirty trick like `playAnimation(None)`

SirEdvin:
Okey, I am returning to deal! 

If you call `remove` on CompoundTag, update will not be triggered and need to be performed manually, as for 1.16 1.98.1

SquidDev:
Updates are never triggered automatically - we can't detect mutation in the compound tag without doing an (obviously expensive) deep copy.

SquidDev:
Closing in favour of #903.

PR #881, opened 2021-08-02
Update and rename

DrHesperus:
## A quick checklist
 - 
Too bad there is no translation key in the graphical interface of the computer


SquidDev:
Thank you very much! I cannot believe we've had broken Russian translations for 13 months now. I thought Minecraft picked up `ru` and `ru_ru`, but evidently not.

I've fixed my weblate config, so this should hopefully not happen again.

Afraid I've merged this manually in 06f35e4997451ed0010235478dec4f3e1bb4b9cc - I made a couple of formatting tweaks to other language files which caused merge conflicts with this one.

PR #895, opened 2021-08-18
Create ja_jp.json

MORIMORI0317:
## A quick checklist


MORIMORI0317:
I'm sorry, I didn't notice Weblate

PR #900, opened 2021-08-20
Fix for computer IDs greater than 65535 crashing Rednet

Ale32bit:
While unlikely to happen, computers with an ID greater than 65535 crash rednet.open because of Modem API channels limits.


src/main/resources/data/computercraft/lua/rom/apis/rednet.lua
```diff
@@ -28,6 +28,15 @@ local tReceivedMessages = {}
 local tReceivedMessageTimeouts = {}
 local tHostnames = {}
 
+--[[ Patch the bug whenever a computer with an ID grater than 65535 crashes Rednet upon opening.
+-- It modulos the ID to always have an ID lower or equal to 65532
+-- This makes it compatible with the Modem API
+-- Reason to choose 65532 to modulo is because 65533, 65534 an nd 65535 are used respectivel
+]]
+local function idAsChannel(id)
+    return (id or os.getComputerID()) % 65532
```
fatboychummy, line 37: If you wish `65532` to be used as well, you should modulo by `65533` instead, currently the range is `0-65531`

Ale32bit: So, it was actually correct!
Used 65533 before and somehow started to have doubts about math.

src/main/resources/data/computercraft/lua/rom/apis/rednet.lua
```diff
@@ -381,13 +395,14 @@ function run()
         if sEvent == "modem_message" then
             -- Got a modem message, process it and add it to the rednet event queue
             local sModem, nChannel, nReplyChannel, tMessage = p1, p2, p3, p4
-            if isOpen(sModem) and (nChannel == os.getComputerID() or nChannel == CHANNEL_BROADCAST) then
+            if isOpen(sModem) and (nChannel == idAsChannel() or nChannel == CHANNEL_BROADCAST) then
                 if type(tMessage) == "table" and type(tMessage.nMessageID) == "number"
                     and tMessage.nMessageID == tMessage.nMessageID and not tReceivedMessages[tMessage.nMessageID]
+                    and tMessage.nRecipient and tMessage.nRecipient == os.getComputerID() or nChannel == CHANNEL_BROADCAST
```
SquidDev, line 401: Nitpick I know, but maybe

```suggestion
                    and ((tMessage.nRecipient and tMessage.nRecipient == os.getComputerID()) or nChannel == CHANNEL_BROADCAST)
```

Could also probably drop the the `tMessage.nRecipient and `, but maybe not worth it. Sorry, I just always forget `and`/`or` precedence so like to keep these things explicit.

SquidDev (changes requested):
Thanks for looking into this! One minor nitpick but otherwise looks good.

I think we'll need to update the `repeat` program here too<sup>1</sup> - I think that just needs a `tMessage.nRecipient % rednet.CHANNEL_REPEAT` when re-transmitting the message?

I think it's probably worth having some tests here, but not really sure of the best way to do it yet. If you've any ideas, do implement, but otherwise can have a fiddle.


[^1]: A program nobody has probably run in years. Thanks ender modems for making it obsolete D:.

SquidDev:
Oh lol, that's ironic.

Ale32bit:
I suggest using a channel range of 0-65500 just in case more protocols are added in CraftOS in the future.
Maybe also exposing a constant for the modulo operation like CHANNEL_BROADCAST and CHANNEL_REPEAT already do so other programs like `repeat` use it instead of needing a change.

Wojbie:
Well lovely. This PR got complicated enough to confuse parenthesis check.

src/main/resources/data/computercraft/lua/rom/apis/rednet.lua
Ale32bit: This change breaks linter.

SquidDev (approved):
Sorry, wanted to write some tests for this, but took me ages to get round to.

SquidDev:
Thank you!

PR #918, opened 2021-09-03
Allow player world interaction while pocket pc opened with no terminal

MAGGen-hub:
Dear developers of CC:Tweaked! 

I hope I will see this feature in future versions of your modification. This thing will make players able to use their poket computers as keyboards, and monitors as mouses. With this it's now posible to create something like game with better resolution (monitor.setTextScale) or other cool things. I am sorry for using java.lang.reflect, but ther is no other way to make this function works. I realy want you to use my code in your project, or if it to cursed, because of reflection, make this work using better way. Thank you for your attantion, and sorry for my weird looking english.

P.S. I decide to fully remade my request case I mess up with git and branches too much

This is a simple remake of NoTermComputerScreen.java that makes player able to move his head and interact with world objects by right clicking while poket computer in his second hand opened with no terminal.

Stuff you can interact with: 
Any Items that can be right clicked; Items/Blocks with GUI(workbench) (will close the poket computer); sScrooll menu; Move camera; Place blocks; Eat.

Stuff you can't do: 
Any actions that start with left click; Move player (case keyboard is connected with computer lol :P ).

The case you can't left click: the key.mouse.left is really deep inside game code and have strict rule - it can't be used while any GUI open, and it is to difficult to override it (even with reflection).

Demo video:

[Demo video](https://user-images.githubusercontent.com/56871670/131592321-07b0b639-dc56-46c0-809f-ff58fbe86444.mp4)

### DIFF
--- a/src/main/java/dan200/computercraft/client/gui/NoTermComputerScreen.java
+++ b/src/main/java/dan200/computercraft/client/gui/NoTermComputerScreen.java
@@ -18,6 +18,7 @@ import net.minecraft.util.IReorderingProcessor;
 import net.minecraft.util.text.ITextComponent;
 import net.minecraft.util.text.TranslationTextComponent;
 import org.lwjgl.glfw.GLFW;
+
 import javax.annotation.Nonnull;
 import java.util.List;
 
--- a/src/main/java/dan200/computercraft/client/gui/NoTermComputerScreen.java
+++ b/src/main/java/dan200/computercraft/client/gui/NoTermComputerScreen.java
@@ -43,7 +43,9 @@ public class NoTermComputerScreen<T extends ContainerComputerBase> extends Scree
     protected void init()
     {
         this.passEvents = true; // to allow gui click events pass through mouseHelper protection (see MouseHelper.OnPres:105 code string)
+        minecraft.mouseHandler.grabMouse();
         minecraft.screen = this;
+        
         super.init();
         minecraft.keyboardHandler.setSendRepeatsToGui( true );
 
--- a/src/main/java/dan200/computercraft/client/gui/NoTermComputerScreen.java
+++ b/src/main/java/dan200/computercraft/client/gui/NoTermComputerScreen.java
@@ -46,7 +46,7 @@ public class NoTermComputerScreen<T extends ContainerComputerBase> extends Scree
         this.passEvents = true; // to allow gui click events pass through mouseHelper protection (see MouseHelper.OnPres:105 code string)
         minecraft.mouseHandler.grabMouse();
         minecraft.screen = this;
-        
+
         super.init();
         minecraft.keyboardHandler.setSendRepeatsToGui( true );
 
--- a/src/main/java/dan200/computercraft/client/gui/NoTermComputerScreen.java
+++ b/src/main/java/dan200/computercraft/client/gui/NoTermComputerScreen.java
@@ -10,6 +10,7 @@ import dan200.computercraft.ComputerCraft;
 import dan200.computercraft.client.gui.widgets.WidgetTerminal;
 import dan200.computercraft.shared.computer.core.ClientComputer;
 import dan200.computercraft.shared.computer.inventory.ContainerComputerBase;
+import net.minecraft.client.Minecraft;
 import net.minecraft.client.gui.FontRenderer;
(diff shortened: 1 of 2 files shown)


MAGGen-hub:
There is a bug fix, case in non develop mode it throws "no such field exception": (the true name of "screen" is "field_71462_r"). Sorry for reuploading.

MAGGen-hub:
https://user-images.githubusercontent.com/56871670/132141600-2c5ccc11-b632-42db-b68a-9e9bdd88e434.mp4

First test of monitor with remote pocket keyboard
remote_keyboard:
```
local expect = require "cc.expect"
local tArgs = {...}
if tArgs[1] and tArgs[1] == "unload" and os.pullEventOld then
    os.pulEvent = os.pullEvetOld
    os.pullEventOld = nil
end

local modem = peripheral.find("modem", function(name, modem) return modem.isWireless() end)
local chanel = 18112

if tArds[1] then
expect.expect(1, tArgs[1], "number")
    chanel = tArgs[1]
end

os.pullEventOld = os.pullEvent

local function MyPullEvent(sFilter)
    local eventData = table.pack(os.pullEventRaw(sFilter))
    if eventData[1] == "terminate" then
        error("Terminated", 0)
    end
    if eventData[1] == "modem_message" and eventData[3] == chanel then
	eventData = eventData[5]
    end
    return table.unpack(eventData, 1, eventData.n)
end

os.pullEvent = MyPullEvent
```

key_sender:
```
local modem = peripheral.find("modem")
local chanel = ...
if not chanel then
chanel = 18112
end
while true do
    local ev = {os.pullEventRaw()}
    modem.transmit(chanel,0,ev)
    if ev[1] == "terminate" then
    return  
    end
end

```

src/main/java/dan200/computercraft/client/gui/NoTermComputerScreen.java
```diff
@@ -43,6 +48,17 @@ public T getMenu()
     @Override
     protected void init()
     {
+        this.passEvents = true; // to allow gui click events pass through mouseHelper protection (see MouseHelper.OnPres:105 code string)
+        if (mineScreen != null) // if reflection failed - pocket pc still would work but with no mouse move mechanic
+        {
+            minecraft.mouseHandler.grabMouse();
+            try {
+                mineScreen.set(minecraft, this);
```
SquidDev, line 56: Instead of using reflection, it's probably better to use an access transformer to make the field public. You can add `public net.minecraft.client.Minecraft field_71462_r # currentScreen` to `src/main/resources/META-INF/accesstransformer.cfg`, rebuild and then just set the field normally.

SquidDev (changes requested):
Sorry for taking a while to get to this, thank you very much for putting it together.

I've just left one comment below which should mean we can remove all the reflection, and make this patch much simpler.

MAGGen-hub:
AAhhh!!!! WHAT the HACK!!!! I allways get error at this!!!  `minecraft.field_71462_r = this;`
For some reason I can't use access transformer... I rebuild it... More than 10 times... And there no result.

NoTermComputerScreen.java:63: error: cannot find symbol
        minecraft.field_71462_r = this;
symbol:   variable field_71462_r

How to load it? What comand I need to use? I just tired trying to fix it... I use idea IDE...

MAGGen-hub:
Finaly! Fixed...
I need to use gradlew runData comand (build change nothing for some reason)... Only after that I get access...

MAGGen-hub:
> Sorry for taking a while to get to this, thank you very much for putting it together.
> 
> I've just left one comment below which should mean we can remove all the reflection, and make this patch much simpler.

Done) 
I remove all checkStyle "bugs", but still can't build normaly, because of: "AL lib: (EE) alc_cleanup: 1 device not closed" some errors in tests...

src/main/java/dan200/computercraft/client/gui/NoTermComputerScreen.java
MAGGen-hub: Done.

SquidDev (approved):
Sorry, took me a while to get round to sitting down and testing this out. It's a bit janky in places (as you've mentioned, can right click but not left click), but that's pretty unavoidable when doing odd things with Minecraft's engine. Still a definite improvement! :heart:

PR #920, opened 2021-09-04
Serialize better repetition error and infinity+nan handling.

Wojbie:
This PR changes 2 things in serialize logic:

1. Changes way tracking table works to allow differentiating errors between repeated and recursive table to allow end user to understand why it did not serialize and be able to optionally use allow_repetitions option if they want to.

2. Allows for nan and infinity values to be correctly serialized and not disappear in the process like they do with current logic (due to use of `tostring(num)` on them leading to nil value).

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
+++ b/src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
@@ -273,7 +273,7 @@ local function serialize_impl(t, tracking, indent, opts)
     local sType = type(t)
     if sType == "table" then
         if tracking[t] ~= nil then
-            if tracking[t] == t then
+            if tracking[t] == false then
                 error("Cannot serialize table with repeated entries", 0)
             else
                 error("Cannot serialize table with recursive entries", 0)
@@ -315,7 +315,7 @@ local function serialize_impl(t, tracking, indent, opts)
         if opts.allow_repetitions then
             tracking[t] = nil
         else
-            tracking[t] = t
+            tracking[t] = false
         end
         return result
 
--- a/src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
+++ b/src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
@@ -268,11 +268,16 @@ local g_tLuaKeywords = {
     ["while"] = true,
 }
 
+local serialize_infinity = math.huge
 local function serialize_impl(t, tracking, indent, opts)
     local sType = type(t)
     if sType == "table" then
         if tracking[t] ~= nil then
-            error("Cannot serialize table with recursive entries", 0)
+            if tracking[t] == t then
+                error("Cannot serialize table with repeated entries", 0)
+            else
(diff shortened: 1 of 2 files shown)


src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -307,13 +312,26 @@ local function serialize_impl(t, tracking, indent, opts)
...
 
     elseif sType == "string" then
         return string.format("%q", t)
 
-    elseif sType == "number" or sType == "boolean" or sType == "nil" then
+    elseif sType == "number" then
+        if t ~= t then --nan
+            return "0/0"
+        elseif t == serialize_infinity then 
+            return "1/0"
```
Lupus590, line 329: Does this work with negative infinity?

Wojbie: Oh.. right. That exist. Opse.

Wojbie:
That build fail is not on me.. this time.

SquidDev:
I wonder if a better strategy here is to add `nan = 0/0` and `inf = 1/0` in to the unserialise environment instead. It's a little more janky, but does mean we preserve the readable behaviour of serialise.

I don't know! Would like to know your opinion.

Wojbie:
Hmm that would be option too but personally i don't think readability is worth it like that? Messing with unserialise environment feels weird.

Lupus590:
Don't we have a human readable or nice layout arg? Maybe that should be done if the arg is set.

Wojbie:
Adding a option for that would be over complicating it. Honestly going with Squid or my solution makes more sense then adding a option that would mean both have to be considered and supported later on via tests and unserialise environment.

Lupus590:
I thought that we already had an arg for it, I wouldn't bother to add one for this but if it's already there it might be ok to make use of it,

src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -268,11 +268,16 @@ local g_tLuaKeywords = {
     ["while"] = true,
 }
 
+local serialize_infinity = math.huge
 local function serialize_impl(t, tracking, indent, opts)
     local sType = type(t)
     if sType == "table" then
         if tracking[t] ~= nil then
-            error("Cannot serialize table with recursive entries", 0)
+            if tracking[t] == t then
```
SquidDev, line 276: Maybe `true`/`false` here instead, where `true` marks us as currently visiting. Just feels a cleaner tri-state, and avoids us having to worry about `__eq`.

SquidDev (changes requested):

src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
Wojbie: That would indeed be smarter way to do it. 😊

SquidDev (approved):

SquidDev:
Thank you!

PR #921, opened 2021-09-11
Window scroll bugfix

fatboychummy:
Windows cause crash when passed a non-integer value, but default terminal can handle this case perfectly fine, so windows should be able to handle this as well.

This issue causes advanced computers and command computers to crash CraftOS.


SquidDev:
If we're doing this, possibly also worth handling the other functions which accept numbers? Maybe we need a helper function to guarantee the number is a real, and reject `nan`/`inf` too.

PR #925, opened 2021-09-18
Make Rednet deduplication more efficient

MCJack123:
This PR adjusts the behavior of the Rednet message deduplication system to be more efficient and to reduce the surface for denial-of-service attacks. Instead of queueing one timer for every message sent, a single timer loops around until all messages are cleared from the dedupe list, at which point the timer stops until a new message is sent/received. This reduces the pressure of adding a bunch of timers on the Java side. It also does not remove messages from the list until they are at least 9.5 seconds old, in case the timer is triggered right after message is added to the list. Lastly, the period for removal was reduced from 30 seconds to 10 seconds, as this should still be plenty of time to wait for any delays in message arrival. From quick testing in CraftOS-PC (which implements timers differently from CC:T), this reduces the memory usage of `while true do rednet.broadcast(1) end` from 500 MB to 15 MB.


src/main/resources/data/computercraft/lua/rom/apis/rednet.lua
```diff
@@ -414,14 +414,15 @@ function run()
...
-        elseif sEvent == "timer" then
+        elseif sEvent == "timer" and p1 == nClearTimer then
             -- Got a timer event, use it to clear the event queue
-            local nTimer = p1
-            local nMessage = tReceivedMessageTimeouts[nTimer]
-            if nMessage then
-                tReceivedMessageTimeouts[nTimer] = nil
-                tReceivedMessages[nMessage] = nil
+            nClearTimer = nil
+            local nNow, bFound = os.epoch("utc"), nil
```
SquidDev, line 420: Maybe "hasMore" rather than "found"? Not clear what it's found here. Similarly, would be inclined to call `k`/`v` something more intuitive (a break from tradition, I know!) as in this case their meaning is a little subtle.

src/main/resources/data/computercraft/lua/rom/apis/rednet.lua
```diff
@@ -394,8 +394,8 @@ function run()
                 if type(tMessage) == "table" and type(tMessage.nMessageID) == "number"
                     and tMessage.nMessageID == tMessage.nMessageID and not tReceivedMessages[tMessage.nMessageID]
                 then
-                    tReceivedMessages[tMessage.nMessageID] = true
-                    tReceivedMessageTimeouts[os.startTimer(30)] = tMessage.nMessageID
+                    tReceivedMessages[tMessage.nMessageID] = os.epoch("utc")
```
SquidDev, line 397: I think it's more sensible to use `os.clock()` here instead. It's tied to TPS (like `os.startTimer`) and guaranteed to be monotonic, which `os.epoch` isn't. Otherwise we're going to end up with some odd behaviour in the case of daylight savings time!

I think what it might be more intuitive to do here is set the deadline here instead, so:

```suggestion
                    tReceivedMessages[tMessage.nMessageID] = os.clock() + 9.5
```

src/main/resources/data/computercraft/lua/rom/apis/rednet.lua
```diff
@@ -26,8 +26,8 @@ CHANNEL_BROADCAST = 65535
 CHANNEL_REPEAT = 65533
 
 local tReceivedMessages = {}
-local tReceivedMessageTimeouts = {}
 local tHostnames = {}
+local nClearTimer
```
SquidDev, line 30: *cries about Hungrarian notation*. Probably not worth doing the cleanup right now. Soon :tm:.

src/main/resources/data/computercraft/lua/rom/apis/rednet.lua
```diff
@@ -414,14 +414,15 @@ function run()
...
             -- Got a timer event, use it to clear the event queue
-            local nTimer = p1
-            local nMessage = tReceivedMessageTimeouts[nTimer]
-            if nMessage then
-                tReceivedMessageTimeouts[nTimer] = nil
-                tReceivedMessages[nMessage] = nil
+            nClearTimer = nil
+            local nNow, bFound = os.epoch("utc"), nil
+            for k, v in pairs(tReceivedMessages) do
+                if nNow - v >= 9500 then tReceivedMessages[k] = nil
```
SquidDev, line 422: Following on from my `os.clock` message above, then you'd just have:

```suggestion
                if deadline >= now then tReceivedMessages[k] = nil
```

SquidDev (changes requested):
Oooooh, thank you! Sorry, left some nitpicks below but otherwise looks good.

Iterating through the whole hashmap does feel a little sad, but guess we don't have much of an alternative. Guess most of the time we're clearing the whole thing anyway.

SquidDev:
> this reduces the memory usage of [...] from 500 MB to 15 MB.

As an aside, the fact that it's this high is very sad (I imagine it's pretty similar in CC). Even if we could limit the memory of the CC computers, restricting non-Lua resources like timer handles is such a hard job.

src/main/resources/data/computercraft/lua/rom/apis/rednet.lua
MCJack123: Yeah, as much as I'd like to use more normal variable naming, I'm leaving it as-is to keep it consistent. At least consistently bad is better than inconsistent?

SquidDev (approved):

SquidDev (approved):
We should probably apply this optimisation do `repeat.lua` at some point, but it's a bit less of a priority.

Issue #535, opened 2020-09-03
State persistence and making complex programs possible


SirHall:
I felt as though this may have been suggested plenty of times before, but I was unable to find any previous issues mentioning this problem so I'm making it now.

As is very well known _[OpenComputers](https://www.curseforge.com/minecraft/mc-mods/opencomputers)_ and _[RetroComputers](https://www.curseforge.com/minecraft/mc-mods/retrocomputers)_ advertise this feature upfront, however the fact that neither of them have been updated past 1.12.2 mean that they cannot be used in any more recent mod packs.

While I love both ComputerCraft and OpenComputers for many reasons, it is often hard to use ComputerCraft for anything more complicated than directly transforming simple inputs into simple outputs and simple monitoring of the world around it(which is persistent) and immediately acting on it.

Lua to my knowledge supports saving it's global environment to a file and loading it back up again, whilst I expect the actual implementation to be quite complex, I do find it hard to use ComputerCraft to its fullest without the reliability of whatever it's doing staying around in its memory by the time it has loaded again.

The primary workaround includes writing each variable change to a file which is then loaded up when the program is run again usually as a startup script. This is purely a hack that fails to fix the issue that the position in a script that is currently being run is also lost. It simply isn't reasonable to attempt to write a program in such a way where it can be stopped at any time, and expect to continue running from any arbitrary point in the code purely by reloading saved variables, especially should recursion be used at any point.

ComputerCraft in my opinion is simply wonderful, but since 2013-2014 this has continued to be the number 1 issue that makes this mod difficult to use. Thankyou for the great work you have done in continuing to support this wonderful mod, and I would argue that this is the last real kink in this mod that needs fixing. I realize given the age of this mod it may be rather difficult, but I feel it would truly take ComputerCraft to a stage where it is nearly complete.

Thanks again for your amazing work!

Lupus590:
[...]

PR #926, opened 2021-09-20
Add new `getCommandID` API to turtles make it possible to perform crash/state recovery

gnif:
The following pattern is commonly implemented to track the current position of the turtle when a script needs to be able to keep track of where the turtle is in the world.

```lua
loadState()

turtle.forward()
state.x = state.x + 1
saveState()
```

The problem is if the script is terminated before the state is saved, but after the turtle has moved, the state information does not get updated and the script no longer knows where the turtle is located. This change resolves this by making it possible to implement an intent log/journal by providing a persistent running ID to the lua script.

For example

```lua
function processJournal()
  local lastID = turtle.getCommandID()
  
  while next(state.journal) do
    local entry = table.remove(state.journal, 1)
    
    -- only process journal entries with newer IDs
    if entry[0] > lastID then
      if entry[1] == 'f' then turtle.forward()
      elseif entry[1] == 'b' then turtle.back()          
      elseif entry[1] == 'u' then turtle.up()
      elseif entry[1] == 'd' then turtle.down()
      elseif entry[1] == 'l' then turtle.turnLeft()
      elseif entry[1] == 'r' then turtle.turnRight() 
      end
    end
  end

  saveState()
end

function forward()
  -- write the intent to move forward to the journal and update the position
  local id = turtle.getCommandID()  + 1
  table.insert(state.journal, {id, "f"})
  state.x = state.x + 1
  saveState()
    
  -- perform the action
  turtle.forward()
  
  -- remove the processed entry from the journal
  table.remove(state.journal, 1)
  saveState()  
end

loadState()
processJournal()
while state.x < 10 do
  forward()
end
```

This could also be further optimized to avoid excessive disk reads/writes, for example:

```lua
function forward()
  -- write the intent to move forward to the journal and update the position
  local id = turtle.getCommandID() + 1
  table.insert(state.journal, {id, "f"})
  state.x = state.x + 1
end

loadState()
processJournal()

while state.x < 10 do
  forward()
end
saveState()
processJournal()
```

Idea initially mentioned in #535

### DIFF
--- a/src/main/java/dan200/computercraft/api/turtle/ITurtleAccess.java
+++ b/src/main/java/dan200/computercraft/api/turtle/ITurtleAccess.java
@@ -45,6 +45,13 @@ public interface ITurtleAccess
     BlockPos getPosition();
 
     /**
+     * Returns the ID of the last executed command.
+     *
+     * @return the ID of the last executed command.
+     */
+    int getCommandID();
+
+    /**
      * Attempt to move this turtle to a new position.
      *
      * This will preserve the turtle's internal state, such as it's inventory, computer and upgrades. It should
--- a/src/main/java/dan200/computercraft/shared/turtle/apis/TurtleAPI.java
+++ b/src/main/java/dan200/computercraft/shared/turtle/apis/TurtleAPI.java
@@ -55,6 +55,18 @@ public class TurtleAPI implements ILuaAPI
     }
 
     /**
+     * Get the ID of the last successful command.
+     *
+     * @return the ID of the last successful command
+     * @cc.return The ID of the last successful command
+     */
+    @LuaFunction
+    public final int getCommandID()
+    {
+        return turtle.getCommandID();
+    }
+
+    /**
      * Move the turtle forward one block.
      *
      * @return The turtle command result.
--- a/src/main/java/dan200/computercraft/shared/turtle/core/TurtleBrain.java
+++ b/src/main/java/dan200/computercraft/shared/turtle/core/TurtleBrain.java
@@ -59,6 +59,7 @@ public class TurtleBrain implements ITurtleAccess
     public static final String NBT_LEFT_UPGRADE = "LeftUpgrade";
     public static final String NBT_LEFT_UPGRADE_DATA = "LeftUpgradeNbt";
     public static final String NBT_FUEL = "Fuel";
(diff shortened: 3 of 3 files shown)


SirEdvin:
I am not sure, that this is good approach to solve this problem. I guess, there is no actual difference between _issued command_ and _successfully issued command_, so this code will be useful in some cases, and useless in another. 

Why you don't like approach with detecting coordinates every start?

gnif:
Because this would not work if the script is terminated

```lua
function turtle.logged_forwar1d(journal)
    journal:add("turtle", "forward", "start")
    local res, err = turtle.forward()
    -- If the script is terminated HERE, which can happen when the chunk is unloaded, the turtle has moved but our script never gets the error.
    if err ~=nil then
        journal:add('turtle", "forward", "failure")
    else
        journal:add('turtle", "forward", "success")
    end
end
```
> Why you don't like approach with detecting coordinates every start?

As in using a GPS system? Because there is no reason why relative tracking of coordinates should not be possible.

gnif:
Updated to only count successful commands

SquidDev:
I'm afraid I'm not going to merge this.

While I do think the journaling approach is the most sensible approach available to use now, I don't really want to add interim systems for persistence - I think CC just has to do this properly (i.e. #535).

gnif:
Sure, I would prefer full persistence also, however from what I understand this is a far more complex thing to implement into CC and as such the timeframe to have it done I would also expect to be quite long.

This patch would at the very minimum give us a way forward while we wait for such a persistence system to be implemented. While I understand the desire to keep the codebase clean from temporary workarounds/hacks (I am also the project lead of several large FOSS projects), sometimes it makes sense to allow something like this to give the users an option in the interim.

The current recommendation to people is to move to OpenComputers due to this issue, if this simple patch consisting of a handful of lines with zero complex logic were to be provided it would give us another option now.

Either way, respect and appreciation for your management of this project, I have had many hours of enjoyment from it, new APIs like `identify` were something missing from the original CC that drove me away from it and I was glad to see it finally seeing such enhancements.

SquidDev:
> While I understand the desire to keep the codebase clean from temporary workarounds/hacks 

The main problem CC has in that it's a programming environment with reasonable back-compat commitments. There isn't such thing as a _temporary_ hack - if it's in the computer's environment, it really has to stay there.

PR #954, opened 2021-11-14
Optimize peripheral calls in rednet.run

xAnavrins:
Currently, rednet.run will do a peripheral call (.isOpen) every time it receives a modem message on an open port, regarless if rednet.open was called beforehand.

This changes it so that isOpen is run last, after checking that the message is destined to the host, and is a valid rednet formatted message.

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/apis/rednet.lua
+++ b/src/main/resources/data/computercraft/lua/rom/apis/rednet.lua
@@ -403,10 +403,11 @@ function run()
         if sEvent == "modem_message" then
             -- Got a modem message, process it and add it to the rednet event queue
             local sModem, nChannel, nReplyChannel, tMessage = p1, p2, p3, p4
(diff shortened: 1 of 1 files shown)


SquidDev:
Thanks!

LoganDark:
For those looking at broken TLCOs (top-level coroutine overrides) after this change, rednet can still be reliably crashed by queueing this event:

```lua
os.queueEvent('modem_message', {}, rednet.CHANNEL_BROADCAST, nil, {nMessageID = -1})
```

xAnavrins:
So you're telling me that there's TLCO methods out there that overrides `isOpen`?
Instead of using the time proven method you just sent?

LoganDark:
> So you're telling me that there's TLCO methods out there that overrides `isOpen`? Instead of using the time proven method you just sent?

Traditionally, rednet was crashed using `os.queueEvent('modem_message', 0)`, but this PR fixed that. I did the reversing to figure out the new payload required to cause a crash.

SkyTheCodeMaster:
I just use `debug.setupvalue` to change the `bRunning` variable to false :)

Issue #902, opened 2021-08-23
Allow different peripheral act like modem


SirEdvin:
Currently, to have ability to propagate peripherals, peripheral should have type `modem`.
This is little confusing in some cases, when peripheral are not actually a modem, but need to have this type, but not provide full API compatibility (for example, peripheral container, etc). 

So, I guess, this api should be changed to allow different peripherals act like modems and propagate peripherals, but not to be modem itself.

SquidDev:
Now that #963 is in, we're in a position to do this pretty easily - can have another peripheral type for things which provide modems.

That said, don't know if people have thoughts on a suitable type name? Some ideas:
 - "peripheral_provider" is obviously accurate but slightly janky.
 - "hub" also feels appropriate (both in the networking and "usb hub" sense), but maybe a little _too_ non-descriptive?

SirEdvin:
Voting for hub, i guess

PR #963, opened 2021-11-24
Allow peripherals to have multiple types

SquidDev:
A screenshot of a ComputerCraft turtle showing the output of the "peripherals" program. It lists two peripherals - a furnace and a chest - each with their own distinct type as well as an additional "inventory" type

Peripherals can now have multiple types:
 - A single primary type. This is the same as the current idea of a type - some identifier which (mostly) uniquely identifies this kind of peripheral. For instance, "speaker" or "minecraft:chest".
 - 0 or more "additional" types. These are more like traits, and describe what other behaviour the peripheral has - is it an inventory? Does it supply additional peripherals like a wired modem)?.

This is mostly intended for the generic peripheral system, but it might
prove useful elsewhere too - we'll have to see!

 - `peripheral.getType` (and `modem.getTypeRemote`) now returns 1 or more values, rather than exactly one.
 - Add a new `peripheral.hasType` (and `modem.hasTypeRemote`) function which determines if a peripheral has the given type (primary or additional).
 - Change peripheral.find and all internal peripheral methods to use `peripheral.hasType` instead.
 - Update the peripherals program to show all types.

This effectively allows you to do things like `peripheral.find("inventory")` to find all inventories.

---

### Still left to do
I put this together in an hour while waiting for food to cook, it's very WIP.

 - [x] Documentation. I think this is a good opportunity to flesh out the peripheral API docs in general - I suspect there's more detail we could go into in the introductory paragraphs.
 - [x] Tests.
 - [x] Add additional types for fluids and energy methods too.
 - [ ] Add an additional type for "peripheral providers" (see #902) and swap over to this.

---

API is undecided on, and still not sure if this is a good/useful feature. Would appreciate any feedback!

Some outstanding questions:
 - Is `peripheral.hasType` or `peripheral.isType` better?
 - Do we need to have a separate `getTypes` function instead of changing the signature of `getType`?

### DIFF
--- a/doc/events/redstone.md
+++ b/doc/events/redstone.md
@@ -2,7 +2,7 @@
 module: [kind=event] redstone
 ---
 
-The @{redstone} event is fired whenever any redstone inputs on the computer change.
+The @{event!redstone} event is fired whenever any redstone inputs on the computer change.
 
 ## Example
 Prints a message when a redstone input changes:
--- a/src/main/java/dan200/computercraft/api/peripheral/IPeripheral.java
+++ b/src/main/java/dan200/computercraft/api/peripheral/IPeripheral.java
@@ -10,6 +10,8 @@ import net.minecraftforge.common.capabilities.Capability;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+import java.util.Collections;
+import java.util.Set;
 
 /**
  * The interface that defines a peripheral.
@@ -30,6 +32,18 @@ public interface IPeripheral
      */
     @Nonnull
     String getType();
+
+    /**
+     * Return additional types/traits associated with this object.
+     *
+     * @return A collection of additional object traits.
+     * @see PeripheralType#getAdditionalTypes()
+     */
+    @Nonnull
+    default Set<String> getAdditionalTypes()
+    {
+        return Collections.emptySet();
+    }
 
     /**
      * Is called when when a computer is attaching to the peripheral.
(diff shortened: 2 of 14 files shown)


Lupus590:
So with this new system, a turtle would turn up as a `computer` type and an `inventory` type? That will make network-based inventories with turtles so much easier.

SquidDev:
> So with this new system, a turtle would turn up as a computer type and an inventory type? 

Well, could. This PR doesn't actually make that specific change.

hugeblank:
If I'm reading this correctly the changes to `getType` and `hasType` would break CC programs that use those methods. Maybe a couple new methods:`getTrait(s)` and `hasTrait(s)` to provide and test the traits of a peripheral could be better?

SquidDev:
> If I'm reading this correctly the changes to `getType` and `hasType` would break CC programs that use those methods.

Well, `hasType` is new, so shouldn't break anything.

`getTypes` might break when used in a varargs context (e.g. `("hello"):gsub("hello", peripheral.getType())` will error if `getType` returns multiple values) - I'm not sure if this is a concern in practice. If we think it is, possibly could use `getTypes`? - I think I want to avoid using "trait" in user-facing documentation, as it's a whole 'nother term to introduce.

SquidDev:
Checking my local notes I had on this, and I originally suggested `peripheral.isType` instead of `peripheral.hasType`. What do people think is a better name?

Lupus590:
`has` implies that a peripheral can have multiple, whereas `is` doesn't, I'd go with `has`

SquidDev:
I need to write some tests for this, but otherwise I think this is ready. Will leave it to sit for a few days in case anyone else has any comments, and then will merge and push an update on the 30th (assuming all goes to plan).

PR #977, opened 2021-12-01
Add package.searchpath

MCJack123:
This PR adds `package.searchpath` from Lua 5.2, which searches a path string for the specified module name. This was implemented by moving the search functionality out of `from_file` into a new function. Adding this function will enhance compatibility with Lua 5.2 in the shell.


SquidDev:
Thank you!

Hadn't really registered this in the docs before. Guess it's probably worth adding `package.searchers` at some point, and making the various loaders + `require` return their extra data, but that can wait for another day.

PR #990, opened 2021-12-16
Add support for Lua 5.2 via Cobalt#50

MCJack123:
This PR implements the necessary CC-side changes to implement Lua 5.2 in CC, following SquidDev/Cobalt#50. It adds the `disableGotoStatement` config option, which makes `goto` always act like a name for compatibility, disabling its functionality.

This does not build at the moment since SquidDev/Cobalt#50 hasn't landed and isn't on the Maven. However, you can still test it by replacing the `shade` line for Cobalt with `shade files('../path/to/Cobalt.jar')`.


Lupus590:
Well now I have to adjust cc:n to use this pr.

PR #982, opened 2021-12-04
Add arbitrary audio support to speakers

SquidDev:
This updates speakers to allow them to play arbitrary audio as discussed in #822.

We add a single new method to the speaker, `playAudio(audio_data[, volume])`. This accepts a list of amplitudes between -128 and 127 and sampled at 48kHz, and then stored in an internal buffer. The internal buffer is then sent to the client when needed and a `speaker_audio_empty` event is fired, marking that the internal buffer can be populated again.

In practice, this ends up looking something like this:

```lua
local dfpwm = require("cc.audio.dfpwm")
local speaker = peripheral.find("speaker")

local decoder = dfpwm.make_decoder()
for chunk in io.lines("example.dfpwm", 16 * 1024) do
    local buffer = decoder(chunk)

    while not speaker.playAudio(buffer) do
        os.pullEvent("speaker_audio_empty")
    end
end
```

While the API is a little odd, I think it's the one which gives us the most flexibility - it allows for streaming audio from a file, over a websocket, etc... and also allows for doing real-time audio processing:

<details>
<summary>A fancier script which adds an echo to the audio</summary>

```lua
local dfpwm = require("cc.audio.dfpwm")
local speaker = peripheral.find("speaker")

-- Speakers play at 48kHz, so one second is 48k samples. We first fill our buffer
-- with 0s, as there's nothing to echo at the start of the track!
local samples_i, samples_n = 1, 48000
local samples = {}
for i = 1, samples_n do samples[i] = 0 end

local decoder = dfpwm.make_decoder()
for chunk in io.lines("data/example.dfpwm", 16 * 1024) do
    local buffer = decoder(input)

    for i = 1, #buffer do
        local original_value = buffer[i]

        -- Replace this sample with its current amplitude plus the amplitude from one second ago.
        -- We scale both to ensure the resulting value is still between -128 and 127.
        buffer[i] = original_value * 0.7 + samples[samples_i] * 0.3

        -- Now store the current sample, and move the "head" of our ring buffer forward one place.
        samples[samples_i] = original_value
        samples_i = samples_i + 1
        if samples_i > samples_n then samples_i = 1 end
    end

    while not speaker.playAudio(buffer) do
        os.pullEvent("speaker_audio_empty")
    end
end
```

</details>

This PR previously had the speaker accept DFPWM1a blobs directly. This is no longer the case - it now uses PCM. _However_, the network layer still uses DFPWM1a, so we can keep that terrible audio quality you know and love.

## Open questions 
[...]

PR #993, opened 2021-12-18
Reduce latency of speakers with direct playback

MCJack123:
This PR replaces the client side of audio playback with direct OpenAL calls, allowing the use of smaller buffers to reduce latency. It also reduces the size of the buffer on the server so that data is sent quicker.

This is pretty experimental, and there's likely some incompatibility/bad practices that I'm overlooking, but from my testing it appears to work very well. There may be some issues with distance attenuation, as it uses its own constants to calculate that, but otherwise all of the positional stuff works fine. This also makes `playAudio` and `playSound` separate, which may not be desirable - this can be fixed with a few simple `if` checks.


SquidDev:
Could you share the program you're using to test this? At first blush this looks identical to what MC is doing internally (aside from the change to `CLIENT_BUFFER`).

MCJack123:
The program is available [on GitHub](https://github.com/MCJack123/tracc/blob/playAudio/tracc.lua), and the module files I'm using are available here: http://gsf.joshw.info/t/The%20Game%20of%20Life,%20Yahtzee,%20Payday%20(conv)%20(2005-11-07)(Black%20Lantern%20Studios)(Destination%20Software%20Inc.)[GBA].7z

SquidDev:
I just tried this out on my machine, and it sounds incredibly stuttery. Even the `speaker play` program will cut out and then only play occasional starts of audio.

znepb:
A couple of issues I've noticed - it seems the audio isn't categorized under any audio volume slider, nor does it respect distancing (it only stops playing when the chunks are unloaded). Plus, as SquidDev mentions, the audio becomes choppy even if the server is running at 20 TPS.

Toad-Dev:
I had stutter in my testing too. Stutter happens on a coin flip the first time a speaker peripheral is used after freshly loading the world, but always occurs when reusing a speaker peripheral. The reuse issue feels like something wrong with the AL buffer code as it just never recovers and gets increasingly worse with each new stream until the world is reloaded.

I think the change to `CLIENT_BUFFER` is a bit optimistic in general. A 20ms window, when minecraft is ticking the peripheral (ideally) every 50ms, means it's possible for a speaker to pass on sending pending audio one tick, and then be a few ms late on sending that audio the next tick.

SquidDev:
@MCJack123 The above commit hopefully alleviates some of your concerns raised here and #982. We're never going to be able to reduce the latency to a single tick, as that's just not sensible on an SMP server, but I think the current values are Good Enough :tm:.

MCJack123:
It appears the second commit was causing the stuttering. I've reverted it to fix the issue.

> I think the change to `CLIENT_BUFFER` is a bit optimistic in general. A 20ms window, when minecraft is ticking the peripheral (ideally) every 50ms, means it's possible for a speaker to pass on sending pending audio one tick, and then be a few ms late on sending that audio the next tick.

I agree; the value I set it to was mostly just for testing. I've bumped it up to 0.1s, which is lower than Squid changed it to, but it should theoretically be able to handle as low as 10 tps without issues (?).

> @MCJack123 The above commit hopefully alleviates some of your concerns raised here and #982. We're never going to be able to reduce the latency to a single tick, as that's just not sensible on an SMP server, but I think the current values are Good Enough:tm:.

I tried it out, and it is working much better. With the changes, my tracker's now only about 4 rows behind the video. One thing that would be nice is to have `speaker_audio_empty` be sent a bit more regularly at the beginning, as it's sent immediately when starting to play audio and the buffers aren't full yet, but that's probably a bit more work than is necessary.

SquidDev:
I'm going to close this for now, as things are at a state I'm happy with. There's definitely room to change how the internals work once 1.100 is released.

PR #1000, opened 2021-12-26
Clarify term_resize event.

### LINKED ISSUE #999: Multishell scrolls the window when the terminal is resized and the cursor is offscreen
Multishell scrolls the window when the terminal is resized and the cursor is offscreen. This breaks some more advanced drawing techniques (like only redrawing content of the screen that changed) because the screen scrolls unexpectedly.

There are two solutions while multishell has this behavior:
- Always put the cursor at (1, 1) after a render. I do this currently. It means I can't use the cursor inside the program and I have to use some other visual indicator, such as a reverse-video block.
- Always redraw the entire screen after a resize. Hardcoding routines for calculating how the screen changed when resized under multishell is too fragile, especially since it depends on the cursor position. Redrawing the entire screen is the only reasonable option.

The culprit is this block:

https://github.com/cc-tweaked/CC-Tweaked/blob/e558b31b2b0195a210c9a542df07aae1460178f4/src/main/resources/data/computercraft/lua/rom/programs/advanced/multishell.lua#L195-L198

This doesn't seem conducive to anything other than the shell - it's unexpected behavior for the screen contents to move when the terminal resizes. It's useful to keep the shell prompt on-screen in that case, but it's an unexpected implementation detail that can't be disabled by programs. I think the drawbacks outweigh the benefits. :/

Please consider removing it from multishell. IMHO, it would have made more sense in `read`, but CC already forgets offscreen contents by the time you receive a `term_resize` event, so it's not possible to reimplement there.

---

I'd like to note that I've seen the recent announcement post detailing SquidDev's departure from CC:T. That is quite unfortunate, especially on the day that I discovered this issue, and I'm not sure if it's possible to solve this if CC:T is no longer making releases.

I'm posting this bug report regardless just to have.

Wojbie:
Resolves #999

Clarifies what `term_resize` event indicates.

### DIFF
--- a/doc/events/term_resize.md
+++ b/doc/events/term_resize.md
@@ -2,7 +2,9 @@
 module: [kind=event] term_resize
 ---
 
-The @{term_resize} event is fired when the main terminal is resized, mainly when a new tab is opened or closed in @{multishell}.
+The @{term_resize} event is fired when the main terminal is resized, mainly when a new tab is opened or closed in @{multishell} or when the terminal out is being redirected by the "monitor" program and monitor size changes.
+
+This event indicates that there is no guarantee of what is on the screen and where it is located. GUI based software should redraw its contents but simple terminal programs can safely ignore it.
 
 ## Example
 Prints :


LoganDark:
Guess that works.

PR #1002, opened 2021-12-27
Fix hasTypeRemote not working with additional types. (#1001)

### LINKED ISSUE #1001: peripheral.find can't find remote peripherals by their additional types.
### Minecraft Version

1.16.x

### Version

1.100.0

### Steps to reproduce:

1. Place a computer, modem, and chest in the following arrangement:

2. Enter lua prompt and run `peripheral.find("inventory")`
3. No peripherals will be found.

---

Peripherals placed directly against the computer can be found like this.

Toad-Dev:
Fixes #1001. The problem was actually in the modem method `hasTypeRemote`. It looks like a simple typo snuck in and the modem was checking additional types against it's own type rather than the queried type.


SquidDev:
Hah, good catch.

PR #1009, opened 2022-01-04
Added interaction when turtle places empty bucket on cauldron.

IICarst:
Fix for this issue: https://github.com/cc-tweaked/CC-Tweaked/issues/1008


dathide:
Looks good to me, hope this gets merged soon!

dathide (changes requested):
This would allow BoatItem, WaterLilyBlockItem, and BottleItem to be used on cauldrons just like buckets.

IICarst:
I added an condition to make sure only buckets can be used.

SquidDev:
Thank you for the PR. I'm afraid this isn't something I really want to special case.

PR #1015, opened 2022-01-13
speaker_audio.md: Fix missing add/sum typo

Paspartout:
Just found this while reading the great guide/documenation on audio.


SquidDev:
Thanks!

PR #1017, opened 2022-01-19
Fix table with mouse button codes in documentation.

Possseidon:
Codes for right and middle mouse buttons were swapped.


SquidDev:
Thanks!

PR #1021, opened 2022-01-20
add concatenation utility

Ocawesome101:
This is a simple file concatenation utility.  It's a bit like `cat`, except CraftOS doesn't have pipes so its usage is a little different.  I'll add a help page in a bit.


Ocawesome101:
Help page added, also changed to use `shell.resolve` on all file paths.

PR #1022, opened 2022-01-21
Some printout fixes.

toad-dev:
This is a small thing that was contributed downstream, but I want to make it available here in case there are more releases.

Non-book printouts weren't centered as a whole because the logic to left align the first page had an off-by-one error that added an extra bit of offset to the transformation matrix. Also the constants for the text margins were slightly different than what they should be, adding another ~1px of error to both the transform and the text location relative to background. Here's a comparison in an item frame, where it's most noticeable:

centered-printouts

Note the little bit of item frame peeking out on the ~left~ right side in the before image. I know the text almost looks too far to the left in the after image, but it's because the cracks in the background extend further in on the left side. It's centered "properly" in the after image.

Also I fixed some z-fighting with the bold borders that render on the farthest left and right sides of the printouts. It doesn't really come across in a screenshot but was visible in motion.

Oh, and finally, there was a bug in `GuiPrintout` that made it not respond to the arrow keys. That was fixed too.

### DIFF
--- a/src/main/java/dan200/computercraft/client/gui/GuiPrintout.java
+++ b/src/main/java/dan200/computercraft/client/gui/GuiPrintout.java
@@ -52,8 +52,7 @@ public class GuiPrintout extends ContainerScreen<ContainerHeldItem>
     @Override
     public boolean keyPressed( int key, int scancode, int modifiers )
     {
-        if( super.keyPressed( key, scancode, modifiers ) ) return true;
-
+        // Check for key presses.
         if( key == GLFW.GLFW_KEY_RIGHT )
         {
             if( page < pages - 1 ) page++;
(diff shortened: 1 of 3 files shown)


SquidDev:
Thank you for submitting this patch, and sorry for taking so long to properly look at this.

> Non-book printouts weren't centered as a whole because the logic to left align the first page had an off-by-one error that added an extra bit of offset to the transformation matrix. 

I've cherry picked this fix in 118b89ea415aea9e10637eb221dc29c338339041. WRT the other changes:

 - I've been unable to reproduce any issues with keyboard navigation. Do you know if this was Fabric specific (I know Forge does patch some things in the underlying Screen classes), or caused by a combination of other mods? No worries if you don't know, just kinda curious!

 - I don't think I want to merge the margin changes right now. As you say, the texture is a bit weird, so I'm not sure if centring it looks any better than it did before. Might be possible to adjust the texture if someone wants to have a go at that, but I think for now I'll just leave it consistent with CC.

toad-dev:
No problem on the wait! These were very small fixes.

1. I can't remember if I tested whether the keyboard issues were present before cherry picking that commit, but if things are working now then I'd guess I just forgot, sorry. It must've been a Fabric only issue.
2. Seeing it with fresh eyes now, I agree the margins looked better before! Those constants must've been chosen to suit the texture.

I think the second commit to fix the z-fighting is still good, if minor.

SquidDev:
Manually merged with 51c3a9d8afa9b5800ee5bf1848447f451a19cbbe! Thank you very much

Issue #1008, opened 2022-01-04
Cannot use turtles with cauldrons


IICarst:
### Minecraft Version

1.18.x

### Version

1.100.1

### Details

I was trying to automate my dripstone lava farm but when I tried to empty a lava cauldron with a turtle it gave an error: "Cannot place item here". I used the turtle.place() function with an empty bucket selected on the cauldron.

dathide:
I also tried this today. I think this feature should be added.

JohnnyIrvin:
Based on https://github.com/cc-tweaked/CC-Tweaked/pull/1009#issuecomment-1172919291 --- Should we assume this issue to be closed as its not going to be implemented?

Lupus590:
> Should we assume this issue to be closed as its not going to be implemented?

Yes, SquidDev has expressed that this is not something he wants the mod to do.

PR #1032, opened 2022-01-31
Add turtle bucket-cauldron functionality to fix issue #1008.

angel1254mc:
Attempt at adding functionality for turtles using buckets on cauldrons. Based on #1008 and the solution provided by @IICarst in #1009, but with added ability to both empty and fill buckets/cauldrons.


Seniorendi:
Sweet

SquidDev:
Thank you for the PR. I'm afraid this isn't something I really want to special case.

PR #1058, opened 2022-04-02
Add note about special JSON values in docs for `textutils.unserializeJSON`

MCJack123:
This PR adds a quick blurb noting that `textutils.unserializeJSON` returns special values for `null` and `[]`.

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
+++ b/src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
@@ -658,6 +658,9 @@ do
     -- This may be used with @{textutils.serializeJSON}, or when communicating
     -- with command blocks or web APIs.
     --
+    -- If a `null` value is encountered, it is converted into @{textutils.json_null}.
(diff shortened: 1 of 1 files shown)


Wojbie:
If we are touching this part of docs/api then probably note about null should be added to `serializeJSON` and both special tables should get __newindex meta table to error if someone attempts to modify them.

SquidDev:
An alternative approach might be to expand the docs to `empty_json_array` and `json_null` to mention they serialised to/from an empty array/null, and then just tag them with `@see empty_json_array Use to serialise an empty array.` Not sure if that's any better - thoughts? 

Happy to merge either way.

SquidDev:
@MCJack123 Any thoughts on the above question?

MCJack123:
Oops, didn't realize you were asking a question - sorry! I think the deserialization behavior should definitely be noted separately in `textutils.unserializeJSON`, as users are likely not going to notice the placeholders when just looking for how to use the JSON serialization functions. I'll also add `@see` references for easy linking, including for `serializeJSON`.

SquidDev:
Yeah, that seems seasonable. Thanks!

PR #1062, opened 2022-04-06
Make readme more consistent with CC:R's

3prm3:
This is to make the readme look more like CC:R's in style, look here for CC:R's repo https://github.com/cc-tweaked/cc-restitched


SquidDev:
I'm not really a fan of putting images as a replacement for headings. I think I'm fine with the current setup.

PR #1064, opened 2022-04-07
[1.16.x] fallback to sided cabability access if no internal access is supported

mcenderdragon:
This adresses https://github.com/cc-tweaked/CC-Tweaked/issues/1061

This adds a sided cabaility acces for generic capabilities if no nullside (internal access) is supported.


SquidDev:
Applied a modified version of this manually - this commit doesn't actually build as `wrapper` is no longer final and thus can't be captured by the lambda.

PR #1065, opened 2022-04-07
Some further VBO optimisations

SquidDev:
A continuation of #1057, following on from @Toad-Dev's investigations (thank you!).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -78,6 +78,18 @@ minecraft {
             workingDirectory project.file('run')
         }
 
+        clientJitwatch {
+            parent runs.client
+            inheritJvmArgs true
+            jvmArgs [
+                "-XX:+UnlockDiagnosticVMOptions",
+                "-Xlog:class+load=info",
+                "-XX:+LogCompilation",
+                "-XX:+PrintAssembly",
+                "-XX:+DebugNonSafepoints",
+            ]
+        }
+
         server {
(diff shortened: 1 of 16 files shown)


SquidDev:
Also made some optimisations to the TBO renderer, most notably skipping rendering the depth buffer, which saves about 4ms/frame (bumping us to 100fps when rendering constantly-changing 480 monitors).

I think there's some optimisations which can be made here:
 - Generating the TBO buffer is still pretty slow (almost 50% of our render time)! I don't think this is the ByteBuffer here - possibly bounds checks on the array? Time to go back to JITWatch here I think and see what's holding us up.
 - The other slow bit is actually rendering the terminal quad:
   
image

   Obviously we could shave off a tiny bit by storing our single quad in a VBO, but I'm honestly not sure what else to do here - is there anything in our shader itself we can speed up?

I don't know if there's anybody here who understands OpenGL who a) could review these changes and check I'm not doing anything silly or b) offer any suggestions on other improvements.

Would also be helpful if folks could download the jar from GH actions (or build locally) and test to see if nothing has broken on their graphics card - I seem to remember NVIDIA causing problems (or as they call it "being spec compliant") last time I went near the shader code.

Obviously this is all gratuitous at this point - I'm rendering 480 full-sized monitors (8k effective resolution, almost 2k blocks wide) at 100fps when idle, so there's really not much point in making this any faster.

Issue #1068, opened 2022-04-13
Block Setup Documentation - ingame or on website


LordHelmchen:
I was missing some basic block guides. The programming API is well documented, but without knowing what to look for it took quite a while to find out, that I have to right click a modem for it to work. I simply didn't understand why modems would not do anything or if I have to write messages or do remote calls on the modem or if I have to set up a computer next to each inventory/peripheral I want to modify.

https://github.com/cc-tweaked/CC-Tweaked/issues/1023#issuecomment-1018125204 has picture that explained basic networking and made sure, that I was on the right track. In the mentioned issue the modems were red, which was different from mine.

After I found out about it, I looked for someone else mentioning it:
http://www.computercraft.info/forums2/index.php?/topic/11244-151-using-network-cables-for-dummies/ is a small example that mentions this important fact.

A small block setup + wrap peripheral + transfer item example would make it much easier to understand how CC works in general.

Lupus590:
Adding to this, an example GPS setup would also be useful.

The documentation system does support wiki/guide like entries (got to tweaked.cc and scroll the index to the bottom), so one just needs to make a PR assign a new file to https://github.com/cc-tweaked/CC-Tweaked/tree/mc-1.16.x/doc/guides. Just a note that legally we can't copy from the old wikis, so someone who already knows how to do some of this stuff would have to write it.

Lupus590:
Going to try to make a couple of PRs covering this kind of stuff. It's probably going to take me a while though.

Lupus590:
What I'm currently writing:
* using startup files
* setting up GPS
* using modems for communication and connecting peripherals

I might do these as separate PRs

SquidDev:
Worth mentioning that there's _some_ stuff about monitors in the [introduction to the `peripheral` API](https://tweaked.cc/module/peripheral.html), but it's not great - could probably do with fleshing out or cross-referencing with other pages.

PR #1069, opened 2022-04-13
Document how to use startup files

Lupus590:
Related: #1068


doc/guides/startup_files.md
```diff
@@ -0,0 +1,53 @@
...
+
+To reiterate, startup files names are all lowercase. CC is looking for `startup` and `startup.lua` not `Startup` or any other caplitalisations.
+:::
+
+## Searching drives for startup files
+If `shell.allow_disk_startup` is true (which it is by default), then the CC computer will look for disk drives that have atleast one startup file, it only uses the first disk that fulfils this criteria. Disk drives are searched in the order they are found via `peripheral.getNames()`. It then runs the found startup files as described below.
+
+If `shell.allow_startup` is true (which it is by default) and no disk startup file was used, then the computer will look for startup files in it's internal storeage. Valid files are ran as described below.
```
DVD-DAVIDE, line 32: Typo on line 32: "storeage". Should be changed in "storage".

SquidDev:
Thanks for putting this together! I wonder if it's worth leaving off the section on case sensitivity, and just saying "keep it lower case". It's good content to have, but probably better off in the main `fs` API.

SquidDev (changes requested):
Thank you very much for putting this together! I always forget how terribly even the core bits of functionality are documented.

I think the content is good here, I'm just quite sure about the structure yet. What I'm trying to do with guides[^1] is provide something which is relatively incremental - users can follow along, stop at any point, but still have something _actionable_ they can take away from it.

I think right now things are a little confusing because so much is introduced in one go at the beginning, and it's not clear what 'I' can/should do[^2]. I think I'd propose splitting the docs into several sections like so:

 - **startup.lua:** Motivation (run program when a computer turns on) and a basic example: "create a `startup.lua` file and run the reboot program". We have some [funky support for mounting code snippets on computers](https://github.com/cc-tweaked/CC-Tweaked/blob/be45b718b32fb80f29e1e9ccba06a515bf3144b1/doc/guides/using_require.md?plain=1#L36)[^3], so you could have two snippets here: one with the startup file and another just saying "Press the run button to see the startup file in action".
 - **startup directory:** Talk about `startup` vs `startup.lua` and the `startup/` directory. We could even have another snippet here, talking about creating a `startup/another-startup.lua` file, mount that along with the previous startup file, and demo it again.
 - **Starting from disks:** Move that "Mass configuration" section to the top (though leave out the stuff on settings for now)! This is a great "why would I want to do this?".

   We can't really demo this in-browser, but can still explain how you'd do this (i.e. move your startup.lua and startup/ files onto a disk and reboot).

   At the end of this section it's probably worth talking about what happens when you've got multiple drives.

 - **Settings:** Mention the two `shell.*` settings. Possibly anything else you think is relevant?

[^1]: And to be clear, I'm not sure I've been successful yet, even in the stuff I wrote!

      Really I should be following something more structured like the [Diataxis Framework](https://diataxis.fr/), but I really struggle writing in a way which doesn't end up being three kinds of topic in one. Though I probably should move the Lua 5.2/5.3 compatibility stuff into a "Reference" section at least.

[^2]: As always, I'm the worst person to be commenting on docs, because I don't know what beginners need.

[^3]: Worth noting the example runner will override a startup.lua file, but that should be easy to fix - just [swap these two elements](https://github.com/cc-tweaked/CC-Tweaked/blob/mc-1.16.x/src/web/index.tsx#L105). Not sure if you've got CC:T's full build environment (and illuaminate) set up locally, so happy to do that bit if you'd prefer.

Lupus590:
> Not sure if you've got CC:T's full build environment (and illuaminate) set up locally

I'm fairly sure that I don't, the proper build steps break before doc gen so I've been calling that manually and it's not produced quite how I expect it to. Some CSS or markdown-to-html might not be working. I wouldn't mind pointers, can do it via discord if it's easier.

Edit: thanks

PR #1071, opened 2022-04-16
Add documentation on full list of 5.2/5.3 features

MCJack123:
This PR adds a new page of documentation on exactly which features in Lua 5.2/5.3 (as well as old 5.0 functions) are available in CC:T.

### DIFF
--- a/doc/guides/feature_compat.md
+++ b/doc/guides/feature_compat.md
@@ -9,10 +9,10 @@ CC: Tweaked is based off of the Cobalt Lua runtime, which uses Lua 5.1. However,
 | Feature | Supported? | Notes |
 |---------|------------|-------|
 | `goto`/labels | :x: |  |
+| `_ENV` | :large_orange_diamond: | The `_ENV` global points to `getfenv()`, but it cannot be set. |
 | `\z` escape | :heavy_check_mark: |  |
 | `\xNN` escape | :heavy_check_mark: |  |
 | hex literal fractional/exponent parts | :heavy_check_mark: |  |
-| `_ENV` | :large_orange_diamond: | The `_ENV` global points to `getfenv()`, but it cannot be set. |
 | empty statements | :x: |  |
 | `__len` metamethod | :heavy_check_mark: |  |
 | `__ipairs` metamethod | :x: |  |
@@ -20,7 +20,7 @@ CC: Tweaked is based off of the Cobalt Lua runtime, which uses Lua 5.1. However,
 | `bit32` library | :heavy_check_mark: | Replaces `bit` library, which is still available for compatibility. |
 | `collectgarbage` isrunning, generational, incremental options | :x: | `collectgarbage` does not exist in CC:T. |
 | new `load` syntax | :heavy_check_mark: |  |
-| `loadfile` mode parameter | :heavy_check_mark: |  |
+| `loadfile` mode parameter | :heavy_check_mark: | Supports both 5.1 and 5.2+ syntax. |
 | removed `loadstring` | :large_orange_diamond: | Only if `disable_lua51_features` is enabled in the configuration. |
 | removed `getfenv`, `setfenv` | :large_orange_diamond: | Only if `disable_lua51_features` is enabled in the configuration. |
 | `rawlen` function | :heavy_check_mark: |  |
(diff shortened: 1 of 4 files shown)


doc/guides/feature_compat.md
```diff
@@ -0,0 +1,95 @@
...
+| `goto`/labels | :x: |  |
+| `_ENV` | :large_orange_diamond: | The `_ENV` global points to `getfenv()`, but it cannot be set. |
+| `\z` escape | :heavy_check_mark: |  |
+| `\xNN` escape | :heavy_check_mark: |  |
+| hex literal fractional/exponent parts | :heavy_check_mark: |  |
+| empty statements | :x: |  |
+| `__len` metamethod | :heavy_check_mark: |  |
+| `__ipairs` metamethod | :x: |  |
+| `__pairs` metamethod | :heavy_check_mark: |  |
+| `bit32` library | :heavy_check_mark: | Replaces `bit` library, which is still available for compatibility. |
```
SquidDev, line 20: I don't think it's worth mentioning the `bit` library - that's not actually documented anywhere.

doc/guides/feature_compat.md
```diff
@@ -0,0 +1,95 @@
...
+| `debug.getlocal` negative indices for varargs | :x: |  |
+| `debug.getuservalue`/`debug.setuservalue` | :x: | Userdata are rarely used in CC:T, so this is not necessary. |
+| `debug.upvalueid` | :heavy_check_mark: |  |
+| `debug.upvaluejoin` | :heavy_check_mark: |  |
+| tail call hooks | :x: |  |
+| `=` prefix for chunks | :heavy_check_mark: |  |
+| yield across C boundary | :heavy_check_mark: |  |
+| removal of ambiguity error | :x: |  |
+| identifiers may no longer use locale-dependent letters | :heavy_check_mark: |  |
+| ephemeron tables | :x: | Weak tables are not supported. |
```
SquidDev, line 60: I thought weak tables were supported (as long as you set `__mode` before setting the metatable). If not, it's probably worth filing a bug on Cobalt.

doc/guides/feature_compat.md
```diff
@@ -0,0 +1,95 @@
...
+| functions may be reused | :x: |  |
```
SquidDev, line 61: What does this mean?

doc/guides/feature_compat.md
```diff
@@ -0,0 +1,95 @@
...
+| `table.move` | :x: |  |
+| `math.atan2` -> `math.atan` | :x: |  |
+| removed `math.frexp`, `math.ldexp`, `math.pow`, `math.cosh`, `math.sinh`, `math.tanh` | :x: |  |
+| `math.maxinteger`/`math.mininteger` | :x: |  |
+| `math.tointeger` | :x: |  |
+| `math.type` | :x: |  |
+| `math.ult` | :x: |  |
+| removed `bit32` library | :x: |  |
+| remove `*` from `file:read` modes | :heavy_check_mark: |  |
+| metamethods respected in `table.*`, `ipairs` | :heavy_check_mark: |  |
```
SquidDev, line 84: I don't think this is true. We use `__lt`, but don't use `__index`/`__len`.

SquidDev (changes requested):
Thanks for putting this together.

I kinda wonder if it's worth dropping the "removed XYZ" features (or shunting them to a separate set of tables). Most of the time they're less interesting (or confusing, in the case of `__ipairs`, as we didn't support it in the first place!)

doc/guides/feature_compat.md
MCJack123: PUC Lua 5.2 states that it may reuse the same function object if it has the same code and upvalues. Lua 5.1/Cobalt don't do this (yet).

Example:
```lua
local upval = 2
local function a(n)
  print(n + upval)
end
local function b(n)
  print(n + upval)
end
print(a == b) -- false on 5.1/Cobalt, true on 5.2
```

doc/guides/feature_compat.md
MCJack123: Ah, I couldn't find it when I was working on 5.2 - just found it, and I'll correct this.

doc/guides/feature_compat.md
MCJack123: Huh, I could've sworn it was there before, but just checked and it indeed isn't (since it's not an API, it's preloaded in the BIOS).

SquidDev (approved):

SquidDev:
CSS is hard, so this doesn't look as good as it should, but I still think worth merging. Are you happy with this Jack?

MCJack123:
LGTM :+1:

-----

<sup>Took me way too long to build illuaminate to see the docs properly. The AUR package really needs to be updated. It was complaining about the `lsp` package missing `unvendor.ml`, which is one of the files removed in the stripped version. Not sure why it's trying to run files that it shouldn't have?</sup>

SquidDev:
Ahh, hadn't even realised there was an AUR package. Yeah, those build steps are very out of date.

SquidDev:
Thank you!

PR #1075, opened 2022-04-19
Add info about DFPWM support in FFmpeg

MCJack123:
This PR adds some documentation about the recently added support for DFPWM audio in FFmpeg.


SquidDev:
I kinda wonder if it's just worth waiting until a new ffmpeg version is released.

MCJack123:
I don't see that happening until at least September, considering the last release just came out three months ago and they average just over one release per year.

SquidDev:
Ahh fair enough. Maybe worth moving this later in the list then. I'm trying to list the recommended (or at least easiest) method first, and using a dev version of ffmpeg is definitely not that :p.

SquidDev:
It's out!

PR #1078, opened 2022-04-24
Fixed typo

Hasaabitt:
"Instead, it is a standard program, which its API into the programs that it launches."
becomes
"Instead, it is a standard program, which injects its API into the programs that it launches."


SquidDev:
Thanks!

Issue #1082, opened 2022-04-30
CC Blocks do not rotate in structures


Seniorendi:
### Minecraft Version

1.16.x

### Version

1.100.5

### Details

I have this problem also for 1.18

I use blocks from CC:T for my structures in AP. But they don't rotate when the feature is placed. 
The block classes([BlockComputer](https://github.com/cc-tweaked/CC-Tweaked/blob/mc-1.16.x/src/main/java/dan200/computercraft/shared/computer/blocks/BlockComputer.java) for example) do not override the `mirror` method, so they don't rotate. See [here](https://github.com/Seniorendi/AdvancedPeripherals/issues/284) for context

Seniorendi:
![image](https://user-images.githubusercontent.com/67484093/166107778-4daf02f9-6892-4b8d-99da-4917ecc97abd.png)
👀 
approved

SquidDev:
You know you want that sweet "Contributor" badge!

PR #1083, opened 2022-04-30
[1.16] Make blocks rotatable for structures

Seniorendi:
For further explanation, see #1082 
Every block should now be rotatable for structures. If you see that I missed one, just tell me (:
I also updated forge so that it supports newer versions of java

I'll open a PR for 1.18 in the next days

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,5 +4,5 @@ mod_version=1.100.5
 # Minecraft properties (update mods.toml when changing)
 mc_version=1.16.5
 mapping_version=2021.08.08
-forge_version=36.2.20
+forge_version=36.2.34
 # NO SERIOUSLY, UPDATE mods.toml WHEN CHANGING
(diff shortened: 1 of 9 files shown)


Seniorendi:
image
Left is normal, right is rotated by 270°

SquidDev (changes requested):
Looks good, thank you very much! Just a couple of small nitpicks:

 - We tend to use javax's Nonnull annotation instead of Jetbrain's one. JB's is probably better to use nowadays, but I do want to keep things consistent.
 - Can you remove the `p` prefix from parameters? I think Parchment or ForgeGradle adds them in :/.

I assume these rotations only apply around the Y axis - we don't need to worry about a computer trying to face UP/DOWN?

Don't worry about 1.18 (unless it has changed massively!), as I can do that when merging things across.

Seniorendi:
> I assume these rotations only apply around the Y axis - we don't need to worry about a computer trying to face UP/DOWN?

Computers shouldn't be able to rotate UP/DOWN
So I just *hope* this will never happen :P

> * We tend to use javax's Nonnull annotation instead of Jetbrain's one. JB's is probably better to use nowadays, but I do want to keep things consistent.
> * Can you remove the `p` prefix from parameters? I think Parchment or ForgeGradle adds them in :/.

Done (:

> Don't worry about 1.18 (unless it has changed massively!), as I can do that when merging things across.

Kay!
Nothing changed in between 1.16 to 1.18

SquidDev (approved):

SquidDev:
Thank you very much!

PR #1092, opened 2022-05-08
Added use_empty_json_array to textutils.unserialiseJSON

ChickChicky:
It  can be useful for users who are unaware that empty json arrays are replaced by `empty_json_array` instead of an empty LUA table, which can lead to unexpacted behaviors when using `table.insert` for example.
Adding on option to enable the usage of the `empty_json_array` can help with this, as well as better documentation about it.


SquidDev:
Renamed the option and added some docs, but otherwise looks good to me. Thanks!

PR #1100, opened 2022-05-23
Make ComputerScreenBase methods extendable

Lemmmy:
Incredibly minor tweak, removes `final` from all the methods in ComputerScreenBase to allow you to extend it, particularly methods like `init()`.

Of course this isn't stable API but it is handy to have access to the existing implementation for file D&D and the sidebar buttons.


SquidDev:
Thanks :)

PR #1101, opened 2022-05-23
Add Detail Providers

Lemmmy:
This pull request implements a new API to register "detail providers" - an interface that allows you to inject additional metadata for blocks, fluids, and items. The interface is called whenever *full* details are requested for a given object.

Currently, this affects the following methods:

Providing for **BlockReference**
- `turtle.inspect()`
- `turtle.inspectUp()`
- `turtle.inspectDown()`
- `command.getBlockInfo(...)`
- `command.getBlockInfos(...)`

Providing for **ItemStack**
- `turtle.getItemDetail(slot, detailed=true)`
- `inventory.getItemDetail(slot)`

Currently, fluids are supported by the API, but CC:Tweaked does not invoke `FluidData.fill` anywhere, so there is no way to get fluid metadata.

# Example usage

Basic usage of the API to add metadata to all items:

1. Create a class that implements `IDetailProvider`, targeting a `BlockReference`, `FluidStack` or `ItemStack`.
    ```java
    public final class ExampleItemDetailProvider implements IDetailProvider<ItemStack> {
        public void provideDetails(@Nonnull Map<? super String, Object> data, @Nonnull ItemStack stack) {
            // `stack` is an instance of the ItemStack being inspected.
            // You can get the underlying item with `stack.getItem()`.
            
            // New properties can be added by modifying the supplied `data` map.
            data.put("example", "value"); // Adds a new `example` key to ALL items.
        }
    }
    ```
    
2. Register the provider with `ComputerCraftAPI.<T>registerDetailProvider(Class<T>, IDetailProvider<T>)`. This should be done during mod initialization.
    ```java
    ComputerCraftAPI.registerDetailProvider(ItemStack.class, new ExampleItemDetailProvider());
    ````

For ItemStacks, a helper class (`BasicItemDetailProvider`) has been provided that allows you to filter the underlying item. It also allows you to specify a namespace, to automatically push all keys to a new sub-table in the details object. Here is an example item detail provider for printed pages:

```java
public class PrintoutItemDetailProvider extends BasicItemDetailProvider<ItemPrintout> {
    public PrintoutItemDetailProvider() {
        // The details we provide will be placed in a `printout` sub-table.
        // This provider will be filtered to instances of `ItemPrintout`.
        super("printout", ItemPrintout.class);
    }

    @Override
    public void provideDetails(@Nonnull Map<? super String, Object> data, @Nonnull ItemStack stack,
                               @Nonnull ItemPrintout printout) {
        // As the provider is now filtered, we are given an instance of our
        // ItemPrintout as a convenience.
        data.put("type", printout.getType().toString());
        data.put("title", ItemPrintout.getTitle(stack));
        data.put("pages", ItemPrintout.getPageCount(stack));

        Map<Integer, String> lines = new HashMap<>();
        String[] lineArray = ItemPrintout.getText(stack);
        for (int i = 0; i < lineArray.length; i++) {
            lines.put(i + 1, lineArray[i]);
        }
        data.put("lines", lines);
    }
}
```

# Notes and questions

- Fluid providers are supported but do not do anything right now, because `FluidData.fill` is never called anywhere (except possibly in peripheral mods?)
- In 1.16 there was support for augmenting the turtle inspection methods by hooking `TurtleInspectItemEvent` and `TurtleBlockEvent.Inspect` (though this was dropped in 1.17). This PR is essentially a different approach at achieving the same thing, but it works for inventory methods and command computers too. It does not conflict with it, but it may be a little silly.
    - Should this PR work in the same way, using events instead of using a registry system? This would allow cancellation of the provider altogether, but may incur more boilerplate.


src/main/java/dan200/computercraft/api/ComputerCraftAPI.java
```diff
@@ -221,6 +222,21 @@ public static void registerAPIFactory( @Nonnull ILuaAPIFactory factory )
...
+     * Registers a detail provider to provide additional details for blocks, fluids and items when inspected by methods
+     * such as {@code turtle.getItemDetail()} or {@code turtle.inspect()}.
+     *
+     * @param provider The detail provider to register.
+     * @param <T> The type of object that this provider can provide details for.
+     * @see dan200.computercraft.api.detail.IBlockDetailProvider
+     * @see dan200.computercraft.api.detail.IFluidDetailProvider
+     * @see dan200.computercraft.api.detail.IItemDetailProvider
+     */
+    public static <T> void registerDetailProvider( @Nonnull IDetailProvider<T> provider )
```
SquidDev: Rather than having three sub-interfaces, I wonder if it's just cleaner to only have `IDetailProvider<T>` and either have three kinds of registration method, or just a `registerDetailProvider(Class<T>, IDetailProvider<T>)`. This has the added advantage that you can register with a lambda without having to specify a class.

src/main/java/dan200/computercraft/shared/Registry.java
```diff
@@ -376,6 +377,9 @@ private static void registerProviders()
         ComputerCraftAPI.registerGenericCapability( CapabilityEnergy.ENERGY );
         ComputerCraftAPI.registerGenericCapability( CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY );
 
+        // Register detail providers
+        ComputerCraftAPI.registerDetailProvider( new PrintoutItemDetailProvider() );
```
SquidDev: Sorry to be a pain, but can we remove this? I think I just want to add the API and nothing else, as it saves me having to think about what's exposed to users (as CraftOS has more stability guarantees than the Java API).

src/main/java/dan200/computercraft/shared/peripheral/generic/data/DetailProviders.java
```diff
@@ -0,0 +1,79 @@
...
+
+        for ( IDetailProvider<T> provider : providers )
+        {
+            try
+            {
+                provider.provideDetails( data, value );
+            }
+            catch ( Exception e )
+            {
+                ComputerCraft.log.error( "Error while providing details for {}", value, e );
```
SquidDev: I know we do this for peripherals, but I think we should probably just bubble the exception. Better to signal an error to the computer than silently drop data.

SquidDev (changes requested):
Thank you for putting this together. Left some comments on more specific things.

> There is already support for augmenting the turtle inspection methods by hooking TurtleInspectItemEvent and TurtleBlockEvent.Inspect. 

These are removed in 1.17+, so not too worried.

> Fluid providers are supported but do not do anything right now, because FluidData.fill is never called anywhere (except possibly in peripheral mods?)

:flushed: Woops.

> Block providers only provide details for BlockStates and Blocks. Since they (and BlockData.fill) are not passed a World or Position.

Maybe we bundle world, position, block state and block entity into a single class and get data on that instead, like Plethora's `BlockReference`. I don't think there's ever a time when we'll need to get metadata on a block state not in the world. Right? Right???

Lemmmy:
Thank you for the feedback, I have made the requested changes and updated the original post to reflect the current state of the API.

src/main/java/dan200/computercraft/shared/peripheral/generic/data/DetailProviders.java
```diff
@@ -0,0 +1,44 @@
...
+{
+    private static final Map<Class<?>, Collection<IDetailProvider<?>>> allProviders = new HashMap<>();
+
+    public static synchronized <T> void registerProvider( Class<T> type, IDetailProvider<T> provider )
+    {
+        Objects.requireNonNull( type, "type cannot be null" );
+        Objects.requireNonNull( provider, "provider cannot be null" );
+
+        if ( !type.isAssignableFrom( BlockReference.class ) && !type.isAssignableFrom( ItemStack.class )
+            && !type.isAssignableFrom( FluidStack.class ) )
```
SquidDev: Possibly worth using `==` here instead. I realise all three classes are final, but expresses the intent a little better.

src/main/java/dan200/computercraft/shared/peripheral/generic/data/ItemData.java
```diff
@@ -98,6 +98,12 @@ public static <T extends Map<? super String, Object>> T fill( @Nonnull T data, @
             data.put( "unbreakable", true );
         }
 
+        /*
+         * Execute the detail providers to fill additional data if any.
+         * @see IItemDetailProvider
+         */
```
SquidDev: I think we can probably delete these comments. Not sure they add much now that `IItemDetailProvider` (and the like) are gone.

src/main/java/dan200/computercraft/api/detail/BlockReference.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Nonnull
+    public BlockState getState()
+    {
+        return state;
+    }
+
+    @Nullable
+    public TileEntity getTile()
```
SquidDev: Should probably call this `getBlockEntity`, just so it's consistent with MojMap.

SquidDev (changes requested):
Thank you so much! Just a couple of nitpicks, but otherwise looks good.

SquidDev (approved):

Lemmmy:
Thank you!

SquidDev:
:heart:  Thank you.

PR #1112, opened 2022-06-16
Add WAV support to speaker.lua

MCJack123:
This PR adds basic support for playing WAV files through speaker.lua. It supports playing either 8-bit PCM or DFPWM (following the spec [here](https://gist.github.com/MCJack123/90c24b64c8e626c7f130b57e9800962c)) WAV files, at 48 kHz and single channel. It does not attempt any resampling or channel splitting, and will continue to play files with the wrong parameters with a warning. (For playing files with multiple channels or a different sample rate (and more!), see my [AUKit](https://github.com/MCJack123/AUKit) library & program set.)

This is accomplished by checking if the first four bytes of the file match a WAV file. If so, it will read the format chunk (to validate the file's compatibility and check PCM or DFPWM), then skip forward to the data chunk. After that, the decoding's the same as before (with an additional decoder for unsigned PCM). If the file's plain DFPWM, it will save the four bytes it read and add it on to the first chunk.

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/programs/fun/speaker.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/fun/speaker.lua
@@ -18,6 +18,14 @@ local function get_speakers(name)
     end
 end
 
+local function pcm_decoder(chunk)
+    local buffer = {}
+    for i = 1, #chunk do
+        buffer[i] = chunk:byte(i) - 128
+    end
+    return buffer
+end
+
 
 local cmd = ...
 if cmd == "stop" then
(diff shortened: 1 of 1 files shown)


src/main/resources/data/computercraft/lua/rom/programs/fun/speaker.lua
```diff
@@ -40,12 +48,60 @@ elseif cmd == "play" then
...
+        until chunk == "data"
+
+        handle.read(4)
+        start = nil
+    end
+
     print("Playing " .. file)
 
-    local decoder = require "cc.audio.dfpwm".make_decoder()
+    local decoder = pcm and pcm_decoder or require "cc.audio.dfpwm".make_decoder()
```
Lupus590, line 100: Seeing this makes me think that we should have a "cc.audio.pcm" with its own `make_decoder()`.

I have no idea how practical that might be though, I'm not an audio person.

MCJack123: PCM decoding (at least unsigned 8-bit) is as simple as using `string.byte` and subtracting 128. Signed formats can all be accomplished with a loop of `string.unpack` and division. It's really very few lines of code - I don't really see the need for it (except maybe helping some newbies, but at that point just copy and paste it).

Also, no need to have encoder/decoder objects - PCM is stateless. DPCM/ADPCM ((adaptive) differential PCM) would, but those are not as efficient as DFPWM anyway.

src/main/resources/data/computercraft/lua/rom/programs/fun/speaker.lua
```diff
@@ -40,12 +48,60 @@ elseif cmd == "play" then
...
+    local size = pcm and 16 * 1024 * 8 or 16 * 1024 - 4
     while true do
-        local chunk = handle.read(16 * 1024)
+        local chunk = handle.read(size)
         if not chunk then break end
+        if start then
+            chunk, start = start .. chunk, nil
```
SquidDev, line 105: Do you need to bump `size` by 4 here too? Maybe also cleaner to set `size` within the earlier `if` clauses - you only want the `-4` when consuming raw DFPWM, not DFPWM inside WAV.

I also wonder if it'd be nicer just to seek back to the beginning of the file when its not a WAV, rather than concatenating the two values. Not sure, this sort of code always ends up being a little gross.

SquidDev (changes requested):
Thank you! One small bug/question about size handling, but otherwise looks good.

src/main/resources/data/computercraft/lua/rom/programs/fun/speaker.lua
SquidDev: I do agree with Lupus here, if only for the header parsing. However, I'm not sure what the API would look like: you really need to consume a file handle, which isn't really something we do anywhere else in CC.

IDK, I retired (lol) so I didn't have to make any more decisions. Let's leave it as is for now and can always change it if someone comes up with a nice design.

src/main/resources/data/computercraft/lua/rom/programs/fun/speaker.lua
```diff
@@ -40,12 +48,60 @@ elseif cmd == "play" then
...
+
+        local fmtsize = ("<I4"):unpack(handle.read(4))
+        local fmt = handle.read(fmtsize)
+        local format, channels, rate, _, _, bits = ("<I2I2I4I4I2I2"):unpack(fmt)
+        if not ((format == 1 and bits == 8) or (format == 0xFFFE and bits == 1)) then
+            handle.close()
+            error("Could not play audio: Unsupported WAV file", 0)
+        end
+        if channels ~= 1 or rate ~= 48000 then
+            print("Warning: This WAV file may not play correctly.")
```
SquidDev: Maybe there should be a little more detail in this message? Something like "Only WAV files with 1 channel sampled at 48kHz are supported. This file may play incorrectly". 

Not sure, it's a little verbose!

MCJack123: I shortened it a little bit, but hopefully people can look up what it means (or ask for help - preferably not in your DMs though!).

MCJack123:
Oops, I started editing this a while ago, but then got sidetracked :P - the changes are now implemented.

SquidDev:
Thanks! The handling of the various formats is still not wonderful, but not really sure how to make it cleaner honestly. 

Time to write a parser combinator library for binary file formats or something!

Lupus590:
> Time to write a parser combinator library for binary file formats or something!

[AUKit](https://github.com/MCJack123/AUKit)?

SquidDev:
> [AUKit](https://github.com/MCJack123/AUKit)?

My complaints are very much in the actual parsing, not the audio handling. AUKit's parsing is definitely more rigorous, but I wouldn't say it's any nicer than what's here :D. That sort of thing is just very awkward to write Correctly(TM).

PR #1125, opened 2022-06-29
Inventory peripheral API tweaks

itisluiz:
Adds a method for swapping item stacks between slots/inventories without the need for a buffer slot and now returns the item groups (Creative tabs) an item pertains to when calling **getItemDetail**().

Gives the generic inventory api some love, allowing for more elaborate and faster systems like item sorting. The **swapItems** method isn't completely failsafe, just like the current **pushItems** and **pullItems** aren't, as there's no real great way to do it. This should allow for more dynamic item movement, as it will no longer require a buffer slot, which currently makes thing slower and more cumbersome. Returns the creative tabs an item is part of in an effort to make the **getItemDetail**() method return data a bit more detailed.

### DIFF
--- a/src/main/java/dan200/computercraft/shared/peripheral/generic/data/ItemData.java
+++ b/src/main/java/dan200/computercraft/shared/peripheral/generic/data/ItemData.java
@@ -10,6 +10,7 @@ import dan200.computercraft.shared.util.NBTUtil;
 import net.minecraft.enchantment.Enchantment;
 import net.minecraft.enchantment.EnchantmentHelper;
 import net.minecraft.item.EnchantedBookItem;
+import net.minecraft.item.ItemGroup;
 import net.minecraft.item.ItemStack;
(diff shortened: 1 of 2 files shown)


itisluiz:
I'm aware that ItemTransfer is already regarded as overpowered, but I don't believe this to make it even more so, it just makes what was already possible a bit less cumbersome and less compute-intensive. Perhaps in the future there could be a balancing patch that tackles all of the inventory api.

SquidDev:
Thanks for the PR! I'm afraid I'm not going to merge this - I'm pretty happy with the API surface exposed by the inventory API right now (balance notwithstanding :p), in that it exposes the primitives and nothing else, and so don't really want to add to it.

Happy to pull in the item group changes though (CC @Lemmmy here - I know you've been doing stuff with item data here, probably happy to merge in some of your extensions too).

PR #1127, opened 2022-07-02
ItemGroup info on getItemDetail

itisluiz:
Get information (id, displayName) pairs for all ItemGroups (creative tabs) an item pertains to. Very useful for categorizing items.

This is similar to what my recently closed (due to the api changes) pr had, but it now returns both the ID and display name of a given item group. Could go back to returning just the IDs if it feels too bloated. Also updates the documentation accordingly and adds an example of usage.


src/main/java/dan200/computercraft/shared/peripheral/generic/data/ItemData.java
```diff
@@ -116,6 +118,28 @@ private static ITextComponent parseTextComponent( @Nonnull INBT x )
...
+    @Nonnull
+    private static List<Map<String, Object>> getAllItemGroups( @Nonnull ItemStack stack )
+    {
+        ArrayList<Map<String, Object>> groups = new ArrayList<>( 1 );
+
+        for ( ItemGroup g : stack.getItem().getCreativeTabs() )
+        {
+            Map<String, Object> group = new HashMap<>( 2 );
+            group.put( "id", g.getId() );
+            group.put( "displayName", g.getDisplayName().getString() );
```
SquidDev: On 1.16, `getDisplayName` is client only. I think we'll need an access transformer here to access the underlying `displayName` field.

src/main/java/dan200/computercraft/shared/peripheral/generic/data/ItemData.java
```diff
@@ -116,6 +118,28 @@ private static ITextComponent parseTextComponent( @Nonnull INBT x )
...
+     */
+    @Nonnull
+    private static List<Map<String, Object>> getAllItemGroups( @Nonnull ItemStack stack )
+    {
+        ArrayList<Map<String, Object>> groups = new ArrayList<>( 1 );
+
+        for ( ItemGroup g : stack.getItem().getCreativeTabs() )
+        {
+            Map<String, Object> group = new HashMap<>( 2 );
+            group.put( "id", g.getId() );
```
SquidDev: It feels like it's probably more useful to access expose the `langId` rather than the numerical ID - AFAIK the latter can change between runs.

SquidDev (changes requested):

SquidDev:
Thanks! Just left a couple of comments, but otherwise looks good.

src/main/java/dan200/computercraft/shared/peripheral/generic/data/ItemData.java
itisluiz: No longer uses `getDisplayName` or `getId`, just the exposed `langId`

src/main/java/dan200/computercraft/shared/peripheral/generic/data/ItemData.java
itisluiz: Done, now exposes `langId` through an access transformer and returns a table of them, in a similar way to item `tags`.

itisluiz:
Thanks for the feedback! Sorry for making the pull request so all over the place for such a simple change.

SquidDev (approved):
Thanks for the quick change, sorry for being so slow on my end!

PR #1131, opened 2022-07-07
Fix off-by-1 error in generic inventory's getItemLimit()

heap-underflow:
It's difficult to think of a situation where this would matter or cause an issue except for the one I ran into, trying to use a StorageDrawers Drawer Controller as a generic inventory peripheral.

But when you do:
- `getItemLimit(0)` throws (fine)
- `getItemLimit(1)` returns the stack limit of the *second* drawer
- `getItemLimit(n)` returns 0.

Test setup:
image

Test script:
```lua
local dc = peripheral.wrap("right")
for slot,stack in pairs(dc.list()) do
    print(slot, stack.name, stack.count, dc.getItemLimit(slot))
end
```

Results (no upgrades):
image

Results (with a storage upgrade in the middle drawer -- the one with 2 cobble):
image

In this specific case, the user can kind of work around this by instead calling `dc.getItemLimit(slot - 1)`, but this seems to rely on the fact that StorageDrawers starts their drawer indexing at 1 (or is using 0 for the controller or something).  (Note: When I say "1" here I'm talking about in the 0-based Java world, and therefore 2 in the 1-based Lua world as shown in first column of the the screenshot output)

This is a PR against the 1.16.x branch (since that's the version I ran into the issue on), but it looks like it exists across (at least some) others, e.g. [1.19.x](https://github.com/cc-tweaked/CC-Tweaked/blob/mc-1.19.x/src/main/java/dan200/computercraft/shared/peripheral/generic/methods/InventoryMethods.java#L171)

### DIFF
--- a/src/main/java/dan200/computercraft/shared/peripheral/generic/methods/InventoryMethods.java
+++ b/src/main/java/dan200/computercraft/shared/peripheral/generic/methods/InventoryMethods.java
@@ -166,7 +166,7 @@ public class InventoryMethods implements GenericPeripheral
     public static int getItemLimit( IItemHandler inventory, int slot ) throws LuaException
     {
         assertBetween( slot, 1, inventory.getSlots(), "Slot out of range (%s)" );
-        return inventory.getSlotLimit( slot );
+        return inventory.getSlotLimit( slot - 1 );
     }
 
     /**


heap-underflow:
Just to be clear, this is just to mirror the index shifting that's already done (correctly) in:
- [list()](https://github.com/cc-tweaked/CC-Tweaked/blob/mc-1.16.x/src/main/java/dan200/computercraft/shared/peripheral/generic/methods/InventoryMethods.java#L99)
- [getItemDetail()](https://github.com/cc-tweaked/CC-Tweaked/blob/mc-1.16.x/src/main/java/dan200/computercraft/shared/peripheral/generic/methods/InventoryMethods.java#L141)
- [pushItems()](https://github.com/cc-tweaked/CC-Tweaked/blob/mc-1.16.x/src/main/java/dan200/computercraft/shared/peripheral/generic/methods/InventoryMethods.java#L216)
- [pullItems()](https://github.com/cc-tweaked/CC-Tweaked/blob/mc-1.16.x/src/main/java/dan200/computercraft/shared/peripheral/generic/methods/InventoryMethods.java#L263)

SquidDev:
Good catch, thank you!

PR #1139, opened 2022-07-22
Clearify GPS documentation note

piprett:
The GPS note was confusing and wrong. I fixed that here :)
Hopefully I did everything right, please comment if I dident.

## A quick checklist


Lupus590:
The note as I wrote it assumes that the user is following the guide and so the modem will be placed on the computer.

I suppose that this is a good thing to clarify, good catch.

SquidDev:
I assume this is discussing the case where modems are placed against wired modems, rather than computers?

piprett:
> I assume this is discussing the case where modems are placed against wired modems, rather than computers?

yes, like this:
image

PR #1156, opened 2022-08-28
Fixed usage example of textuils.pagedTabulate

### LINKED ISSUE #1149: Fix textutils.pagedTabulate example on tweaked.cc
As the title says, i noticed that the entry for textutils.pagedTabulate uses textutils.tabulate instead, thus making the example invalid.
here's the link to the entry:[https://tweaked.cc/module/textutils.html#v:tabulate](https://tweaked.cc/module/textutils.html#v:tabulate).

IvoLeal72:
I believe this fixes #1149


SquidDev:
Thank you so much!

I think this might still be badly behaved, but that's due to some other bugs in pagedTabulate (#487), not the docs itself :)

PR #1170, opened 2022-09-22
Fixed Issue #1167

### LINKED ISSUE #1167: snowy grass isn't updating
### Minecraft Version

1.18.x

### Version

1.100.9

### Details

if you move a turtle over snowy grass it will not update and wont make it too normal grass
Minecraft_ 1 18 2 - Singleplayer 12 09 2022 17_53_05

roland-a:
https://github.com/cc-tweaked/CC-Tweaked/issues/1167

### DIFF
--- a/src/main/java/dan200/computercraft/shared/turtle/core/TurtleBrain.java
+++ b/src/main/java/dan200/computercraft/shared/turtle/core/TurtleBrain.java
@@ -24,6 +24,7 @@ import dan200.computercraft.shared.util.HolidayUtil;
 import dan200.computercraft.shared.util.InventoryDelegate;
 import net.minecraft.block.Block;
 import net.minecraft.block.BlockState;
+import net.minecraft.block.GrassBlock;
 import net.minecraft.block.material.PushReaction;
 import net.minecraft.entity.Entity;
 import net.minecraft.entity.MoverType;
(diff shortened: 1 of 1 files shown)


SquidDev:
Thank you very much for looking into this! I think the fix looks sensible, but I fear it's a bit too specific - it's fixing a specific symptom rather than the underlying cause.

The problem here is that we don't trigger a block update when setting the block turtle initially (the flags parameter is 0, which means do nothing).

https://github.com/cc-tweaked/CC-Tweaked/blob/c3b7302108928fa6b9920c75c28121d67dea70b3/src/main/java/dan200/computercraft/shared/turtle/core/TurtleBrain.java#L327

I think there's two solutions which might be worth trying here:
 - Change the `0` there to a `1`. This will trigger neighbour updates when the turtle is initially set. We'll want to make sure that this doesn't cause https://github.com/dan200/ComputerCraft/issues/131 to occur again though.
 - Add a call to `world.blockUpdated(pos, newState.getBlock());` after `newTurtle.updateInputsImmediately();`.

roland-a:
I did assume it wasn't block-update related and that it was specific to snowy grass, as I saw double tall grass update when the turtle broke its top half, but not for snowy grass. Also, for some reason, neither of those two solutions fixed the issue when I added them on my workspace.

SquidDev:
Aha, because it does the update inside `updateShape` which isn't called `blockUpdated`. Doing `world.setBlock( pos, newState, 1 )` _does_ work, but doesn't broadcast the state to the client. Changing that to `3` instead should work.

roland-a:
Actually, flag 2 is enough to fix it.  Flag 2 is what broadcasts the state to the client.  Flag 3 is flag 1 and 2 together, which could cause the ghost block error you mentioned.

roland-a:
The underlying cause should be fixed now.

SquidDev (approved):

SquidDev:
Sorry for the delay in looking at this. Thank you very much for bearing with me :)

PR #1172, opened 2022-09-29
Always add HTTP programs to the path

SquidDev:
An attempt to make the UX better when HTTP is disabled. See the discussion at https://github.com/SquidDev-CC/FAQBot-CC/pull/44.

The programs already failed with the existing message:

wget requires the http API. Set http.enabled to true in CC: Tweaked's config

I'm not sure if this is as clear as it should be. Don't know if anyone has alternative phrasings?

### DIFF
--- a/src/main/resources/data/computercraft/lua/rom/programs/http/pastebin.lua
+++ b/src/main/resources/data/computercraft/lua/rom/programs/http/pastebin.lua
@@ -13,8 +13,8 @@ if #tArgs < 2 then
 end
 
 if not http then
-    printError("Pastebin requires the http API")
-    printError("Set http.enabled to true in CC: Tweaked's config")
+    printError("Pastebin requires the http API, but it is not enabled")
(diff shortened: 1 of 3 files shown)


Lupus590:
I think mentioning that HTTP is disabled could help. And that the option is on the server side.

So something like:
```
wget requires the HTTP API which is currently disabled.
Set http.enabled to true in CC: Teraked's server-side config.
```

SquidDev:
Yeah. I also wondered about that. My concern here is making the message too verbose, as (especially on turtles) there's very limited space.

I think adding something about server config is definitely worth mentioning, need to have a think about "which is currently disabled."

Lupus590:
Testing in copycat I get this, which seems alright, there's enough space on the last lines that I think it'll be fine on turtles.
It does take up a lot of space on the screen but I think that's more because CC has tiny screens than the error being too verbose.

image

PR #1182, opened 2022-10-08
Add all but 3 of the missing dutch translations

Quezler:
Skipped these 3, since honestly i couldn't think of something short & catchy in my native language:
```
gui.computercraft.tooltip.turn_on.key
gui.computercraft.tooltip.turn_off.key
gui.computercraft.tooltip.terminate.key
```

(the equivalent of "hold <keycombination>" would be "houd <keycombinatie> ingedrukt", and that's visually just ugly)


SquidDev:
Thank you!

PR #1188, opened 2022-10-21
Expose getters for the detail registry too

SquidDev:
This allows peripheral mods to use the same mechanism as CC for exposing item details (and blocks and fluid stacks).

This also replaces class-based registration with named registries:

```java
// Before
ComputerCraftAPI.registerDetailProvider(ItemStack.class, f);

// After
DetailRegistries.ITEM_STACK.addProvider(ItemStack.class, f);
```

CCing @Lemmmy here - is this still okay for your needs with Plethora?


Lemmmy:
LGTM, I like this a lot more

PR #1190, opened 2022-10-26
Handle file transfers inside CraftOS

### LINKED ISSUE #974: File uploads security concern.
CC: Tweaked 1.97.0 added ability to upload files via simple drag and drop method directly into computer file system.

It was pointed out to me today that it allows someone with access to gui to overwrite the startup file even if current state of terminal would normally don't allow them to have any meaningful interaction. (Per example: running program that ignores user input and terminate event)

Would it make sense to either give ability from lua code level to disable drag and drop mechanic to disallow adding/overwriting files.
Or change how it works so that drag dropping file is a event that is handled by active shell prompt instead of something that happens in background. So uploading file(s) will only work in shell or if user of alternative shell supports/allows it?

Or is this whole thing not something that should really be considered an issue?

SquidDev:
Closes #974. There's some polishing up needed to do here, but otherwise I think this is ready:

When a file is dropped on a computer, we now queue a `file_transfer` event. This has a single argument which contains the files which have been transferred.

```lua
local event, arg = os.pullEvent("file_transfer")
local files = arg.getFiles()
for _, file in ipairs(files) do
  print(file.getName()) -- Get the file name
  print(file.readLine()) -- Files are otherwise just binary file handles.
end
```

We now add an `import` program (open to changing the name here, I'm not wild about it) which writes transferred files to the current directory:

A screenshot of a ComputerCraft running the "import" program. We're trying to upload a .log file - the first time works, and the second fails due to trying to overwrite a file.

Definitely not wild about the UI, so ideas on how to make it even slightly legible are welcome :).

If a file_transfer event is not handled within 5 seconds (i.e. `getFiles()` is called), we send a toast to the user telling them to run the `import` program. Hopefully this'll make a reasonable transfer path for people used to the existing, more-seamless system.

A screenshot of Minecraft with a toast on the top right.

Still to do:
 - [x] Documentation. I think we'll need to patch cct-javadoc (and maybe illuaminate, ughr) to handle everything here.
 - [x] 0 nag delays should mean the toast doesn't show. I meant to do this and forgot.
 - [x] Bake `import` into the default shell, so that transferring a file while at the prompt will interrupt typing and upload files instead.
 - [x] Tests.


Lupus590:
> open to changing the name here

* `drop_copy`
* `forth_wall_breaker`
* `player_drop`

> Definitely not wild about the UI, so ideas on how to make it even slightly legible are welcome :).

Colours will help I think, even the different shades of grey that the basic computers have access to will help.

Definitely a good step in the right direction for security, having the shell automatically call the import program when it gets the event will help migrate users to this new more secure system. I like the toast notification too, it's a little bit disappointing that we need it now but it's better that we have it than if we didn't.

SquidDev:
> `forth_wall_breaker`

Sshhhhhhh. I'm not wild about it either. I guess _technically_ someone could write a shim which converts (signed/authenticated) rednet messages to `file_transfer` events too. But yes, the "drop files" is definitely a compromise I'm having to make for the sake of UX.

SquidDev:
Have changed the file list to cyan, which helps a bit. This also demonstrates shell integration:



And on normal computers:

SkyTheCodeMaster:
Something I think would be nice is another argument for the username/uuid of the player that dropped the files, for tracking purposes maybe? (Like if someone had a system log that tracks events it can list who dropped what files)

SkyTheCodeMaster:
Also, something that I think might be a valid way to go about this is that the second return of the event is just a direct table of:
```lua
{
  ["file.lua"]="contents\nof\nthe\nfile",
}
```
So that someone can just 
```lua
for name,content in pairs(arg) do
  local f = fs.open(name,"w")
  f.write(content)
  f.close()
end
```

SquidDev:
> second return of the event is just a direct table of:

So I did debate some alternative argument types, but settled on this for a couple of reasons:

 - I want/need a way to detect if the event is consumed within Lua code (so we can decided whether to trigger a toast or not), hence the `getFiles` method. It's definitely a little ugly how implementation requirements are leaking into the API, but I'll cope.
 
   Guess it makes it easier to extend the event in the future, though not sure when that would be needed!

 - List vs map really could go either way. I guess my thinking here is that you won't need to (or at least rarely) to index into the map, and so there's no strong need for it.

   _Technically_ it's possible to upload two files with the same name (it's pretty easy to do this using the "Recent" tab on a file browser), but given the `import` program doesn't support that, I'm not _too_ fussed.

 - RE: file handle vs raw string, I think file handles are the right choice here. It's consistent with what we do for http responses, and does give more flexibility expanding the API (though like with `getFiles`, not sure I'll need it).

SquidDev:
Merging this now so I can start pulling it into the other MC versions. Still happy to hear folk's thoughts on naming, etc... - can always change stuff before the release!

PR #1193, opened 2022-10-29
Add `peripheral_hub` type for wired-modem-like peripherals

### LINKED ISSUE #902: Allow different peripheral act like modem

SquidDev:
This allows other mods to create wired-modem alike blocks, which expose
peripherals on the wired network, without having to reimplement the main
modem interface.

This is not currently documented, but a peripheral_hub should provide
the following methods:

 - isPresentRemote
 - getTypeRemote
 - hasTypeRemote
 - getMethodsRemote
 - callRemote

Closes #902. CC @SirEdvin - does this look usable for your needs?


SirEdvin:
Holy moly, this is happening.

👍

PR #1194, opened 2022-10-29
Possible fix for issue #1150

### LINKED ISSUE #1150: Shell Shouldn't Autocomplete Hidden Directories
I feel like Shell shouldn't autocomplete hidden directories unless the user first types a dot. Or perhaps the user must type the first letter after the dot for the autocompletion to suggest the hidden directory.

image

IvoLeal72:
Possible fix for #1150, the hidden file/folder will still show up if the user writes the dot

### DIFF
--- a/src/main/resources/data/computercraft/lua/bios.lua
+++ b/src/main/resources/data/computercraft/lua/bios.lua
@@ -763,19 +763,19 @@ function fs.complete(sPath, sLocation, bIncludeFiles, bIncludeDirs)
         local tFiles = fs.list(sDir)
         for n = 1, #tFiles do
             local sFile = tFiles[n]
-            if #sFile >= #sName and string.sub(sFile, 1, #sName) == sName then
+            if #sFile >= #sName and string.sub(sFile, 1, #sName) == sName and (
+                bIncludeHidden or sFile:sub(1, 1) ~= "." or sName:sub(1, 1) == "."
+            ) then
                 local bIsDir = fs.isDir(fs.combine(sDir, sFile))
                 local sResult = string.sub(sFile, #sName + 1)
-                if bIncludeHidden or string.sub(sFile, 1, 1) ~= "." or string.sub(sResult, 1, 1) ~= "." or sResult ~= sFile then
-                    if bIsDir then
-                        table.insert(tResults, sResult .. "/")
-                        if bIncludeDirs and #sResult > 0 then
-                            table.insert(tResults, sResult)
-                        end
-                    else
-                        if bIncludeFiles and #sResult > 0 then
-                            table.insert(tResults, sResult)
-                        end
+                if bIsDir then
+                    table.insert(tResults, sResult .. "/")
+                    if bIncludeDirs and #sResult > 0 then
+                        table.insert(tResults, sResult)
+                    end
+                else
+                    if bIncludeFiles and #sResult > 0 then
+                        table.insert(tResults, sResult)
                     end
                 end
             end
@@ -912,7 +912,7 @@ settings.define("paint.default_extension", {
 
 settings.define("list.show_hidden", {
     default = false,
-    description = [[Show hidden files (those starting with "." in the Lua REPL)]],
+    description = [[Show hidden files (those starting with "." in the Lua REPL).]],
     type = "boolean",
 })
 
@@ -949,7 +949,7 @@ settings.define("bios.strict_globals", {
 })
 settings.define("shell.autocomplete_hidden", {
     default = false,
-    description = [[Autocomplete hidden files and folders (those starting with ".")]],
+    description = [[Autocomplete hidden files and folders (those starting with ".").]],
     type = "boolean",
 })
 
--- a/src/main/resources/data/computercraft/lua/bios.lua
+++ b/src/main/resources/data/computercraft/lua/bios.lua
@@ -713,12 +713,17 @@ end
 
 -- Install the lua part of the FS api
 local tEmpty = {}
-function fs.complete(sPath, sLocation, bIncludeFiles, bIncludeDirs, bIncludeHidden)
+function fs.complete(sPath, sLocation, bIncludeFiles, bIncludeDirs)
     expect(1, sPath, "string")
     expect(2, sLocation, "string")
-    expect(3, bIncludeFiles, "boolean", "nil")
-    expect(4, bIncludeDirs, "boolean", "nil")
-    expect(5, bIncludeHidden, "boolean", "nil")
+    if type(bIncludeFiles) == "table" then
+        bIncludeDirs = bIncludeFiles.include_dirs
+        bIncludeHidden = bIncludeFiles.include_hidden
+        bIncludeFiles = bIncludeFiles.include_files
+    else
+        expect(3, bIncludeFiles, "boolean", "nil")
+        expect(4, bIncludeDirs, "boolean", "nil")
+    end
 
     bIncludeHidden = bIncludeHidden ~= false
     bIncludeFiles = bIncludeFiles ~= false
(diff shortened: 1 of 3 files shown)


Wojbie:
Should this interaction be dependent on list program setting? I feel it should either be always on or have it's own setting. Using list. setting feels like missuse.

SquidDev:
Thank you for looking into this!

I think I agree with Wojbie here that we probably want a different setting here. Maybe worth adding an `fs.autocomplete_hidden` setting [here](https://github.com/cc-tweaked/CC-Tweaked/blob/mc-1.16.x/src/main/resources/data/computercraft/lua/bios.lua#L939) (not sold on that name, so feel free to pick something else!).

IvoLeal72:
Changed it to use fs.autocomplete_hidden setting, it really does make more sense

SquidDev (approved):
Looking at this again, I wonder if this should be an argument to the function instead, and the shell passes in the setting instead.

I don't know - think I'm inclined to leave as is for now, but would appreciate other people's thoughts!

IvoLeal72:
I thought about using an argument, should make more sense for programs using the fs.complete(), but due to my lack of knowledge on how to make that work, I have gone with this option. But with some help, I believe I could make those changes

Wojbie:
As a counterpoint what if we don't use a setting and only complete hidden files if user input starts with "." ? Then it will only complete if someone explicitly wants hidden one?

SquidDev:
> As a counterpoint what if we don't use a setting and only complete hidden files if user input starts with "." ?

That's the behaviour that Ivo has implemented (ignoring the setting). It's tempting though I do like going for the most conservative choice first.

Not sure, do you have any opinions @IvoLeal72?

Wojbie:
I might be mistreating the code but would this code not interfere with autocompleting a dot anywhere in filenames? Say with user input of `example` then file `example.txt` would not be autocompleted cause `.txt` starts with dot? Sorry for not testing code but can't do that on phone :(

IvoLeal72:
In my view the best option is the parameter one, and the setting is passed by the shell, and should then be shell.autocomplete_hidden or something in that spirit, and with that in mind, the default should be true if we want to keep the behavior, and leave it to the user to change it, or false to create this new behaviour

We can make that shell option false for the new behaviour, but the argument is optional with a default of true, so any programs that use fs.complete keep the behaviour

IvoLeal72:
Now that you mention it @Wojbie, yes that is a bug, I will think of a way to fix that, need to give another look to the original fs.complete function

SquidDev:
Ahh sorry, missed your first comment before Wojbie's. I should know better than to check GitHub on mobile!

> We can make that shell option false for the new behaviour, but the argument is optional with a default of true, so any programs that use fs.complete keep the behaviour

Yeah, agreed. I wonder if we switch to using table-style options for handling this, [like we do for `http.request`](https://github.com/cc-tweaked/CC-Tweaked/blob/14cb97cba106a4981f7990c792ad5f8b7405172b/src/main/resources/data/computercraft/lua/bios.lua#L605-L640). So preserve the previous signature, but if the third argument `bIncludeFiles` is a table, extract the arguments from there instead, allowing for something like:

```lua
fs.complete(path, location, {
  include_files = true,
  include_dirs = false,
  include_hidden = true,
})
```


Sorry, realise I'm messing you around here a bit!

IvoLeal72:
So the include_hidden argument should only be available in the table format, or should it be added as an in-line argument too?

IvoLeal72:
Other question is if fs.complete() is set to not show the hidden_files, should they appear, after the user inputed the initial dot ?

SquidDev:
> So the include_hidden argument should only be available in the table format

I think just table. I have a general rule that functions shouldn't take more than 4 arguments. As most people are writing CC without any IDE tooling, it gets pretty hard to keep track of what's what.

> Other question is if fs.complete() is set to not show the hidden_files, should they appear, after the user inputed the initial dot ?

I think so, yes!

Wojbie:
I see what you did here to fix the issue i mentioned but i think it still will happen if auto-completed file is hidden one?
Like if user input of `.example` then file `.example.txt` would not be autocompleted cause `.txt` starts with dot and file name starts with dot too?

I think this is getting odd. I think all logic going on here might need going back to design board and we should specify what exactly is end goal logic we are going on for here? Cause i feel a bit lost now.

IvoLeal72:
So, the objective is to be able to make the shell autcompletion to ignore hidden files, to make them more hidden.
Points that I believe are set are:

- After the user inputs a dot, the autcompletion should work again
- This should be done through an argument of fs.complete
- The arguments should be passed by table so to not exceed @SquidDev 's 4 arguments rule
- The arguments hould have a default of true, to not change behaviour of user-made scripts, but the shell should start using this new behaviour
- I need to think better on the handling because of the bug @Wojbie pointed.

IvoLeal72:
I believe I have a version that fits all the mentioned points, I will push it for evaluation

src/main/resources/data/computercraft/lua/bios.lua
```diff
@@ -758,14 +766,16 @@ function fs.complete(sPath, sLocation, bIncludeFiles, bIncludeDirs)
...
                 local sResult = string.sub(sFile, #sName + 1)
-                if bIsDir then
-                    table.insert(tResults, sResult .. "/")
-                    if bIncludeDirs and #sResult > 0 then
-                        table.insert(tResults, sResult)
-                    end
-                else
-                    if bIncludeFiles and #sResult > 0 then
-                        table.insert(tResults, sResult)
+                if bIncludeHidden or string.sub(sFile, 1, 1) ~= "." or string.sub(sResult, 1, 1) ~= "." or sResult ~= sFile then
```
SquidDev: Just going through this logic, just wondering what the `sResult ~= sFile` is for? I couldn't find a case where it affected things.

An alternative way of doing this would be to replace the if statement on line 766 with the following:

```lua
if #sFile >= #sName and string.sub(sFile, 1, #sName) == sName and (
    bIncludeHidden or sFile:sub(1, 1) ~= "." or sName:sub(1, 1) == "."
) then
```
It appears to be have the same behaviour, but you've definitely looked at this more carefully than me, so will defer to you!

src/main/resources/data/computercraft/lua/bios.lua
```diff
@@ -937,6 +947,11 @@ settings.define("bios.strict_globals", {
     description = "Prevents assigning variables into a program's environment. Make sure you use the local keyword or assign to _G explicitly.",
     type = "boolean",
 })
+settings.define("shell.autocomplete_hidden", {
+    default = false,
+    description = [[Autocomplete hidden files and folders (those starting with ".")]],
```
SquidDev:
```suggestion
    description = [[Autocomplete hidden files and folders (those starting with ".").]],
```

Sorry, tiny nitpick!

SquidDev (changes requested):
I've left a couple of comments, but otherwise this looks fantastic! Thanks for your patience with this :)

src/main/resources/data/computercraft/lua/bios.lua
IvoLeal72: In my defense, the list.show_hidden settings doesn't have a finishing dot, I copied from that ;P

But sure, I can change it ;)

src/main/resources/data/computercraft/lua/bios.lua
IvoLeal72: I added the comparison for the hidden file with extension case @Wojbie mentioned.
I have to check if that works for completing hidden files after inserting the dot, since sName and sResult are different, no?

src/main/resources/data/computercraft/lua/bios.lua
SquidDev: Oh, does it not? Can you fix that one as well then please? That's probably my fault :flushed:, I'm always forgetting to end sentences.

IvoLeal72: Haha, I can check that and fix it in all the settings descriptions when I'm able to have another look at the code :smile:

src/main/resources/data/computercraft/lua/bios.lua
IvoLeal72: I looked at it, your change makes sense, pushing it now

src/main/resources/data/computercraft/lua/bios.lua
IvoLeal72: Done, pushing it now, only list.show_hidden missed the .

SquidDev (approved):

PR #1195, opened 2022-10-30
Don't unnecessarily scroll in edit run wrapper

### LINKED ISSUE #1192: file reading behaving weird when running in edit
### Minecraft Version

1.18.x

### Version

1.100.10

## This bug only appears when the code is run in ingame ide
The following code should read the file and print the content:
```
file = fs.open("myFile","r")
print(file.readAll())
file.close()
```
but it **does not**

To make it print the content of the file i need to add any print statement anywhere before the printing of the file content.
Like this:
```
file = fs.open("myFile","r")
print("hello world")
print(file.readAll())
file.close()
```

The first print will actually not be visible on the screen.

SquidDev:
Fixes #1192. CC @Wojbie, does this look okay to you?


Wojbie:
Looks way more sensible then what I wrote :)

PR #1199, opened 2022-11-02
Check for valid nSender field on RedNet message receive

Ale32bit:


Ale32bit:
VS code git integration...

PR #1200, opened 2022-11-02
Check for valid nSender field on RedNet message receive

Ale32bit:


SquidDev:
Can you move the `and message.nMessageID == message.nMessageID` check to be above this, just so its next to the `type(message.nMessageID)` check?

SquidDev:
Maybe we need a whole schema validation library for this :D:.

Wojbie:
This check might need to be more complicated. After all it can be nil for backwards compatibility too. So it should allow both nil and number but not NaN right?

SquidDev:
Oh yeah, good point! The rednet protocol is an internal detail so we technically _can_ change it, but given we allow `nSender` to be nil elsewhere, should probably keep that here too.

SquidDev:
Thank you!

SquidDev:
Just to clarify that test failure, `and` binds more tightly than `or`, so it was actually parsed as

```lua
if (
    type(message) == "table" and type(message.nMessageID) == "number"
    and message.nMessageID == message.nMessageID and not received_messages[message.nMessageID]
    and type(message.nSender) == "nil"
) or (
    (type(message.nSender) == "number" and message.nSender == message.nSender))
    and ((message.nRecipient and message.nRecipient == os.getComputerID()) or channel == CHANNEL_BROADCAST)
    and isOpen(modem)
) then
```

I thought I had an illuaminate check for this, but evidently not!

PR #1203, opened 2022-11-03
Reformat and start using Java 9+ language features

SquidDev:
Going to leave this up for a few days in case I change my mind and decide this is a terrible idea.

I think there's a lot of benefits to this (code style is more consistent with the rest of the Java ecosystem, makes everything a little more concise), but it is also a pretty major change, so I don't want to rush it.

This commit does bring us down below 100k LOC again, which I think is pretty funny. Don't worry, I've plans to change that :D:.


Lemmmy:
IMO better late than never! I don't know what it's like for other people (especially those with less Java experience or without proper IDE formatters set up), but I can certainly say that every time I've tried to contribute, even being well aware of the style guidelines, I have spent an embarrassing amount of time wrestling against checkstyle. Hopefully this will make contributing easier

PR #1205, opened 2022-11-04
Move dan200.computercraft.core into a separate module

SquidDev:
In which I break Git history. Unfortunately, I don't think there's a good way to avoid this :/.

## Motivation
It should be possible to consume the ComputerCraft's core (i.e. non-Minecraft code) in other projects, such as emulators.  While this has been possible for years, it's somewhat tricky from a maintenance perspective - it's very easy to accidentally add an MC dependency somewhere!

By publishing a separate "core" jar, we can better distinguish the boundaries between our Lua runtime and the Minecraft-specific code.

The end goal of this is we publish four API jars:

 - `core-api`: The API in this project.
 - `common-api`: The common API for Forge and Fabric. This should only be consumed by multiloader mods.
 - `forge-api`/`fabric-api`: Loader specific APIs. These include `common-api` and _possibly_ `core-api`: I have a feeling ForgeGradle doesn't handle transitive dependencies with `fg.deobf`, so it may be easier just to merge them.

## Alternative options:
 - Have one core project (rather than separate `core` and `core-api` modules), and publish a separate `:api` jar, like we currently do for the main mod. I like the separation this gives us, but does double the number of sub-projects one ends up having.
 - Possibly something else?

Don't know if people have any thoughts on this?


Lemmmy:
From an API consumer point of view: provided it is convenient on your end with the custom build tooling etc. I think the four jar solution is a good one, if it can be relatively frictionless for the consumer, and a simple "how to import" readme paragraph goes a long way for alleviating any potential confusion there. 

Also right now it's too easy to end up bringing in the whole CC jar and using internal methods, so having users bring in a higher-level dependency by default, until they need(?) to drill into more specific loader functionality with a fabric/forge jar, should hopefully lead to better code in peripheral mods, and better contributions to the API. Maybe wishful thinking but that's just my bed rambling

As a general opinion of the whole idea, not to take away from emulator focus but a goal to move towards multi-loader support on CC's side may also lead to eventual multi-loader support of consuming peripheral mods, so I think this is a fantastic effort!!! Exciting things ahead

SquidDev:
Thanks, that's very helpful! 

> Also right now it's too easy to end up bringing in the whole CC jar and using internal methods 

We do ship a separate API jar, though only started doing so this year, so suspect many mods aren't using it. Though of course Plethora needs access to all those nasty internals anyway :p.

SirEdvin:
This should be really nice, especially for me, since I hope to port my mods into Forge one day. And more interconnection between CCR and CCT

PR #1208, opened 2022-11-06
Move our public API into separate modules

SquidDev:
This adds two new modules: common-api and forge-api, which contain the common and Forge-specific interfaces for CC's Minecraft-specific API.

We add a new PlatformHelper interface, which abstracts over some of the loader-specific functionality, such as reading registries[^1] or calling Forge-specific methods. This interface is then implemented in the main mod, and loaded via ServiceLoaders.

## Implementation notes

 - We now split shared and client-specific source code into separate modules. This is to make it harder to reference client code on the server, thus crashing the game.

   Eventually we'll split the main mod up too into separate source sets - this is, of course, a much bigger problem!

 - There's currently some nastiness here due to wanting to preserve binary compatibility of the API. We'll hopefully be able to remove this when 1.19.3 releases.

 - In order to build a separate Forge-specific API jar, we compile the common sources twice: once for the common jar and once for the Forge jar.

   Getting this to play nicely with IDEs is a little tricky and so we provide a cct.inlineProject(...) helper to handle everything.

## API notes and questions
 - The plan here is to remove `IPeripheralProvider` from the common/Fabric API and make it Forge specific. For Fabric, peripherals will _just_ be provided via the block lookup API.

   This basically means the Fabric-side API will look something like:
   ```java
   // Register for a specific block (there's also a registerForBlocks(Block...) version)
   PeripheralLookup.get().registerForBlockEntity(
     (b, dir) -> ConfigState.enableCommandBlock ? new CommandBlockPeripheral(b) : null,
     BlockEntityType.COMMAND_BLOCK
   );
   // Register a fallback provider, much closer to the current IPeripheralProvider interface.
   PeripheralLookup.get().registerFallback((world, pos, blockState, blockEntity, dir) -> idk);
   ```

   Does this work for people (cc @Lemmmy, @SirEdvin, not sure who else develops peripheral mods for Fabric)?

[^1]: We /can/ do this with vanilla's APIs, but it gives a lot of deprecation warnings. It just ends up being nicer to abstract over it.


SirEdvin:
You want to get rid of `IPeripheralProvider` because it returns `Lazy<IPeripheral>` in forge and just `IPeripheral` in fabric?

Lemmmy:
Sounds good to me. The only thing I'm using IPeripheralProvider for at the moment is manipulators. Looking closer I'm not sure if it even needs to (IPeripheralTile should work fine), but both of those interfaces would work fine for all the uses I can think of!

SirEdvin:
I use `IPeripheralProvider` for [this](https://github.com/SirEdvin/UnlimitedPeripheralWorks/blob/1.18/src/main/kotlin/site/siredvin/peripheralworks/computercraft/ComputerCraftProxy.kt#L87) kind of integrataions, but I guess, if you will have Fallback logic, that will be pretty much the same, this should be fine.

SquidDev:
> You want to get rid of `IPeripheralProvider` because it returns `Lazy<IPeripheral>` in forge and just `IPeripheral` in fabric?

In part, yeah. Ideally `IPeripheralProvider` would only be used on Forge for peripherals targeting non-BlockEntity blocks (can't use capabilities) and for fallback cases like UPW (which you don't want to register a cap). The block lookup API covers both those cases already, so there's not really any case when you need it.

PR #1211, opened 2022-11-10
Draw the rest of the owl

SquidDev:
This is the culmination of many months of on-and-off work trying to get CC:T to properly support multiple mod loaders. I was hoping that carefully arranging everything we'd give one final tug and it'd all work out, but this is actually a much chunkier commit than I expected.

I'm not expecting anyone to review this (1005 files changed, lol) - mostly leaving this up so I can have another look tomorrow and see if there's anything I missed.

This is woefully untested - I've not even checked the client works. But tests pass, so it must be fine right!


Lupus590:
> But tests pass, so it must be fine right!

Famous last words.

With a commit this big that changes so much of the deep workings, I would try to test everything.

SquidDev:
Hey, we're up to 40% test coverage! That's ... not anywhere near everything.

But yes, it was crashing on client load due to the shader location changing. That is now fixed.

SquidDev:
Though let this be a lesson in waiting for CI to pass before merging.

SirEdvin:
Oh, there is some kotlin code. 

Interesting, will this happen, that in some day you port all of CC to kotlin :)

SquidDev:
The test code has been in Kotlin for a while now, it's just slowly expanding to encompass more of the build system :). I think it's unlikely that I'll port the actual mod though - don't want to have to drag in another dependency!

Issue #1214, opened 2022-11-14
[MC-next] Breaking API changes


SquidDev:
This is mostly a todo list for breaking API changes scheduled for the next Minecraft version, be that 1.19.3 or 1.20:

## Removals
 - Remove Forge-specific methods from `ComputerCraftAPI`.
 - Remove `IComputerSystem.getFileSystem` (maybe we should mark it as deprecated now). This has always been a little broken, and I don't think it gets any use.

## Changes
 - Mount getters in `ComputerCraftAPI` (e.g. `createUniqueNumberedSaveDir`) should accept a `MinecraftServer`, not a `Level`. Possibly also true for the wireless network getter?

   1.19.3 also redoes registries a little, so maybe the detail registries should also accept a `ServerLevel`/`MinecraftServer`. Less sure there.

 - Rethink how `IPacketNetwork` and `IWiredNode` interact: the fact that `IWiredNode` extends `IPacketNetwork` has always been ugly, and I think there's probably better ways.
 - Redo `IComputerAccess.mount` entirely:

   This is a pre-requisite to #676: if we're going to move mounting logic to the Lua side (which I think we probably are) then we need the actual mounting operation to be asynchronous. As a result, the semantics of this API call need to change a bit:
   - The API call should always succeed: mounting will never "fail", it just may be ignored.
   - However, there's no guarantee that the mount will occur immediately (or at all). As such, we should return some `AttachedMount` interface, which contains a `getMountPath(): String?` method (as well as a `close()`/`unmount()` method).
   - The implementation should be responsible for handling conflicting paths (for instance, mounting disks at `/disk2`, etc... instead) - currently that's done by the caller.

     It's possible that some mods may want to guarantee something is mounted at a specific path (some mods used to mount additional programs when a peripheral was attached, though really that should be done with datapacks now), and not allow mounting it multiple times. Maybe we need an option for that?

 - Rethink `IMount`/`IWritableMount`: This is much less thought through right now, but some things I'd quite like to look into:
   - Can we make the API closer to Java's built-in `Filesystem`? Is there any merit in doing so?
   - Support moving files (maybe whole directories too?) within a mount (see #962).
   - Support opening files in r/w mode.
   - Make the current default methods non-default.

 - Rename `ForgeComputerCraftAPI` to `ComputerCraftAPIForge`.

Lemmmy:
[...]

PR #1226, opened 2022-11-24
Check the filesystem for isReadOnly

Lemmmy:
As mentioned in #1214. The IMount implementation is default to avoid breaking any existing implementations, and I have provided implementations for FileMount (actually checks the real filesystem file for canWrite), FileSystemWrapperMount (delegates) and SubMount (delegates). Everything else uses the default behaviour of returning `true` - I am not sure if returning `false` would be more sensible and just letting the operation silently fail or return an IOException (depending on the situation and operating system). As far as I understand for all existing mounts this will always get hit by the `writableMount` check anyway.

I haven't written tests for this yet as I am not sure it is sensible or possible to test due to filesystem differences, but as far as I have tested everything is working as intended on Windows and Linux.

To recap, the use case for this is for `edit` etc. to prevent saving if the filesystem/directory/file is marked by the OS as read-only. It relies on [File.canWrite](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#canWrite()).

### DIFF
--- a/projects/core-api/src/main/java/dan200/computercraft/api/filesystem/IMount.java
+++ b/projects/core-api/src/main/java/dan200/computercraft/api/filesystem/IMount.java
@@ -42,17 +42,6 @@ public interface IMount {
     boolean isDirectory(String path) throws IOException;
 
     /**
-     * Returns whether a file with a given path is read-only or not.
-     *
-     * @param path A file path in normalised format, relative to the mount location. ie: "programs/myprograms".
-     * @return If the file exists and is read-only.
-     * @throws IOException If an error occurs when checking whether the file is read-only.
-     */
-    default boolean isReadOnly(String path) throws IOException {
-        return true;
-    }
-
-    /**
      * Returns the file names of all the files in a directory.
      *
      * @param path     A file path in normalised format, relative to the mount location. ie: "programs/myprograms".
--- a/projects/core-api/src/main/java/dan200/computercraft/api/filesystem/IMount.java
+++ b/projects/core-api/src/main/java/dan200/computercraft/api/filesystem/IMount.java
@@ -42,6 +42,17 @@ public interface IMount {
     boolean isDirectory(String path) throws IOException;
 
     /**
+     * Returns whether a file with a given path is read-only or not.
+     *
+     * @param path A file path in normalised format, relative to the mount location. ie: "programs/myprograms".
+     * @return If the file exists and is read-only.
+     * @throws IOException If an error occurs when checking whether the file is read-only.
+     */
+    default boolean isReadOnly(String path) throws IOException {
+        return true;
+    }
+
+    /**
      * Returns the file names of all the files in a directory.
      *
      * @param path     A file path in normalised format, relative to the mount location. ie: "programs/myprograms".
--- a/projects/core-api/src/main/java/dan200/computercraft/api/filesystem/IWritableMount.java
+++ b/projects/core-api/src/main/java/dan200/computercraft/api/filesystem/IWritableMount.java
@@ -78,4 +78,15 @@ public interface IWritableMount extends IMount {
     default OptionalLong getCapacity() {
         return OptionalLong.empty();
     }
+
+    /**
+     * Returns whether a file with a given path is read-only or not.
+     *
+     * @param path A file path in normalised format, relative to the mount location. ie: "programs/myprograms".
+     * @return If the file exists and is read-only.
+     * @throws IOException If an error occurs when checking whether the file is read-only.
+     */
+    default boolean isReadOnly(String path) throws IOException {
+        return true;
+    }
 }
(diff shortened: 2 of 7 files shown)


projects/core-api/src/main/java/dan200/computercraft/api/filesystem/IMount.java
```diff
@@ -41,6 +41,17 @@ public interface IMount {
...
+     * Returns whether a file with a given path is read-only or not.
+     *
+     * @param path A file path in normalised format, relative to the mount location. ie: "programs/myprograms".
+     * @return If the file exists and is read-only.
+     * @throws IOException If an error occurs when checking whether the file is read-only.
+     */
+    default boolean isReadOnly(String path) throws IOException {
+        return true;
+    }
+
```
SquidDev: I think it probably makes more sense to move this to `IWritableMount` and then default to returning `false`.

projects/core/src/main/java/dan200/computercraft/core/filesystem/FileMount.java
```diff
@@ -143,6 +143,14 @@ public boolean isDirectory(String path) {
         return file.exists() && file.isDirectory();
     }
 
+    @Override
+    public boolean isReadOnly(String path) throws IOException {
+        if (!created()) return path.isEmpty();
+
+        var file = getRealPath(path);
+        return file.exists() && !file.canWrite();
```
SquidDev: I'm not sure about this. Just to confirm my understanding, if we've not created the mount, then `/` is considered read-only? However, this should only be the case if the actual "computercraft/computers" directory is read only, which hopefully isn't the case!

I wonder how we should handle missing files in general here. Currently `fs.isReadOnly("read-only-dir/missing-file")` will return `false`, which isn't great. A couple of possibilities I can think of:

 - Keep the behaviour here: it's not a big issue after all!

 - `isReadOnly` walks up the tree until it finds a file which exists, then returns that. My big concern here is someone doing `fs.isReadOnly(("/p"):rep(10000))`, which could cause us to spend ages in filesystem code. 

   Hrmr, maybe you could do a binary search over the file path looking for one which exists :thinking:. Feel like that might be a bit too complicated.

SquidDev (changes requested):
Thank you so much for looking at this!

I've left one question about missing files below: this actually may be more complicated than I thought :/.

projects/core/src/main/java/dan200/computercraft/core/filesystem/FileMount.java
Lemmmy: Looks like that was a total oversight with the created check there. I've removed the check entirely and now I believe the behaviour is correct. I went with traversing the tree in the same way `makeDirectory` does. In practice, 10000 repetitions is near-instantaneous, and 100000 repetitions will throw "Too long without yielding". Since `makeDirectory` would also have the same issue, I don't think this is a major concern for now.

javaw_Y0HY8Bis14

projects/core-api/src/main/java/dan200/computercraft/api/filesystem/IWritableMount.java
```diff
@@ -78,4 +78,15 @@ public interface IWritableMount extends IMount {
...
+
+    /**
+     * Returns whether a file with a given path is read-only or not.
+     *
+     * @param path A file path in normalised format, relative to the mount location. ie: "programs/myprograms".
+     * @return If the file exists and is read-only.
+     * @throws IOException If an error occurs when checking whether the file is read-only.
+     */
+    default boolean isReadOnly(String path) throws IOException {
+        return true;
```
SquidDev:
```suggestion
        return false;
```

I think?

projects/core/src/main/java/dan200/computercraft/core/filesystem/FileMount.java
```diff
@@ -143,6 +143,21 @@ public boolean isDirectory(String path) {
...
+        if (file.exists()) {
+            return !file.canWrite();
+        }
+
+        var parent = file.getParentFile();
+        while (!parent.exists() && parent != rootPath) {
+            parent = parent.getParentFile();
+        }
+
+        return !parent.canWrite();
```
SquidDev: I wonder if it's better to write this something like:

```java
while (true) {
  if (file.exists()) return !file.canWrite();
  if (file.equals(rootPath)) return false;
  file = file.getParentFile();
}
```

I think this is the same, except it means `isReadOnly` will return `false` when the root doesn't exist?

Actually, I wonder if it's worth adding some tests for these? Happy to do so myself if you'd rather not!

SquidDev (changes requested):

projects/core-api/src/main/java/dan200/computercraft/api/filesystem/IWritableMount.java
Lemmmy: Whoops, sorry, completely missed that one!

projects/core/src/main/java/dan200/computercraft/core/filesystem/FileMount.java
Lemmmy: image

Seems to work! Thanks, that's much more concise.

Lemmmy:
Tests would definitely be a good idea but I'm not sure if there's a way to comprehensively and reliably do so in a way that would work cross-platform; Windows for example doesn't actually have attribute-based read-only directories, unless you have some kind of mock filesystem set up or made certain parts of the test only run on certain platforms. I could be wrong though

SquidDev:
:roll_eyes: Maybe I should start paying money for an external CI service. Local tests are definitely not this unreliable!

PR #1231, opened 2022-11-25
Time fs and peripheral operations

SquidDev:
I'd quite like to write this in a way that we don't actually allocate the intermediate `OperationTimer` instance. It's not clear to me if this is actually the case right now: I need to run jitwatch on this.


SquidDev:
It looks like it ends up heap allocated, as it'd end up using too much stack otherwise. It's a shame, but I don't think it's a major blocker: I don't think one extra allocation will be the slow bit in the metrics pipeline :(.

PR #1255, opened 2022-12-17
Deliver x86_64 Illuaminate binaries to all Macs

toad-dev:
On ARM-based Macs Gradle will throw the following error while configuring the project or trying to run the Lua linting tasks:

```
Could not create task ':forge:lintLua'.
Could not create task of type 'IlluaminateExec'.
Failed to query the value of extension 'illuaminate' property 'file'.
Unsupported architecture 'aarch64' for illuaminate
```
This PR adds a special case to the Illuaminate dependency extension logic so that all Macs receive the x86_64 binary. ARM-based Macs will run this binary through Rosetta no problem.


SquidDev:
Amazing, thank you!

PR #1262, opened 2022-12-20
Fix duplicated swing animations on high-ping servers

emmachase:
Use `InteractionResult.sidedSuccess` / `.CONSUME` where applicable instead of `.SUCCESS`. This prevents the server from sending an additional swing animation packet to the client. Normally this isn't a problem, since the client will de-duplicate swing packets if they are within the animation duration of the currently playing swing; however, when connected to a server with a high ping the packet is sent after the animation is already finished on the client, resulting in a duplicate animation.

To be clear, using `InteractionResult.sidedSuccess` causes a result of `SUCCESS` to be used on the client, and `CONSUME` on the server. This has the effect that the client will immediately play the swing animation when the interaction happens since it is a SUCCESS, and the server will treat the interaction as successful, but prevent sending any swing animation to players. For players other than the actor, the client sends a C2S packet for the player swing to the server which gets forwarded to everyone else nearby. This practice appears to be standard in Minecraft's vanilla source.

https://user-images.githubusercontent.com/65421017/208580800-7306516b-8fac-4c5b-95fb-e58504918177.mp4

### DIFF
--- a/projects/common/src/main/java/dan200/computercraft/shared/computer/blocks/AbstractComputerBlockEntity.java
+++ b/projects/common/src/main/java/dan200/computercraft/shared/computer/blocks/AbstractComputerBlockEntity.java
@@ -98,7 +98,7 @@ public abstract class AbstractComputerBlockEntity extends BlockEntity implements
                 setLabel(currentItem.getHoverName().getString());
                 currentItem.shrink(1);
             }
-            return InteractionResult.SUCCESS;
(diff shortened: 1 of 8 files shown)


SquidDev:
Amazing. Thank you for digging into this!

SquidDev:
I really need to make the OSX CI job less flakey :(.

Issue #1256, opened 2022-12-18
cc.shell.complete.file/dir/fileOrDir should quote paths containg a space in them


walksanatora:
cc.shell.complete should quote file paths containg mutiple lines to prevent them from being counted as seperate arguments

SquidDev:
Yeah, this has been a pain for a while. The main problem is that `read`'s auto-completion system doesn't support replacing the preceding text (i.e. you need to replace `hel` with `"hel` as well as auto-completing `lo world"`), and it's not clear how to do that in a backwards-compatible fashion.

KlutzLament:
Can you use backslashes to escape the spaces like linux shells do?

walksanatora:
currently not, but that would be a valid option i think since that doesen't have you editing the allready written part of the string

SquidDev:
Yeah, that seems a reasonable solution!

LeVoid:
I'm working on a solution right now. Escaping the spaces with backslashes seems to pose an issue for me though.

So I'm working on an alternative, which might include an additional optional parameter to "shell.complete()".
Going to mention this issue in the commit.

Edit: Just noticed that I have to be a contributor in order to open a new branch. How would I become one?

walksanatora:
You are supposed to fork the project
Then make changes there
Then open a Pull Request

SquidDev:
It occurs to me that using backwards slashes for escapes may be confusing for those used to Windows paths (for instance `foo\bar\baz.lua`). I don't know if any programs actually rely on that, but we should do a search over Pastebin and GitHub.

Lemmmy:
A cursory regex over the pastebin archive shows that it has been used at least a few times, possibly unsuccessfully:

```
rg 'fs\.(open|move|delete)\(\s*["\u0022].*\\\\'
A8kCA8Zc.txt
729:            local myFile = fs.open("\\cards\\"..fileList[i],"r")

kCTERg1Z.txt
919:            local myFile = fs.open("\\disk\\id","w")

CQUAprWH.txt
9:      local output = fs.open("disk\\inventory.txt","w")
37:             local input = fs.open("disk\\inventory.txt","r")

dBHuAvka.txt
45:             fs.delete("Databases\\" .. dbName)

PyqB5cWf.txt
14:     file = fs.open("\\update","w")
112:local file = fs.open("\\disk\\id","r")
119:file = fs.open("\\startup","w")

rYsd9J8B.txt
[...]

PR #1267, opened 2022-12-24
#1256: "shell.complete should quote filenames with spaces in them"

LeVoid:
Regarding issue #1256 

Added an optional parameter to Shell.complete(sLine, bFull):
-> if bFull is set to true, the function will return the full filenames of all files that start with sLine. It also quotes filenames if they have a space in them. Not supplying bFull has the same result as setting it to false.

This way you can choose which option to use and aren't tied to either one.

### DIFF
--- a/projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
+++ b/projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
@@ -410,7 +410,7 @@ end
 -- @see shell.setCompletionFunction
 -- @see shell.getCompletionInfo
 -- @since 1.74
-function shell.complete(sLine)
+function shell.complete(sLine, bFull)
     expect(1, sLine, "string")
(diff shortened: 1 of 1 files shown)


Wojbie:
How would this interact with current shell autocompleting logic? Don't it expect to be given suffix?

LeVoid:
> How would this interact with current shell autocompleting logic? Don't it expect to be given suffix?

Because the current logic only supplies the first argument (sLine), bFull is set to nil. So it won't return the full names.

Due to this, nothing about the current logic changes. The only time it changes the returned list is when you supply the bFull-argument manually.

SquidDev:
The trouble with that is I'm not sure when this is actually useful. This isn't compatible with `read`, so I'm not sure when you'd be able to use this?

walksanatora:
i still believe that the idea discussed in the mentioned issue is better
(support backslash escapes in shell and autocomplete)

Issue #1245, opened 2022-12-09
Vague thoughts on improving CraftOS UX


SquidDev:
This isn't really a concrete feature request, but is more some musings on things which would be good to improve to make programming in ComputerCraft a little bit friendlier. I suspect some of what's written here is worth adding to CC, some should be in an external program, and some is an utterly terrible idea!

Definitely would like feedback here, both on sorting out my ideas, and suggesting any other frustrations and UX improvements we could be doing.

## Scrolling in the shell
This has been discussed before (#792), where I originally closed it due to concerns about implementation complexity. However, given _how_ useful scrollback is, I do think it's worth revisiting this.

Obviously [mbs](https://github.com/squiddev-cc/mbs) has had this for years, and while I think there's some bits worth mimicking (the handling it handles full screen programs is very sensible) it's probably worth avoiding copying wholesale. We definitely need a better way of handling errors that occur in full-screen programs (as those get entirely hidden right now).

What's less clear to me is when the shell should scroll, as in when should page up/down and the mouse wheel move the terminal up/down, rather than being forwarded to user programs? [mbs intercepts these events][mbs read] when the cursor is blinking (so `read` is being called) and not in a full screen program (e.g. paint or edit). This works fine, but maybe it's is too hacky for CC itself?

## Full filenames in debug information
When CC:T loads a Lua file, it uses [the file path as the chunk name](https://github.com/cc-tweaked/CC-Tweaked/blob/367773e173b2a135f213c19bb3df949ea43b9dfa/projects/core/src/main/resources/data/computercraft/lua/bios.lua#L470), rather than the whole path. This is a bit of a pain for debugging utilities, as you can't map functions back to their original file. [We end up doing some very nasty things in our code coverage tooling][coverage] to work around that.

We've looked at changing this before (#465): the main problem is that for long file names (especially APIs from the ROM) take up far too much space on smaller screens, which makes errors pretty unreadable.

I wonder if the solution here is to patch Cobalt itself: we could make a file's shortsrc just use the last path segment, [rather than the last 60 characters](https://github.com/SquidDev/Cobalt/blob/6e39cba87e59e5672f6a3e992b077538a4df1566/src/main/java/org/squiddev/cobalt/compiler/LoadState.java#L199-L207).

## Tracebacks
There is an existing issue (#231) for this, but I think it's also worth mentioning here. There's been a couple of concerns I have which is why this issue hasn't had much movement:

[...]

PR #1273, opened 2022-12-28
Add support for shebangs, closes #551

### LINKED ISSUE #551: Add hashbangs
Or in this case...hyphen bangs. A special optional line at the top of a text file to tell CraftOS to open a file with a specific program. For instance, `--!/bin/py` at the top of a [Python script](https://github.com/neumond/python-computer-craft), or `--!paint` to load a paint file directly into the paint program when launched. :)

emmachase:
This adds support for shebangs when loading files with loadfile, this is how the shell loads programs as well.

One thing to note is that since this is being done in loadfile which is in the bios rather than the shell, it will not respect the PATH, but this is consistent with how shebangs work in *nix as well.

I also am not sure how I feel about copying the `tokenise` from the shell to parse quoted strings in the argument.

### DIFF
--- a/projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
+++ b/projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
@@ -54,6 +54,71 @@ else
     bgColour = colours.black
 end
 
+local function tokenise(...)
+    local sLine = table.concat({ ... }, " ")
+    local tWords = {}
+    local bQuoted = false
+    for match in string.gmatch(sLine .. "\"", "(.-)\"") do
+        if bQuoted then
+            table.insert(tWords, match)
+        else
+            for m in string.gmatch(match, "[^ \t]+") do
+                table.insert(tWords, m)
+            end
+        end
+        bQuoted = not bQuoted
+    end
+    return tWords
+end
+
+-- Execute a program using os.run, unless a shebang is present.
+-- In that case, execute the program using the interpreter specified in the hashbang.
+-- This may occur recursively, up to the maximum number of times specified by remainingRecursion
+-- Returns the same type as os.run, which is a boolean indicating whether the program exited successfully.
+local function executeProgram(remainingRecursion, path, args)
+    local file, err = fs.open(path, "r")
+    if not file then
+        printError(err)
+        return false
+    end
+
+    -- First check if the file begins with a #!
+    local contents = file.readLine()
+    file.close()
+
+    if contents:sub(1, 2) == "#!" then
+        remainingRecursion = remainingRecursion - 1
+        if remainingRecursion == 0 then
+            printError("Hashbang recursion depth limit reached when loading file: " .. path)
+            return false
+        end
+
+        -- Load the specified hashbang program instead
+        local hashbangArgs = tokenise(contents:sub(3))
+        local originalHashbangPath = table.remove(hashbangArgs, 1)
+        local resolvedHashbangProgram = shell.resolveProgram(originalHashbangPath)
+        if not resolvedHashbangProgram then
+            printError("Hashbang program not found: " .. originalHashbangPath)
+            return false
+        end
+
+        -- Add the path and any arguments to the interpreter's arguments
+        table.insert(hashbangArgs, path)
+        for _, v in ipairs(args) do
+            table.insert(hashbangArgs, v)
+        end
+
+        hashbangArgs[0] = originalHashbangPath
+        return executeProgram(remainingRecursion, resolvedHashbangProgram, hashbangArgs)
+    end
+
+    local dir = fs.getDir(path)
+    local env = createShellEnv(dir)
+    env.arg = args
+
+    return os.run(env, path, table.unpack(args))
+end
+
 --- Run a program with the supplied arguments.
 --
 -- Unlike @{shell.run}, each argument is passed to the program verbatim. While
@@ -84,10 +149,7 @@ function shell.execute(command, ...)
             multishell.setTitle(multishell.getCurrent(), sTitle)
         end
 
-        local sDir = fs.getDir(sPath)
-        local env = createShellEnv(sDir)
-        env.arg = { [0] = command, ... }
-        local result = os.run(env, sPath, ...)
+        local result = executeProgram(100, sPath, { [0] = command, ... })
 
         tProgramStack[#tProgramStack] = nil
         if multishell then
@@ -106,23 +168,6 @@ function shell.execute(command, ...)
         printError("No such program")
         return false
     end
-end
-
-local function tokenise(...)
-    local sLine = table.concat({ ... }, " ")
-    local tWords = {}
-    local bQuoted = false
-    for match in string.gmatch(sLine .. "\"", "(.-)\"") do
-        if bQuoted then
-            table.insert(tWords, match)
-        else
-            for m in string.gmatch(match, "[^ \t]+") do
-                table.insert(tWords, m)
-            end
-        end
-        bQuoted = not bQuoted
-    end
-    return tWords
 end
 
 -- Install shell API
--- a/projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
+++ b/projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
@@ -1,14 +1,39 @@
---- The shell API provides access to CraftOS's command line interface.
---
--- It allows you to @{run|start programs}, @{setCompletionFunction|add
--- completion for a program}, and much more.
---
--- @{shell} is not a "true" API. Instead, it is a standard program, which injects its
--- API into the programs that it launches. This allows for multiple shells to
--- run at the same time, but means that the API is not available in the global
--- environment, and so is unavailable to other @{os.loadAPI|APIs}.
---
--- @module[module] shell
+--[[- The shell API provides access to CraftOS's command line interface.
+
+It allows you to @{run|start programs}, @{setCompletionFunction|add completion
+for a program}, and much more.
+
+@{shell} is not a "true" API. Instead, it is a standard program, which injects
+its API into the programs that it launches. This allows for multiple shells to
+run at the same time, but means that the API is not available in the global
+environment, and so is unavailable to other @{os.loadAPI|APIs}.
+
+## Programs and the program path
+When you run a command with the shell, either from the prompt or
+@{shell.run|from Lua code}, the shell API performs several steps to work out
+which program to run:
+
+ 1. Firstly, the shell attempts to resolve @{shell.aliases|aliases}. This allows
+    us to use multiple names for a single command. For example, the `list`
+    program has two aliases: `ls` and `dir`. When you write `ls /rom`, that's
+    expanded to `list /rom`.
+
+ 2. Next, the shell attempts to find where the program actually is. For this, it
+    uses the @{shell.path|program path}. This is a colon separated list of
+    directories, each of which is checked to see if it contains the program.
+
+    `list` or `list.lua` doesn't exist in `.` (the current directory), so she
+    shell now looks in `/rom/programs`, where `list.lua` can be found!
+
+ 3. Finally, the shell reads the file and checks if the file starts with a
+    `#!`. This is a [hashbang][], which says that this file shouldn't be treated
+    as Lua, but instead passed to _another_ program, the name of which should
+    follow the `#!`.
+
+[hashbang]: https://en.wikipedia.org/wiki/Shebang_(Unix)
+
+@module[module] shell
+]]
 
 local make_package = dofile("rom/modules/main/cc/require.lua").make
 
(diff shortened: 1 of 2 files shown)


SquidDev:
Thank you for looking into this!

I think `loadfile` is probably the wrong place to add this logic. `loadfile` is part of the standard Lua library, and not something we really want to be adding Extra Stuff to.

I think a better location would be `os.run` or just in the shell itself. I've a preference for the latter, as that means we're not duplicating shell code elsewhere.

> it will not respect the PATH 

Given CC doesn't have a `/usr/bin/env` equivalent, I think we probably want to obey the path.

emmachase:
Yeah, I had similar thoughts. I put it in loadfile originally since that was the place where the file was actually being read, and I wanted to avoid reading the file twice. But I suppose it's probably not a big deal.

emmachase:
I've moved the logic to the shell in this most recent commit. It also uses `resolveProgram` now since we have access to it, so the path will be evaluated (and the `.lua` is optional). Let me know if you have any other comments or suggestions :)

Wojbie:
Do we need this shell setting? `shell.loadfile_max_recursion` It don't feel like something end users are gonna want to change? If we don't then i would suggest removing it and just leaving 10 there? If we do then it really should be defined like all other build in settings inside bios.lua file.

emmachase:
Thank you for pointing that out, I wasn't aware of the defines in bios.lua.
I do think it is a reasonable setting to have, I don't like the idea of putting an arbitrary limit on the max recursion depth without a way to change it; sounds like it would be really frustrating if you ever ran into it. Not that I expect anyone would, but I don't think having the option hurts anything.

projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
```diff
@@ -54,6 +54,68 @@ else
...
+-- Returns the same type as os.run, which is a boolean indicating whether the program exited successfully.
+local function executeProgram(nRemainingRecursion, env, sPath, ...)
+    local file = fs.open(sPath, "r")
+    if not file then
+        printError("File not found: " .. sPath)
+        return false
+    end
+
+    -- First check if the file begins with a #!
+    local sContents = file.readLine()
```
SquidDev: I cannot wait until Lua 5.4 when we have `<close>`.

```suggestion
    local sContents = file.readLine()
    file.close()
```

Also, would you be able to drop the Hungrarian notation in this function? Slowly trying to transition the codebase to not use it, it's just a slow process :).

projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
```diff
@@ -54,6 +54,68 @@ else
...
+        nRemainingRecursion = nRemainingRecursion - 1
+        if nRemainingRecursion == 0 then
+            printError("Interpreter recursion depth limit reached when loading file: " .. sPath)
+            return false
+        end
+
+        -- Load the specified interpreter instead
+        local tParts = tokenise(sContents:sub(3))
+        if #tParts > 2 then
+            printError("Interpreter may only take 1 argument, but " .. #tParts - 1 .. " were specified: " .. table.concat(tParts, " "))
```
SquidDev: How come? I know this is a POSIX (or at least OSX?, can never remember) limitation, but I assume we don't need to constrain ourselves to that?

SquidDev (changes requested):
> I wanted to avoid reading the file twice

Given we only read one line, it's not too bad. I suspect I'll probably end up rolling my own shell-specific `loadfile` for #1245, so might be able to remove that duplication then.

> Not that I expect anyone would

I think this is the crux: the limit is there to protect against accidental recursion rather than be anything the user would ever need to tune. I think we should set this a little higher (maybe 100, which is the `__index` limit) and remove the option.

projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
emmachase: Oops, thanks. And sure, I wasn't aware we were trying to phase it out :P

projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
emmachase: Ah yeah, I did this to maintain compatibility with how it works in POSIX. But I suppose it's not really going to be compatible anyway since we won't have any of the standard program paths probably.

MineRobber9000:
Hi! Dropping in uninvited to give my unsolicited two cents.

> Given CC doesn't have a `/usr/bin/env` equivalent, I think we probably want to obey the path.

Actually, we do. `shell`, when invoked with arguments, will run those arguments as if they were a command. So:

```
#!rom/programs/shell.lua cat
(cat in this case is a simple "read-the-file-and-spit-it-out" affair)
```

Would work just fine.

In light of this, I made a datapack that can be dropped into a world with 1.102.0 that adds shebang functionality in `os.run`. It doesn't respect the path, although, as covered above, that's not as much of a problem as it would seem. It reads the first line of the file, and, if it finds a shebang, it recurses (keeping the agreed-upon limit of 100 recursions, with a (to me) easier-to-understand error message when that limit is reached). If it doesn't find a shebang, it reads the rest of the file (reattaching the first line) and loads it, `loadfile` style.

[For your consideration.](https://github.com/cc-tweaked/CC-Tweaked/files/10327185/Shebangs.zip)

MCJack123:
One thing to note is that this won't strip out the shebang line when passed to the interpreter. For languages like Bourne shell script, where `#` is the comment character, this isn't an issue; however, Lua on its own throws a fit from this, and so will `shell` (which doesn't have *any* scripting capabilities right now... `shell a.txt` tries to run the file as Lua).

TIL that for whatever reason, Cobalt breaks Lua spec by removing the first line if it starts with `#`, **even in plain `load`**, so this isn't an issue I guess, but it's probably something to keep in mind. PUC Lua puts this detection in `loadfile` instead - maybe it's a good idea to have that there as a backup & assurance?

--------

Expanding a bit on `shell` not accepting scripts, there's really not much support at all for running scripts in CraftOS at the moment. `shell` takes a path argument, but it executes the file as a Lua script, which means "shell scripts" would need to be something like `#!/rom/programs/shell; shell.run("a"); shell.run("b")`, which is not very nice. `lua` doesn't take any arguments at all - if you try, it complains to you that it's a prompt only. When I wrote cash, I specifically made a wrapper around the `lua` command that allows you to run `lua file.lua` (alongside a convenient single-line execution syntax, `lua print(\"Hello World!\")`). There's no reason why `lua` can't just call out to `shell.run`/`os.run` if it gets an argument to a file.

In a perfect world, I'd say that we move running Lua files to `lua`, and make `shell` able to run command "batch" files (with support for `#` comments). However, that would break compatibility (your worst nightmare!) with things that depend on `shell file.lua`. Perhaps there could be an option like `-s` that tells the shell to execute batch scripts instead of Lua code - this would keep the ability to run Lua code as before, but also opens the possibility of proper shell scripting (ish) in CraftOS's shell.

Being able to bootstrap `multishell` with another program would be nice too - I had to patch this into CraftOS-PC to allow the debugger to use `multishell` with its own suite of tools, instead of booting straight into a plain shell. Some other commands that might be nice to have shebang support for might be `/rom/programs/command/exec`, `/rom/programs/monitor`, `/rom/programs/set`, and (a stretch) `/rom/programs/redstone` or `/rom/programs/help`.

SquidDev:
> Actually, we do. shell, when invoked with arguments, will run those arguments as if they were a command.

Yeeeeeeees. This is more of an accidental feature though, so does more hacky and less elegant than I'd like.

As an aside, I'm definitely not a fan of `os.run`[^1], and honestly would quite like to move away/deprecate it (the way it takes over error handling is definitely going to prove a problem in some of the stuff I want to do for #1245). Really don't want to be adding new functionality to it!

> TIL that for whatever reason, Cobalt breaks Lua spec by removing the first line if it starts with #, even in plain load, so this isn't an issue I guess, but it's probably something to keep in mind. 

Oh, interesting! I think I'm happy to leave the Cobalt side as-is for now, but we should possibly handle this on the Lua side.

> Expanding a bit on shell not accepting scripts, there's really not much support at all for running scripts in CraftOS

Correct, and I think that's probably for the best.

> Being able to bootstrap multishell with another program would be nice too

Can you open an issue explaining exactly you're trying to do?

projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
```diff
@@ -54,6 +54,67 @@ else
...
+            printError("Interpreter recursion depth limit reached when loading file: " .. path)
+            return false
+        end
+
+        -- Load the specified interpreter instead
+        local args = tokenise(contents:sub(3))
+        local originalInterpreterPath = table.remove(args, 1)
+        local resolvedInterpreter = shell.resolveProgram(originalInterpreterPath)
+        if not resolvedInterpreter then
+            printError("Interpreter not found: " .. originalInterpreterPath)
```
SquidDev: I think `interpreter` is probably the wrong terminology to be using (yes, I know it's what POSIX calls it). Maybe "Hashbang"? So "Hashbang program not found" here?

projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
```diff
@@ -54,6 +54,67 @@ else
...
+-- Execute a program using os.run, unless a shebang is present.
+-- In that case, execute the program using the interpreter specified in the shebang.
+-- This may occur recursively, up to the maximum number of times specified by remainingRecursion
+-- Returns the same type as os.run, which is a boolean indicating whether the program exited successfully.
+local function executeProgram(remainingRecursion, env, path, ...)
+    local file = fs.open(path, "r")
+    if not file then
+        printError("File not found: " .. path)
+        return false
+    end
```
SquidDev:
```suggestion
    local file, err = fs.open(path, "r")
    if not file then
        printError(err)
        return false
    end
```

I wonder if it'd be better to do this instead? I don't know.

SquidDev (changes requested):

projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
```diff
@@ -54,6 +54,67 @@ else
...
+        end
+
+        -- Add the path and any arguments to the interpreter's arguments
+        table.insert(args, path)
+        for _, v in ipairs({ ... }) do
+            table.insert(args, v)
+        end
+
+        env.arg = { [0] = resolvedHashbangProgram, unpack(args) }
+        return executeProgram(remainingRecursion, env, resolvedHashbangProgram, unpack(args))
```
SquidDev:
```suggestion
        env.arg = { [0] = resolvedHashbangProgram, table.unpack(args) }
        return executeProgram(remainingRecursion, env, resolvedHashbangProgram, table.unpack(args))
```

`unpack` is notionally deprecated, as it's a Lua 5.1-only feature.

One alternative here would be to do what PUC Lua does, and instead of passing varargs through the functions, do `os.run(env, path, table.unpack(env.arg))`. Not sure if it's any nicer though - will leave that up to you!

projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
```diff
@@ -87,7 +148,7 @@ function shell.execute(command, ...)
         local sDir = fs.getDir(sPath)
         local env = createShellEnv(sDir)
```
SquidDev: I think we need to push this into `executeProgram` too, and use the resolved path of the hashbang program/interpreter instead of the original file. The problem here is that CC's `require` implementation resolves modules relative to the program's directory (rather than PUC Lua, which uses the current directory), so currently the hashbang program will use the wrong program path!

This might be a little cleaner if we do my above suggestion with passing arguments (so we'd have `executeProgram(100, sPath, { [0] = command, ... })` here).

SquidDev (changes requested):
Sorry for the rather tortuous review process, just noticed a couple of things which I'd entirely missed on earlier review rounds.

Other than my comments, I'm pretty happy with this though! Going to let this sit today in case anybody else has thoughts, then will merge :).

SquidDev:
Oh, one thing I forgot to mention is that we should probably document this feature somewhere (though happy to do that Emma, mostly just noting to remind myself!).

Not sure where the best place to do this is - putting it in the description of the [`shell` API docs](https://tweaked.cc/module/shell.html) might make the most sense, though also feels a little weird with how sparse the current docs are. Possibly should flesh that out first :).

Don't know if people have thoughts?

SquidDev (approved):

SquidDev:
I'm hoping to get a 1.102.2 out with a fix for #1282, then will release a 1.103.0 with this and #1298.

I still need to write some docs D:. I have had several goes at doing so, but have only succeeded in writing 3 words.

SquidDev:
@emmachase I've just pushed some docs. Just want to check you're happy with those, and then will merge :).

emmachase:
Looks good to me!

SquidDev:
Amazing! Thank you for all your work on this :heart:.

PR #1275, opened 2022-12-28
Fix crash on Fabric when attempting to use a non-fuel item as fuel.

emmachase:
On the fabric port, trying to use any non-fuel item as fuel in a turtle causes the game to crash.
```
Description: Ticking block entity
java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "net.fabricmc.fabric.api.registry.FuelRegistry.get(net.minecraft.class_1935)" is null
	at dan200.computercraft.shared.platform.PlatformHelperImpl.getBurnTime(PlatformHelperImpl.java:245)
	at dan200.computercraft.shared.turtle.FurnaceRefuelHandler.getFuelPerItem(FurnaceRefuelHandler.java:38)
```

Reproduction steps:
1. Place a turtle
2. Place a non-fuel item in it's inventory, such as `minecraft:grass_block`
3. Run `refuel all` in the turtle.


SquidDev:
Thank you very much!

Sad that our static analysis didn't catch this. Fabric is missing a `@Nullable` annotation, and not 100% sure if ErrorProne correctly handles the implicit cast nullable `Integer`s.

PR #1277, opened 2022-12-29
Fix crash when turtle digs with a full inventory and any slot other than slot 1 is selected

emmachase:
There appears to be a bug where the game will crash if a turtle digs while having any slot other than the first selected, and that slot plus all slots after that are full or don't match the dug item.

Reproduction steps:
1. Place a mining turtle
2. Execute turtle.select(16)
3. Fill its inventory with dirt
4. Place a stone block in front of the turtle
5. Run turtle.dig()


SquidDev:
Uggghr, when writing the inventory transfer code, I was even aware that I might screw up stack-size-limits and slot-count-limits.

Good spot!

PR #1298, opened 2023-01-10
Better Lua parse errors

SquidDev:
This provides a custom Lua parser, which is used to provide more user friendly errors when a program has syntax errors:

image

This is _heavily_ WIP right now, but I've been been sitting on this for a few weeks, so felt I should probably publish this as a PR just to force me to finish this.

## Usage
I'd really appreciate some feedback on this PR. Any feedback is useful: the quality of error messages, how they're printed, any things you think are missing, etc...

I've put together a standalone version of this PR, which can be temporarily used by running the following on a computer:

```
wget run https://gist.githubusercontent.com/SquidDev/9e72bc7d8fec6e98358b92bd52599c9e/raw/install-cct-1298.lua
```

Alternatively, save it to a startup file and then reboot:

```
wget https://gist.githubusercontent.com/SquidDev/9e72bc7d8fec6e98358b92bd52599c9e/raw/install-cct-1298.lua startup/10_syntax_errors.lua
```

## Summary
As mentioned in #1245, Lua's parse errors can sometimes be a little confusing, especially to new users. This PR adds a pure Lua parser, which provides (hopefully!) more friendly errors when a syntax error occurs.

Some quick notes on the implementation here:
 - The code for this lives in the _internal_ `cc.internal.syntax` module. This contains a lexer, parser, a collection of error messages, and a module for pretty-printing error messages.

   As with any `cc.internal` module, this shouldn't be used from user code (yet!). I do want to change this in the future, just need some time to let the implementation settle.

 - The parser is generated from a Lua grammar and some parse error rules. I'm not thrilled about this - it does make this code a pain to maintain for anyone who isn't me[^1]. I did start writing a standard recursive descent parser, but had some issues fitting some of the error messages I wanted into it.

 - This currently only reports errors for files invoked directly by the shell. This does _not_ include any other Lua files/strings, such as those loaded by `require`, `os.loadAPI`, or `load` directly.

   `edit` and `lua` should definitely be updated to use this mechanism too. I'm less sure about how to handle `require` - thoughts welcome!

## Still to do
 - [x] Tests! I've got a load of golden tests locally, but these need to be converted to CC's test framework.
 - [x] Flesh out the error messages: I've mostly focused on handling the fancy cases, and not the boring ones :).

   I think we probably need a style guide for error messages and make sure everything sticks to that. Elm does a really good job here for instance, though I'm not sure I agree with their choice to personify the compiler.

 - [x] Improve error printing: the current layout is pretty unreadable on long lines or pocket computers. We need to handle this much better.
 - [x] Get some actual user feedback. No clue how!

[^1]: Note I will open source the tooling to generate this file. It's just written in OCaml, so...


SquidDev:
This will conflict with #1273 - will need to merge that first.

emmachase:
This is perhaps a matter of design opinion but I think the way the annotations are displayed in the error message look a tad silly if it's just one annotation. IMO, unless there are 2 or more annotations, it should simply underline the affected portion, and put the message on a separate line.

For example, this one looks great, no changes needed:
image

But, for a single annotation, i would prefer this:
image
Or something similar.

Lemmmy:
my mockup of this concept from earlier (intended for pocket computers)
image

SquidDev:
> This is perhaps a matter of design opinion but I think the way the annotations are displayed in the error message look a tad silly if it's just one annotation. 

Oh, that's useful to know! I think it does look better [when the token is more than one character long](https://user-images.githubusercontent.com/4346137/212469110-5e25749d-5a65-445e-96f0-979e7aca9334.png) (though do say if you disagree!), but agreed that it looks bad with one single character.

I think some other options here, not all of which are mutually exclusive.

 - Remove the text next to the underlined text. I think in many cases ("`=` appears here" included), it doesn't really add much. Though there's some cases ([such as our error for standalone names](https://user-images.githubusercontent.com/4346137/212469602-f6b22451-16f0-4395-9431-df4f07bdeb2f.png)) where this might be harder to do.

   The error messages there are incredibly unpolished, so feedback there is very much welcome :).

 - Shift the marker to the right, so it has a more natural flow:
   
   A screenshot of a computer with a syntax error, similar to the original post. The "connector" between the error message and corresponding text has been moved slightly, making it flow more naturally from left to right.

 - Always use "separate line" format (like both Lem and Emma posted). The original plan was just to use this for pocket computers (due to the limited horizontal space), but maybe it's better to use it everywhere!

   My main concern is how to handle this on non-advanced terminals. It would be nice to number these messages instead of just relying on colour.


Not sure. Thoughts!?

Lupus590:
> My main concern is how to handle this on non-advanced terminals. It would be nice to number these messages instead of just relying on colour.

You could just not implement it on non-advanced terminals. It might delay new people getting into the mod but the advanced computers are easier to use in other ways already so maybe it won't be a big loss.

SquidDev:
> You could just not implement it on non-advanced terminals

Given the aim of this is to make things easier for new users, I think it really has to be usable on normal computers too.

MCJack123:
Could we put this behind a `settings` flag? Power users like myself may prefer the "dumb" error messages for terseness.

Also, if this isn't implemented yet, long lines should be truncated with `...` or some single character, with only the relevant portion displayed in the message. (Wouldn't it be nice if ISO-8859-1 had a … character…)

SquidDev:
> Could we put this behind a settings flag? Power users like myself may prefer the "dumb" error messages for terseness.

Maybe, though I'd much rather get feedback on where the error messages are too long and can be improved first.

> Also, if this isn't implemented yet, long lines should be truncated with ... or some single character, with only the relevant portion displayed in the message. (Wouldn't it be nice if ISO-8859-1 had a … character…)

No, not implemented yet - all the error message printing code is very WIP right now. We'll probably have to both truncate the strings at both the start and the end, so will need to be careful about how to do that without making everything look terrible.

SquidDev:
Just rewrote the error printer to be a little more sensible. A quick summary of the changes:

We now use french quotes to indicate when the line has been truncated on the left and right.

image

The markers on the underlines now sit slightly more to the right.

image

If the underlines are too far apart, we'll split them onto multiple lines:

image

I think it's definitely better, but still a long way away from perfect. I will have a fiddle with the alternative design Lem and emma mentioned, and see what I think about that.

Any and all feedback is appreciated here! Will try to put together a more standalone version of this, so people can play around with it on existing computers.

SquidDev:
While there's still a lot of work to be done on this, I think it's probably at the point where it'd be helpful if other people tried it out. I've set up a small script which installs this PR on a computer:

```
wget https://gist.githubusercontent.com/SquidDev/9e72bc7d8fec6e98358b92bd52599c9e/raw/install-cct-1298.lua startup/10_syntax_errors.lua
```

It'd be super helpful if people are willing to run this for a few days and post any issues they have, but any feedback is appreciated! General feelings on the whole feature, any places where the error messages are bad or confusing, ways to improve the display of error messages. Anything!

SquidDev:
I think the solution for pocket computers is to show each annotation separately:

image

Lupus590:
> I think the solution for pocket computers is to show each annotation separately:

I'm assuming that this is used when the screen width is too narrow for the other version.

Now I'm wondering if we could have a setting to always use this view, even when the screen is big enough.

SquidDev:
> I'm assuming that this is used when the screen width is too narrow for the other version.

Yep! It swaps over if the screen is less than 35 characters wide.

> Now I'm wondering if we could have a setting to always use this view, even when the screen is big enough.

Personally I'm loathe to add settings except when strictly needed. Though I'm curious as to why you're asking - do you prefer this view in general?

Lupus590:
> do you prefer this view in general?

I think I do? Maybe I'm finding it a little easier than following long lines that sometimes cross over each other.

Although, I'm now thinking it might work well for non-advanced computers as it's not as dependent on colours as much as the crossing lines version.

SquidDev:
> I think I do? Maybe I'm finding it a little easier than following long lines that sometimes cross over each other.

That's good to know! The long lines design was based on [ariadne](https://github.com/zesterer/ariadne), but it definitely doesn't work as well for CC's smaller screen resolution.

How do you find [this version](https://user-images.githubusercontent.com/4346137/214161086-12ddfe4a-2701-488c-a943-1c1b9ba4306b.png) out of interest? My worry with the pocket-computer layout is that the underlines aren't obvious enough, but maybe it's not really a problem.

Lupus590:
> How do you find [this version](https://user-images.githubusercontent.com/4346137/214161086-12ddfe4a-2701-488c-a943-1c1b9ba4306b.png) out of interest?

That looks fine, but I kind of prefer everything left aligned. It's still better than the crossing lines, if that's what you were asking.

SquidDev:
Right, let's give this a go. It's probably the best way to get a lot of feedback, and it'll be fairly easy to back out if it ends up a disaster!

PR #1304, opened 2023-01-16
Update event docs

ouroborus:
Add "return values" for those events that only return their name. Add break in return values for `computer_command` event.


SquidDev:
Amazing, thank you! I've changed how we display varargs (like in computer_command), which should avoid the need for `<br>`.

Issue #125, opened 2019-02-23
turtle.use() revival


fatboychummy:
[So this was requested a while ago, but moved to dan200's CC repo](https://github.com/SquidDev-CC/CC-Tweaked/issues/58).  However, dan200's repo is kind of... dead now, and I figured this could use a revival.  

A useful feature for us turtle users would be a `turtle.use` function.  Why?

Well, first off, it helps the newbies out a lot.  When I first started in computercraft and I was making a farmer bot, I didnt know that you had to call `turtle.dig` above the block I wanted to hoe.

Also, it can open up some opportunities for more features in the future.  If you set it up so it acts like a right-click, you could also right-click on blocks to change their configuration.  

`turtle.dig` could keep it's normal function, so backwards compatibility is still good.  I just feel it would be much easier to understand if there was `turtle.use`

znepb:
This would be great if it would act as right-clicking. Like placing a turtle, and using `turtle.use` and turning a bucket to milk, if there is a cow in front. This would also be good as a replacement to `turtle.place` for water and lava buckets.

SquidDev:
My stance has always been that this is best served by a peripheral mod, such as plethora. Using arbitrary items is a sufficiently powerful feature that it should be gated behind an upgrade, and I don't want to add such a thing to core CC:T.

PR #1305, opened 2023-01-16
Cauldrons and Bucket / Bottle interaction (and levers/buttons....)

Lindsay-Needs-Sleep:
First time messing around in a mod, so feedback welcome! 

This attempts to use the selected item on the hit block on `place`.  (Only attempts if the block is right beside/above/below to not interfere with actually placing blocks)

This enables:  
- turtles to fill/empty their buckets/bottles at cauldrons. (My main purpose)
- It also has the side effect of enabling turtles to interact with adjacent lever/buttons.
- You can also collect honey from bee hives with bottles, and shear honeycombs with shears (though they drop as items and don't directly go into the turtle's inventory)

Consequently changed behaviors:  
- If a chest is right beside a turtle, and you `place` a water bucket on it, nothing happens. (Previously it would waterlog the chest.)  But, you can still waterlog chests by calling `place` when the chest is not immediately beside the turtle.

Feedback:  
- Maybe there is something already existing to replace `areAdjacent`?
  - Or a better name? o.o;;
- Can anyone think of any other unintended side effects?

---
Tests

`./gradlew check` No new tests fail.  (These 2 failed before I made any changes, so I don't think it was me)

```
> Task :core:test

ComputerTestDelegate > get() > The os library > os.date > produces output consistent with PUC Lua > dan200.computercraft.core.ComputerTestDelegate.get()[73][1][1][26] FAILED
    org.opentest4j.AssertionFailedError at ComputerTestDelegate.java:420

ComputerTestDelegate > get() > The os library > os.date > produces output consistent with PUC Lua > dan200.computercraft.core.ComputerTestDelegate.get()[73][1][1][30] FAILED
    org.opentest4j.AssertionFailedError at ComputerTestDelegate.java:420
```

### DIFF
--- a/projects/common/src/main/java/dan200/computercraft/shared/turtle/apis/TurtleAPI.java
+++ b/projects/common/src/main/java/dan200/computercraft/shared/turtle/apis/TurtleAPI.java
@@ -212,8 +212,8 @@ public class TurtleAPI implements ILuaAPI {
      * Place a block or item into the world in front of the turtle.
      * <p>
      * "Placing" an item allows it to interact with blocks and entities in front of the turtle. For instance, buckets
-     * can pick up and place down fluids, and wheat can be used to breed cows. However, you cannot use {@link #place} to
-     * perform arbitrary block interactions, such as clicking buttons or flipping levers.
+     * can pick up and place down fluids, and wheat can be used to breed cows. You can use {@link #place} to
+     * perform arbitrary block interactions (such as clicking buttons or flipping levers) if the turtle is holding a bucket or bottle.
      *
      * @param args Arguments to place.
      * @return The turtle command result.
--- a/projects/common/src/main/java/dan200/computercraft/shared/turtle/core/TurtlePlaceCommand.java
+++ b/projects/common/src/main/java/dan200/computercraft/shared/turtle/core/TurtlePlaceCommand.java
@@ -30,6 +30,7 @@ import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.Block;
 import net.minecraft.world.level.block.entity.BlockEntity;
 import net.minecraft.world.level.block.entity.SignBlockEntity;
+import net.minecraft.world.level.block.state.BlockState;
 import net.minecraft.world.phys.BlockHitResult;
 import net.minecraft.world.phys.EntityHitResult;
 import net.minecraft.world.phys.Vec3;
@@ -177,7 +178,7 @@ public class TurtlePlaceCommand implements TurtleCommand {
         var item = stack.getItem();
         var existingTile = turtle.getLevel().getBlockEntity(position);
 
-        var placed = doDeployOnBlock(stack, turtlePlayer, hit).consumesAction();
+        var placed = doDeployOnBlock(stack, turtle, turtlePlayer, hit).consumesAction();
 
         // Set text on signs
         if (placed && item instanceof SignItem && extraArguments != null && extraArguments.length >= 1 && extraArguments[0] instanceof String message) {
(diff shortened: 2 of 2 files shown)


SquidDev:
Thank you very much for working on this!

I should probably give some background on the turtle placing code, as the intention and implementation are pretty gnarly:

There's always been a bit of a discussion/debate about how much turtles should be able to interact with the world, and whether they should be able to right click arbitrary items/blocks (i.e. #125). My stance has generally been pretty conservative, in that turtles should be able to place blocks, and right click items into/against blocks (such as the cauldron and bee hive interactions), but shouldn't be able to use items and blocks arbitrarily (such as pressing buttons).

However, Minecraft's interface doesn't really allow this at all:
 - `Item.useOn` is used for placing blocks and items, that's fine. Except:
 - `Item.use` is used to place bucket of water and lillypad (which we want). However, it's also used to eat food and cast fishing rods (which we don't)[^1]. 
 - `Block.use` is used to place water into cauldrons (which we want), but again is also used to flip levers (which we don't).

This gets even messier as which method implements which interaction changes over time. Cauldrons _used_ to work until 1.18 (or 1.17, I can't remember!), as the interaction was moved from `Item.useOn` to `Block.use`.

---

The TLDR here is that turtles not being able to right click blocks _is_ intended, but also maybe the reasoning for that is silly and just causing more problems than it solves.

I'm afraid I'm going to sit on this for a few days and have a think before reviewing this any further. Thoughts from other people very much welcome!

Very sorry for the delay, especially on your first modding PR!

[^1]: We used to limit `Item.use` to a couple of specific items. I relaxed that in 8a7156785deaf49af054fbfa06671f314ff9857b, because the hard coding upset me, but it does now have the interesting behaviour that you can throw an ender pearl from a turtle. Albeit only once - we don't handle cooldowns properly.

Lindsay-Needs-Sleep:
Np!  Thank you for the beautifully formatted (and informative comment!)

I must say I share the same sentiment (bucket/beehives: yes, levers/buttons/etc: no).

Although it would be fairly amusing seeing a turtle going around pressing buttons and flipping levers, this is really not necessary since we already have the redstone api.  Additionally, it's just kind of funny to call `place` to use a button (with the requirement that the turtle is holding an item like a bucket). xD

I was hoping it was just my lack of knowledge, and there would be better way to filter than `if cauldron/beehive` but it sounds like that is not the case.  
(Big thank you for the function breakdown!  It confirmed what I *thought* I was understanding.)

The other thing I just tested, this allows using any hoe in the selected slot with `place` to till the ground.  This does have the draw back of using up durability, but allows you to skip the diamond requirement (not really a fan of that).  But I suppose this is also fairly analogous to sheering sheep with `place` (which currently works).

The linked issue is interesting.  If there was a chance of `use` coming in, I must say I kind of agree with making any kind of bucket use fall under a `use` (vs. `place`).  Limiting it to certain items/interaction makes sense.  The dislike for hardcoding the items, makes even more sense though!  Opening it up to let it `use` whatever is an interesting thing.  I suppose it would give players a lot more freedom with their turtles, but it does remove some of the limitations, which are admittedly a fun part. :p  

This is certainly a fairly substantial consideration, so please take your time!

MineRobber9000:
Adding my two cents where it wasn't asked for again, but you could probably use tags to decide what items can be used and on what blocks (`computercraft:turtle_use_items` and `computercraft:turtle_use_blocks` or similar naming). That way, if someone wants their turtles to be able to press buttons/flip levers/whatever, they can just add `#minecraft:buttons` and `minecraft:lever` (or whatever the button tag and lever item ID are resp.) to `computercraft:turtle_use_blocks` with a datapack, and the base mod can just handle cauldrons and beehives and whatever else we want turtles to be able to interact with without changes.

SquidDev:
Sorry for the slow response here - tried to take the last week off and work on some other projects.

I had considered tags and then discounted them as they're still somewhat hard-coding a list. However, @MineRobber9000's point about them being user-customisable is something I hadn't considered before, and I think probably a good solution to this.

I've switched the usage of `Item.use` to check a tag in 2e5cd29e12db67c925c4ce5bd114c62d1bbae438. @Lindsay-Needs-Sleep, would you be willing to merge that into this branch and then do the same for `Block.use`? I imagine we'd just want cauldrons and hives in there right now - not sure what else is helpful!

Lindsay-Needs-Sleep:
Sorry I seemed to have missed your comment somehow.   D:
My dev environment got nuked for this, I'm not sure when I'll have time again. >.<

But the tags sound great!

PR #1317, opened 2023-02-03
Add `id` to the GPS response packet

SkyTheCodeMaster:
This adds an `id` field to the GPS response packet.
This idea came from some discussion in [SwitchCraft chat](https://i.skystuff.games/P0ht1gYBJdtN.png) about figuring out what dimension a player is in.
This can be used to check against a list of GPS hosts that are known to be in a certain dimension, like:
```lua
local hosts = {
  [1] = "minecraft:overworld",
  [2] = "minecraft:nether",
  [3] = "minecraft:end",
}```
Then a program could just pull the `modem_message` event itself and check the `id` parameter against the list of hosts.


Lemmmy:
(the part of the conversation where I did not endorse this idea was conveniently cropped out)

SquidDev:
Time is a circle: #341.

PR #1320, opened 2023-02-05
Pretty runtime error messages

SquidDev:
A screenshot of an in-game computer. The user has just run a program containing "rednet.recieve()", which errors with "attempt to call field 'recieve' (a nil value)". The exciting thing about this error is that we print the source code alongside the error, underlining the place the error occurred!

As discussed in #1245, Lua's errors can sometimes be a little unhelpful, and it would be nice if we could provide slightly friendlier versions. #1298 hopefully improved the quality of parse errors, but runtime errors are still rather lacklustre.

This PR updates the shell (and a few other places) to print the source code alongside an error, underlining the place the error occurred.

This requires several quite invasive changes, so I would like some feedback on this first!
 - Cobalt now includes column numbers in its debug information (see https://github.com/SquidDev/Cobalt/pull/65). This can be read with `debug.getinfo(...).currentcolumn` (much like `currentline`).

   Note this information is _not_ included in `string.dump` (and so will not be present when loading binary chunks either). I think that's fine.

 - As Lua's errors as just normal strings, there's no way to rethrow an error and preserve the original error location. This is a problem with things like the `parallel` API, where the error occurs inside a separate coroutine.

   To solve this we add a new `cc.exception` module, which provides a way of working with slightly more structured errors. It currently provides the following functions:
   - `is_exception(value): boolean`: Check if a value is an exception.
   - `wrap_error(message, coroutine)`: Possibly convert an error and its source coroutine into an exception. This will convert exceptions (effectively rethrowing them) or string-based error messages: other failures such as `error()` or `error({})` will not be wrapped.
   - `try(func, args...)`: Like `pcall`, but wraps the error with `wrap_error`.

   I am not thrilled by the API here, so any thoughts are welcome here.

 - In order to distinguish between Lua REPL inputs, we now name each input `lua[1]`, lua[2]`. Which is _fine_, but does look rather nasty in error messages. Don't know if there's a good alterative!

## Remaining work
 - [ ] Design an API I'm happy with.
 - [x] Correctly offset columns in the REPL - they currently don't take into account the `return _echo(` we inject!
 - [ ] Write tests!


SquidDev:
Just to note the build failure is expected - I've not published the changes to Cobalt to Maven yet.

emmachase:
Something that I've given little thought to, but was the first thing on my mind: Would it be possible to, in the case of "attempt to call nil field" like in your example, look at the fields that do exist and try to find what they were looking for? So in this case, it may say something like: "Did you mean `receive`?".

SquidDev:
>  Would it be possible to, in the case of "attempt to call nil field" like in your example, look at the fields that do exist and try to find what they were looking for? 

Yeah, I would really like to do this eventually. 

I can definitely think of a ways to implement it (extract the code with some Lua patterns, then use `debug` to find the right variable). It's a bit of a bodge, but it'd cover the common cases at least[^1]. It might be nicer to do something on the VM side (for instance, by inspecting the bytecode), or even with a proper parser - not sure!

I think it's probably something I want to leave until later (will add it to the todo list on #1245), but if people want to prototype some pure-Lua implementations I think that would be valuable.

[^1]: It wouldn't work for something like `f().x`, but I don't think that'd ever be feasible. The Lua VM basically compiles that down to `tmp = f(); tmp = tmp.x`, so you don't have access to `f()` any more.

SquidDev:
> Design an API I'm happy with.

This is proving to be quite tricky. There's a couple of areas I'm not sure about:
 - The current design requires that an exception has a (dead) coroutine attached to it. This is nice (it gives a structured way to inspect the stack trace, debuggers could theoretically use it to read locals), but I worry that it's too general, and will only box us into a corner.

   One specific concern I have is how this will integrate with `coroutine.close` if we ever update to 5.4. When a coroutine is closed, its stack trace is cleared. This means we need to be careful about when coroutines are closed, to make sure we're not preventing the exception from containing useful information.

   We could make the `thread` object optional (or even internal), and instead just provide some views over this data (such as a way to get the `debug.traceback` and an iterator over the individual frames), though not sure if that's any better!

 - In order to prevent interfering with other code's error objects, we don't convert non-string errors into exceptions. However, this also means that libraries can't use structured errors and exceptions together.

   This is probably less of a blocker for merging, as I'm more confident this is something that can be added rather than affecting the internals!

I might merge the core error reporting functionality (i.e. `bios.lua`, `shell`, and `lua`) for 1.103.0, and then spin out the exception work (and changes to the `parallel` API) into a separate discussion where we can get more feedback.

PR #1341, opened 2023-02-17
Small update to help newbies a bit.

fatboychummy:
A common error with tables is that you miss a comma whilst defining them. For those who know Lua, searching for missing commas on a "missing closing bracket" error is just a "standard procedure." However, for those who are new to Lua, they'll see the error and likely take it at face-value, looking instead for some spot they may have missed a closing bracket.

Reasoning: a table missing a comma outputting an error stating it is missing a closing bracket.

# Quick notice
I don't know if this module has any tests or if this will affect anything else negatively. I just had the idea today when looking at an error someone got on the CC:Aide bot and decided to quickly throw in a PR from my phone while at work.

## A quick checklist

### DIFF
--- a/projects/core/src/main/resources/data/computercraft/lua/rom/modules/main/cc/internal/syntax/errors.lua
+++ b/projects/core/src/main/resources/data/computercraft/lua/rom/modules/main/cc/internal/syntax/errors.lua
@@ -528,7 +528,7 @@ function errors.unclosed_brackets(open_start, open_end, token, start_pos, end_po
 
     -- TODO: Do we want to be smarter here with where we report the error?
     return {
-        "Unexpected " .. token_names[token] .. ". Are you missing a closing bracket?",
+        "Unexpected " .. token_names[token] .. ". Are you missing a comma or a closing bracket?",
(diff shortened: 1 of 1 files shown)


piprett (approved):

emmachase:
I think ideally in this case we would match this as a separate pattern as opposed to just changing the error message. So that we can show a hint saying "did you mean to put a comma here? ^"

SquidDev:
So definitely agreed it'd be good to show information about the missing comma.

The current error messages are pretty conservative in what context they show, as I've found it quite hard to write more focused error messages which don't end up with a lot of false positives. For instance, we wouldn't want to be warning about a missing comma in the case of a `print({ 1, 2, 3 )`.

I wonder if a suitable heuristic here is to check if the next token is an expression (so is an identifier, string, number or true/false/nil), and alert about a missing comma. Otherwise keep the existing error?


---

As an aside, please continue to report issues like these! If you spot cases where we're generating bad syntax errors (or, like this, some common case that we're missing), it's definitely something I want to improve :).

SquidDev:
Sorry, ended up writing a custom parse error rule for this instead, ended up being a little easier!

Issue #901, opened 2021-08-20
Gps.locate uses stale ping responses from older pings if they are still in event queue.


Wojbie:
### Minecraft Version

1.16.x

### Version

1.98.1

### Details

When running `gps.locate()` function multiple times in row one can accidentally cause it to give out incorrect coordinates due to it using older ping responses that were still in event queue. This happens due to fact that it will use first set of pings that give it valid coordinates and leftover pings form any extra gps hosts will stay in queue to be picked by another call.

Easiest way to visualize the bug is to run this code on pocket:
```lua
local a,b,c
while true do
 local x,y,z = gps.locate()
 if x ~= a or y ~= b or z ~= c then
  print(x,y,z)
  a,b,c = x,y,z
 end
end
```
Moving straight up generates this output.
![obraz](https://user-images.githubusercontent.com/5893536/130279032-a75e49c6-f9f8-4654-aca7-80b8cb9f8c5e.png)

Issue can be partially negated by adding code that clears out current events from queue example:
```lua
local a,b,c
while true do
 os.queueEvent("gpstest")
 os.pullEvent("gpstest")
 local x,y,z = gps.locate()
 if x ~= a or y ~= b or z ~= c then
  print(x,y,z)
  a,b,c = x,y,z
 end
end
```
But that don't resolve the issue as there is always possibility that there are still gps hosts that have not yet responded to old ping.

This code resolves issue in single run cases by waiting 0.05 s before sending first ping...
```lua
local a,b,c
while true do
 sleep(0.05)
 local x,y,z = gps.locate()
 if x ~= a or y ~= b or z ~= c then
  print(x,y,z)
  a,b,c = x,y,z
 end
end
``` 

But i am not sure if it is a proper solution as it feels hacky and I believe (but was unable to test) that multiple `gps.locate()` calls ran in parallel may still use stale data created by other one if timing happens to be bad...

I spend some time thinking if there is a backwards compatible way to update ping protocol to have unique ids for each ping to ensure that any locate will only listen to its own ping responses but sadly i have not found totally backwards compatible way..
One of solutions would be to move to a different gps protocol that sends some kind of PING table with unique id inside instead of "PING" string and have hosts still respond to both "PING" strings as well as new pings with tables of `{ x, y, z, ["id"]=pingID }`.

Here is another image showcasing the issue using plethora lines: 
[...]

PR #1344, opened 2023-02-18
Remove duplicate pings received by gps.locate to avoid math errors.

Wojbie:
### Problem

GPS api trilaterate and narrow math is expecting to receive unique positions and distances to preform calculations on.

In some cases it is possible to single `gps.locate()` call to receive more then 1 ping from same GPS host which leads to math used to end up with 3 nan values as return.

Possible sources of second ping response is either simultaneous GPS locate call called by another computer or stale ping response in event queue from older call as seen in #901. This is an problem was was in GPS api from very beginning and was exacerbated as side effect of #341 moving all pings into single channel, which allowed for detection and this fix of it.

### Solution
This patch adds simple logic to disallow for use of non-unique position by checking if that position don't already exist in fixes table.
Additionally in case that ping is a duplicate it updates distance stored to latest known one to partially decrease #901 effect.


SquidDev:
Sorry for the delay in looking at this! I wonder if a better solution is just to add an additional check at the top of the `trilaterate` function, which ensures that either point isn't too close to another. This should also handle issues which might occur from having GPS hosts very close to each other:

So replace:

https://github.com/cc-tweaked/CC-Tweaked/blob/111f971ed0f5d4f49849229b4736af60441db6ac/projects/core/src/main/resources/data/computercraft/lua/rom/apis/gps.lua#L39-L41

with something like:

```lua
-- Bail if all our points are close to each other or on a plane.
if a2b:length() < 1 or a2c:len() < 1 or math.abs(a2b:normalize():dot(a2c:normalize())) > 0.999 then
    return nil
end
```

I guess this would be wrong if the first two points are received from the same host (so `a2b:length() == 0`). Maybe we need some better logic here for selecting points in general.

Wojbie:
Hmm, this would work too but would require also changing logic of what points we feed into `trilaterate` as currently we feed first 2 points and current one. So if first two fixes were same it would never get to any solution and timeout. It might make sense to rethink a logic here but i can't really thing of any alternatives that would be better then others. Like if we only use last 3 then there is change it never gets a fix? Check all possible combinations of 3? That would increase how much math gps does.

SquidDev:
Sorry for taking so long to get round to this. Adapted your fix slightly and applied it manually in eef67a04a4593396a8d2db9acaf8839e48465a2d.

PR #1349, opened 2023-02-28
Fix introduction dates of os.cancel[Alarm|Timer]

MCJack123:
This PR updates the `os.cancelAlarm` and `os.cancelTimer` documentation to note that both functions were introduced in CC 1.6. This information was not mentioned in the changelog for 1.6, and as such was missed in the initial pass of `@since` additions.


SquidDev:
Ahhh, good spot and good investigation work!

Issue #1339, opened 2023-02-16
Relicensing (bits of) CC: Tweaked


SquidDev:
I've been wanting to do this for years, and seriously considering it for about 4 months now, so might as well get the ball rolling somehow!

## The Problem
When ComputerCraft was originally open sourced, it was released under the CCPL (ComputerCraft Public License), a derivative of the MMPL (Minecraft Mod Public License). Unfortunately, both the MMPL and CCPL have a couple of issues, which range from irritating to gaping holes:

 - It's not a standard OSI-approved license. While in some sense this is purely a cosmetic issue, it has stopped several people contributing, and also blocks us from participating in some initiatives such as [Modtoberfest](https://modtoberfest.com/).

 - [Code licensed under the CCPL can only be used in other Minecraft mods](https://github.com/cc-tweaked/CC-Tweaked/blob/mc-1.19.x/LICENSE#L83-L89), and that code must be licensed under the CCPL. This means that any usage of the code in a non-mod project (including emulators or CC programs) violates the license.

   This is an incredibly restrictive clause, to the point that it's counterproductive. I can think of half a dozen (well meaning and useful!) projects off the top of my head which violate this clause.

 - [All dependencies of ComputerCraft must be licensed under conditions comparable to the CCPL](https://github.com/cc-tweaked/CC-Tweaked/blob/mc-1.19.x/LICENSE#L72-L74). We inherit this condition from the MMPL, and it's definitely a problematic one. While I do not believe it's the intention, the practical outcome is that all dependencies must be licensed under the CCPL. While there are exceptions for Minecraft and the mod loader (Forge/Fabric), there aren't any exceptions for other dependencies, such as the Java Standard Library or Netty. 

   This suggests that even ComputerCraft itself violates its own license, a position which is clearly untenable!

## The difficulty of relicensing
Given CC/CC:T's license is somewhere between awkward and untenable, I would like to relicense CC: Tweaked under something more standard and permissive.

Relicensing a project requires getting every single previous contributor to sign off on the new license. This is incredibly hard for a couple of reasons:

Firstly, CC: Tweaked has had 98 people contribute to it. While 20-30 of these people have only committed trivial changes, which don't require permission, this is still a lot of people and will be a slow process. I remember Forge's relicensing process took months, and I don't think BuildCraft ever managed to complete theirs!

[...]

PR #1351, opened 2023-03-01
License CC:T according to the REUSE specification

SquidDev:
This is the initial stage of the relicensing work, as set out in #1339.

This adds SPDX license headers to all source code files, following the [REUSE] specification. This does not include any asset files (such as generated JSON files, or textures). While REUSE does support doing so with ".license" files, for now we define these licences using the `.reuse/dep5` file.

Any files where I am the only contributor I've licenced as MPL-2.0, copyright "The CC: Tweaked developers". I think this is a little more friendly than "Jonathan Coates and contributors", but open to comments!

All other files are under the CCPL (as before), either copyright "The CC: Tweaked developers" (when this code does not stem from Dan) and "Daniel Ratcliffe" otherwise.

These headers have been updated on a best-effort basis and SHOULD NOT be taken as accurate at this stage. There's about 1k files in this changelog, so while I have been careful, no doubt some things will have slipped through the gaps.

---

My plan is to let this sit for two weeks[^1] to give me time to do another audit of these changes, and hopefully gather some further comments here or in #1339. Again, I'm no expert in software licensing, so thoughts welcome!

[^1]: All going well, planning to merge this on 16th March. I'm giving an exact date here mostly to force myself not to rush this through.

[REUSE]: https://reuse.software/


SquidDev:
I was on planning on merging this tomorrow, but it's the 16th in half the world and I think it's funny to merge this on the Ides of March. Off to the senate to assassinate the CCPL, yada yada.

PR #1352, opened 2023-03-01
Add certificate workaround for Let's Encrypt on 1.16

MCJack123:
This PR adds some code to automatically insert the Let's Encrypt root CA into CC's certificate trust store on 1.16. Due to some intricacies with how Netty loads trust stores, I had to do a lot of weird class stuff, though I guess that's par for the course when it comes to Java. This should allow CC to finally be able to connect to LE websites without updating Java. I have tested this in the IDE using Java 8u51 as supplied by Minecraft, and can confirm it's able to access both non-LE and LE sites.

I understand 1.16 is effectively EoL at this point, but it would be great to slide this in before it's fully dead. I'm happy to port this to 1.12 too if you'd be willing to push a final-final update for it.

image

### DIFF
--- a/src/main/java/dan200/computercraft/core/apis/http/NetworkUtils.java
+++ b/src/main/java/dan200/computercraft/core/apis/http/NetworkUtils.java
@@ -26,6 +26,7 @@ import io.netty.handler.traffic.GlobalTrafficShapingHandler;
 import javax.annotation.Nonnull;
 import javax.net.ssl.*;
 import java.io.ByteArrayInputStream;
+import java.lang.System;
 import java.net.InetSocketAddress;
 import java.net.URI;
 import java.security.KeyStore;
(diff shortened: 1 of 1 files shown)


SkyTheCodeMaster:
> 1.12 too if you'd be willing to push a final-final update for it.
Not trying to piggyback on this issue here, but wouldn't it be cool if we could also get `speaker.playAudio` into 1.12.2?

MCJack123:
> Not trying to piggyback on this issue here, but wouldn't it be cool if we could also get `speaker.playAudio` into 1.12.2?

No.

SquidDev:
Do we want to do the version sniffing Let'sEncrypt Craft does here and only install the certificate if we're pre u110? Or is that more of a hassle than it's worth?

MCJack123:
I did a very simple check, which only runs on any version matching `1.8.0_xx`, i.e. Java 8 under u100. This should be enough without making too much work.

SquidDev:
Thank you very much!

I feel like there must be a nicer way to manager the cert chain in Java (right, right????), but definitely don't think it's worth spending any more time on :).

MCJack123:
Yeah, I really wish there was a better way, but Java makes things convoluted in the hopes of supporting complex setups, and unfortunately Netty makes things even more complicated. A lot of this does stem from the fact that CC uses its own `SslContext` instead of whatever default one may exist.

First of all, Java doesn't expose the default `KeyStore` object [^1] at all outside the default `TrustManagerFactory` instance, and it cannot be extracted from the `TrustManager` object which owns it (without reflection, which is also unreliable). That means that to add a certificate, you need to create a new `TrustManager` that can first check the original object, and then fall back to the new certificate store if it fails.

But Java also makes this complicated by not allowing us to create a `TrustManager` directly from a `KeyStore`, and that object has required behavior for being embedded into our own `TrustManager`, so we can't use the `KeyStore` directly - instead, we have to use *another* `TrustManagerFactory` with our new key store, which will initialize the `TrustManager` we need, and then we can pull that out of the list.

Furthermore, Netty's `SslContextBuilder` wants an entire `TrustManagerFactory` object, which means that we can't just give it our merged `TrustManager` object directly - we also need to make a wrapper around it. Luckily, Java thought ahead and made `TrustManagerFactory` easy to extend... by using a *third* class which has the actual definitions. So to create the factory we need, we make a derived class which simply calls the `super` constructor with our implementation object. Then in the implementation, we pass through everything to the original `TrustManagerFactory` object, except that when getting the list of `TrustManager`s owned by the factory, we insert a new merged version with the original `TrustManager` + the one with the new certificate in place.

-----

...And all this to make a few more websites work. It's not the best code I've written, but at least it does what it's supposed to.

[^1]: From what I've heard, technically Java doesn't even *have* a default keystore - you're always supposed to supply one yourself. But it still has certificates when you use `TrustManagerFactory.init(null)`? And there's a whole `certtool` program for adding certificates globally via CLI??? ...Yeah, I don't quite understand this, but I don't expect any better from Java personally.

SquidDev:
Thank you for the explanation!

> A lot of this does stem from the fact that CC uses its own `SslContext` instead of whatever default one may exist.

Yeeeah, looking at the original code the whole `getTrustManager` is actually useless - I can just pass `null` to use the system default. Apparently I never read the docs when I first wrote this code????.

> Furthermore, Netty's `SslContextBuilder` wants an entire `TrustManagerFactory` object, which means that we can't just give it our merged `TrustManager` object directly - we also need to make a wrapper around it.

Yeah, this is solved in more recent versions of Netty (you can just pass a `TrustManager` directly), but alas not on the one that 1.16.5 is on :(.

> ...And all this to make a few more websites work. It's not the best code I've written, but at least it does what it's supposed to.

I think that's the crux of it really :). Thank you very much for working on this!

PR #1366, opened 2023-03-11
Fixed a typo in CableBlockEntity

OHakimen:
There was a typo in the CableBlockEntity class

Changed From:
```java
private static final String NBT_PERIPHERAL_ENABLED = "PeirpheralAccess";
```

to 

```java
private static final String NBT_PERIPHERAL_ENABLED = "PeripheralAccess";
```

### DIFF
--- a/projects/common/src/main/java/dan200/computercraft/shared/peripheral/modem/wired/CableBlockEntity.java
+++ b/projects/common/src/main/java/dan200/computercraft/shared/peripheral/modem/wired/CableBlockEntity.java
@@ -35,7 +35,7 @@ import javax.annotation.Nullable;
 import java.util.Collections;
 
 public class CableBlockEntity extends BlockEntity {
-    private static final String NBT_PERIPHERAL_ENABLED = "PeirpheralAccess";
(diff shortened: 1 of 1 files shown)


Lemmmy:
Wouldn't this be a breaking change? Would it be better to wait for a major version update and use a datafixer to fix this typo in existing saves?

https://github.com/cc-tweaked/CC-Tweaked/blob/1008333a846038d6fbf2682033a7684d1e657ea5/projects/common/src/main/java/dan200/computercraft/shared/peripheral/modem/wired/CableBlockEntity.java#L199-L211

OHakimen:
oh, true, that could potencially break stuff

SquidDev:
I don't think we need to do full-blown data fixer here. It might be sufficient to have `peripheralAccessAllowed = nbt.getBoolean(NBT_PERIPHERAL_ENABLED) || nbt.getBoolean("PeirpheralAccess");`?

SquidDev:
Fixed in 06322feb69838f194b47f1c73c316eb2e5581ac6.

PR #1376, opened 2023-03-14
Update to 1.19.4

SquidDev:
It runs, but I've not tested it very much.

 - [x] Re-enable Forge project: It's not updated yet, but should have by tomorrow.
 - [x] Update mod dependencies. There's a crash currently caused by running the 1.19.3 version of ModMenu, and I dare say some more issues.
 - [x] Fix the indentation in `SpeakerPeripheral`. What went on there?!


SquidDev:
Waiting on https://github.com/MinecraftForge/MinecraftForge/pull/9381 to be resolved before merging, as that's blocking some tests from passing.

SquidDev:
Our additional client tests (the Iris/Sodium ones) are non-functional right now due to some Loom 1.1 changes, but everything else should be good.

PR #1378, opened 2023-03-16
Add more explanation on how `os.epoch("ingame")` works

MCJack123:
This PR expands the documentation on what the value returned by `os.epoch("ingame")` means. I recently had to explain what it means to someone, only to realize that the docs for the function are pretty unclear as to what a millisecond actually is. The new docs clarify the meaning to explain that `os.epoch` generalizes milliseconds into a Minecraft day that has 86400 seconds, plus some quick conversion coefficients.


SquidDev:
Thanks!

There's a bit of me which just wants to deprecate and/or remove the ingame time for this function, it's such confusing behaviour. Maybe if there's ever a good time for breaking changes :D.

Issue #1196, opened 2022-10-30
NBTUtil item hashes changing after being pushed/inserted into containers


Gungadean:
### Minecraft Version

1.19.x

### Version

1.100.10

### Details

When the items are transferred to a different inventory via pipes, vanilla hoppers or CC's chest.pushItems() function, ordering of the item's NBT data can shifted changing the hash data. It only seems that some items are affected by this and it's not exclusively modded vs vanilla.  Currently using All The Mods 8 v0.1.8. I did some testing in All The Mods 7 v0.4.34 and there were no issues with NBT reordering. Although this appears to be an issue with either Forge or Minecraft's code with inserting items, with JSON being an unordered set of data, it may be worthwhile to sort the CompoundTag keys of an item before generating a hash as there is potential for other mods to potentially mess with the order of NBT.

Screenshots:
![Screenshot 2022-10-30 114700](https://user-images.githubusercontent.com/19896537/198888130-998862e7-ce32-469b-b287-53b373c252ea.png)
Above picture shows me using the command /advancedperipherals getItemHash command (which calls the CC function) and minecraft's /data entity @s SelectedItem command twice, before and after transfer. As can be seen in the picture, specifically the builder tag's coordinates get shuffled around when the item is transferred.

![Screenshot 2022-10-30 114719](https://user-images.githubusercontent.com/19896537/198888333-d62d7771-e130-4f09-a904-70451da7b1ae.png)
I also made a simple program to transfer items between and getting NBT data before and after the transfer. (code: https://pastebin.com/N6zvNRvR)

SquidDev:
Thanks for the report! I've pushed a commit to sort keys, as you mentioned.

Worth noting that there's never going to be a perfect one-to-one mapping between "identical" items and NBT hash[^1], but definitely agreed it shouldn't be changing mid-transfer!

[^1]: For instance, if two items have the same enchantments but in different orders, the hash will be different.

PR #1391, opened 2023-03-26
fix: sort compounds in NBT lists for hashing

Lemmmy:
A follow-up from #1196, which 14cb97cba106a4981f7990c792ad5f8b7405172b addressed by sorting keys in NBT compounds. The existing code would only traverse compounds themselves, so if it hit a list containing compounds, those compounds would not get sorted.  

We ran into this on SwitchCraft with shulker boxes. I'm still not entirely sure how these compounds are ending up being sorted differently in the first place, I'm sure they're always using ContainerHelper when saving, but as of this change at least they now produce consistent hashes.


SquidDev:
Thank you!

PR #1414, opened 2023-04-15
fix: on/off translation inverted

### LINKED ISSUE #85: Computer power button is inverted
### Describe the bug

When a computer is on, the power button shows as if it is off, and the tooltip on it is "Turn this computer on", but when a computer is off, it shows as if it is on (with the tooltip for a computer that is on).
Here are two screenshots below about it:
Computer that is turned on
Computer that is turned off

### To reproduce

Simply go into a computer on the server. You will see that the power button is inverted.

### Expected behavior

When the computer is off, the power buttons shows that it is off, instead of showing like it is on, and vica versa.

### Relevant log output

_No response_

### Please confirm

- [X] This is not an exploit

### LINKED ISSUE #1413: Computer power button is inverted
### Minecraft Version

1.19.x

### Version

1.104.0

## The problem:
When a computer is on, the power button shows as if it is off, and the tooltip on it is "Turn this computer on", but when a computer is off, it shows as if it is on (with the tooltip for a computer that is on).
Here are two screenshots below about it:
kép
kép
## To reproduce:
Simply go into a computer on the server. You will see that the power button is inverted.
## Expected behavior:
When the computer is off, the power buttons shows that it is off, instead of showing like it is on, and vica versa.

piprett:
Fixes #1413
Fixes SwitchCraftCC/issues#85

Makes the computer power buttons correct

### DIFF
--- a/projects/common/src/client/java/dan200/computercraft/client/gui/widgets/ComputerSidebar.java
+++ b/projects/common/src/client/java/dan200/computercraft/client/gui/widgets/ComputerSidebar.java
@@ -45,11 +45,11 @@ public final class ComputerSidebar {
         x += CORNERS_BORDER + 1;
         y += CORNERS_BORDER + ICON_MARGIN;
 
-        var turnOn = new HintedMessage(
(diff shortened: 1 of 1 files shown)


SquidDev:
Thank you!

PR #1417, opened 2023-04-20
Make maximum upload file size configurable

khankul:
(This is my first time working on a project this large, so any feedback is much appreciated)

Rationale: provide players and server administrators with more control over the mod


EmmaKnijn:
I'd love to have this feature! I can't to QC tho lol

projects/common/src/main/java/dan200/computercraft/shared/config/ConfigSpec.java
```diff
@@ -94,6 +95,10 @@ private ConfigSpec() {
                 .comment("The disk space limit for floppy disks, in bytes.")
                 .define("floppy_space_limit", Config.floppySpaceLimit);
 
+            uploadMaxSize = builder
+                .comment("The file upload size limit, in bytes.")
+                .define("upload_max_size", Config.uploadMaxSize);
```
SquidDev: Can we use `defineInRange` here instead. Probably with a minimum size of 1KiB and maximum of ... idk ... 16MiB?

SquidDev (changes requested):
Thank you for the PR!

One thing to be aware of (which I think is worth documenting in the config's comment) is that file uploads send the whole contents at once[^1]. This means that if you're trying to upload large files (or are on poor internet), uploads could stall the networking thread for several seconds.

[^1]: Well, spread over multiple packets, but still in a single tick.

SquidDev:
> I can't to QC tho lol

I can't to Quality control tho lol? Questionable Content? Queen's Counsel?

EmmaKnijn:
Yes, Quality Control

SquidDev:
Oh, sorry, it's only just occurred to me that this won't work on Fabric servers, as config isn't synced between the client and the server.

We probably need to pass this option to the client in `ComputerContainerData`/`AbstractComputerMenu`, and then read that inside `AbstractComputerScreen`. It should be fairly easy, but might require some extra work to make it not incredibly ugly. Not sure, haven't put too much thought into it.

khankul:
Oh my. I wish I had any idea about what you're describing.  I've little to no idea how this whole system works. Could you please give me some guidance?

khankul:
Spent a whole day on this, couldn't crack it. Sorry.

SquidDev:
Ahh, sorry! So in CC:T, GUIs are basically split into three classes:

 - Menus (so in this case `AbstractComputerMenu`), which stores the current state of the opened UI (such as inventory slots, the current computer's terminal, etc...). This is created on both the client and server, and synced between the two.
 - Screens (`AbstractComputerScreen` here) are responsible for implementing the actual UI on the client (rendering, handling keyboard/mouse input). It has access to the client-side menu, so reads most of the GUIs state from there.
 - `ContainerData` (`ComputerContainerData` here) stores the initial state of the opened GUI, and is used to construct the menu on the client side. In the case of `ComputerContainerData`, this contains the current terminal contents, and a couple of other bits and bobs.

   `ContainerData` implementations write their state to a `FriendlyByteBuf` on the server, and then read it on the client, allowing it to be sent over the network.

The main problem here is that `Config` fields aren't updated on the client side, as the config is only read on the server. Thus we need to make sure the correct value is available on the client.

The easiest way to do this is send it to the client when opening the container. So `ContainerData` stores the current max size, sends it to the client, and then `AbstractComputerMenu` reads it from the provided data. We then just need a getter on the menu, so that  the screen can read that config option.

khankul:
~~Thank you, unfortunately I still can't wrap my head around it. `AbstractComputerMenu` doesn't seem to be reading from any buffers. How do I use buffers in `ComputerContainerData`? I'm confused about where read and write occur.~~

EDIT: Wait, I think I got it.

khankul:
Tested it on a local Fabric server. Wonderful.

khankul:
Please let me know if there are any more changes you want to request.

SquidDev (approved):

SquidDev:
Thank you very much, and apologies for the slow review!

PR #1446, opened 2023-05-17
fix #1367: Add `colors.fromBlit`

### LINKED ISSUE #1367: Add `colors.fromBlit` to the `colors` API
So we currently have `colors.toBlit`, which is useful for converting a `color` value to be used with `Redirect.blit`. However, extracting that same color from a `window` with `window.getLine`, is not feasible in the standard API. What I propose is the following:

# colors.fromBlit(blitColors[, index=1])

Converts the given paint/blit hex character (0-9a-f) to a color.

## Parameters

1. `blitColors : string` The given paint/blit hex character(s).
2. `index? : number = 1` The index of the paint/blit hex character to convert.

## Returns

1. `number` The color value of the given paint/blit hex character.

## Throws

- If the character given is not (0-9a-f).

piprett:
Issue: fixes #1367 
Adds a `fromBlit` function in the `colors` API. I've added some tests, which are passing. I have ignored the index argument, as I am unsure how I should implement it. I am also returning `nil` instead of throwing, after talking with SquidDev.

### DIFF
--- a/projects/core/src/main/resources/data/computercraft/lua/rom/apis/colors.lua
+++ b/projects/core/src/main/resources/data/computercraft/lua/rom/apis/colors.lua
@@ -351,39 +351,40 @@ for i = 0, 15 do
     color_hex_lookup[2 ^ i] = string.format("%x", i)
 end
 
---- Converts the given color to a paint/blit hex character (0-9a-f).
---
--- This is equivalent to converting floor(log_2(color)) to hexadecimal.
---
--- @tparam number color The color to convert.
--- @treturn string The blit hex code of the color.
--- @usage
--- ```lua
--- colors.toBlit(colors.red)
--- -- => "c"
--- ```
--- @see colors.fromBlit
--- @since 1.94.0
+--[[- Converts the given color to a paint/blit hex character (0-9a-f).
+
+This is equivalent to converting floor(log_2(color)) to hexadecimal.
+
+@tparam number color The color to convert.
+@treturn string The blit hex code of the color.
+@usage
+```lua
+colors.toBlit(colors.red)
+-- => "c"
+```
+@see colors.fromBlit
+@since 1.94.0
+]]
 function toBlit(color)
     expect(1, color, "number")
     return color_hex_lookup[color] or
         string.format("%x", math.floor(math.log(color) / math.log(2)))
 end
 
---- Converts the given paint/blit hex character (0-9a-f) to a color.
---
--- This is equivalent to converting the hex character to a number and then 2 ^ decimal
---
--- @tparam string hex The paint/blit hex character to convert
--- @treturn number The color
--- @usage
--- ```lua
--- colors.fromBlit("e")
--- -- => 16384
--- ```
--- @see colors.toBlit
--- @since 1.105.0
+--[[- Converts the given paint/blit hex character (0-9a-f) to a color.
 
+This is equivalent to converting the hex character to a number and then 2 ^ decimal
+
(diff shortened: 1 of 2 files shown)


projects/core/src/main/resources/data/computercraft/lua/rom/apis/colors.lua
```diff
@@ -363,3 +363,20 @@ function toBlit(color)
...
         string.format("%x", math.floor(math.log(color) / math.log(2)))
 end
+
+--- Converts the given paint/blit hex character (0-9a-f) to a color.
+--
+-- This is equivalent to converting the hex character to decimal and then 2 ^ decimal
+--
+-- @tparam string hex The paint/blit hex character to convert
+-- @treturn number The color
+
```
SquidDev, line 377: Can you add the version here too?

```suggestion
-- @since 1.105.0
```

Might be nice to have an example here (and links to/from `colours.toBlit`?) too, but I think that's less important.

projects/core/src/test/resources/test-rom/spec/apis/colors_spec.lua
```diff
@@ -93,4 +93,18 @@ describe("The colors library", function()
...
+    describe("colors.fromBlit", function()
+        it("validates arguments", function()
+            expect.error(colors.fromBlit, nil):eq("bad argument #1 (string expected, got nil)")
+            expect(colors.fromBlit("")):eq(nil)
+            expect(colors.fromBlit("not hex")):eq(nil)
+        end)
+
+        it("converts all colors", function()
+            for i = 0, 15 do
+                expect(colors.fromBlit(string.format("%x", i))):eq(2 ^ i)
```
SquidDev: Could we use `colours.toBlit` here instead of `string.format`? Would be nice to show they're an inverse of each other.

projects/core/src/main/resources/data/computercraft/lua/rom/apis/colors.lua
```diff
@@ -363,3 +363,20 @@ function toBlit(color)
...
+function fromBlit(hex)
+    expect(1, hex, "string")
+
+    if hex:find("[^0-9a-f]") or hex == "" then
+        return
+    end
+
+    return math.floor(2 ^ tonumber(hex, 16))
```
SquidDev: I wonder if we could remove the `hex:find` here, and just rely on `tonumber` failing

Could we do something like this instead:

```lua
if #hex ~= 1 then return nil end

local value = tonumber(hex, 16)
if not value then return nil end

return 2 ^ value
```

I'm not sure if/why we need the `math.floor` is needed here.

projects/core/src/main/resources/data/computercraft/lua/rom/apis/colors.lua
```diff
@@ -363,3 +363,20 @@ function toBlit(color)
     return color_hex_lookup[color] or
         string.format("%x", math.floor(math.log(color) / math.log(2)))
 end
+
+--- Converts the given paint/blit hex character (0-9a-f) to a color.
+--
+-- This is equivalent to converting the hex character to decimal and then 2 ^ decimal
```
SquidDev:
```suggestion
-- This is equivalent to converting the hex character to a number and then 2 ^ decimal
```

I think this is clearer. After all, numbers are represented as binary, not decimal!

SquidDev (changes requested):
Thank you! (and special thanks for adding tests!)

Just a couple of comments, but otherwise looks nice!

SquidDev (approved):

Issue #1373, opened 2023-03-14
The big ol' compedium of flaky tests


SquidDev:
CC:T's test suite is incredibly flaky when run via CI. I think this is just a limitation of how powerful the runners are (they work fine on my machine!) - but having to rerun every third build gets old quickly.

This issue is an attempt to track recent tests failures.

#### Builds
 - Build 2243: https://github.com/cc-tweaked/CC-Tweaked/actions/runs/4413979109/jobs/7735111084
 - Build 2257: https://github.com/cc-tweaked/CC-Tweaked/actions/runs/4430925396/jobs/7773273120, https://github.com/cc-tweaked/CC-Tweaked/actions/runs/4430925396/jobs/7773650001 (only some failed here)
 - Build 2271: https://github.com/cc-tweaked/CC-Tweaked/actions/runs/4525631854/jobs/7970262900

#### Notes
In 2243, the test batch was created at 09:21:22 and finally started running at 09:21:25 (this is 3s for a single tick, which is Bad(TM)). Computers continue to turn on until 09:21:28, and then tests time out at 09:21:29 (after 150 ticks). There's two odd things here:
 - Computers really shouldn't take 3 seconds to all turn on - we're using the fake Kotlin machine here, which has very little startup overhead.
 - Tests time out after 150 ticks, which _should_ take 7.5s, but only appears to be 4s. Unless Minecraft is skipping ticks here or something - maybe we should add some logging here just for debugging?

### `modem_test.have_peripherals.modem_test.have_peripherals`
```
Starts with peripherals ==> 
 expected: <[monitor_0, printer_0, right]> 
  but was: <[right]> at 4,-58,22 (relative: 4,2,2) (t=1)
```

#### Builds
 - Build 2257: https://github.com/cc-tweaked/CC-Tweaked/actions/runs/4430925396/jobs/7773927203
 - Build 2268: https://github.com/cc-tweaked/CC-Tweaked/actions/runs/4515339240/jobs/7952515237

### `inventory_test.fails_on_full.inventory_test.fails_on_full`

```
No peripheral: minecraft:furnace_0 at 47,-58,11 (relative: 2,2,1) (t=1)
```

#### Builds
 - Build 2268: https://github.com/cc-tweaked/CC-Tweaked/actions/runs/4515339240/jobs/7952515237

### `modem_test.full_modems_form_networks.modem_test.full_modems_form_networks`
```
[...]

PR #1457, opened 2023-05-29
Creating FabricComputeCraftAPI with turtle/pocket upgrade serializer registration methods

SirEdvin:
So, this is continuation of #1432, specific turtle and pocket upgrade serializer API for registration, that will help get rid of postpone registration.

For now, due to postpone registration I am not able to have links to registered serializers in minecraft mod code and postpone registration a significant mess right now.

### DIFF
--- a/projects/fabric-api/src/main/java/dan200/computercraft/api/FabricComputerCraftAPI.java
+++ b/projects/fabric-api/src/main/java/dan200/computercraft/api/FabricComputerCraftAPI.java
@@ -1,0 +1,44 @@
+// SPDX-FileCopyrightText: 2022 The CC: Tweaked Developers
+//
+// SPDX-License-Identifier: MPL-2.0
+
+package dan200.computercraft.api;
+
+import dan200.computercraft.api.pocket.IPocketUpgrade;
+import dan200.computercraft.api.pocket.PocketUpgradeSerialiser;
+import dan200.computercraft.api.turtle.ITurtleUpgrade;
+import dan200.computercraft.api.turtle.TurtleUpgradeSerialiser;
+import dan200.computercraft.impl.ComputerCraftAPIFabricService;
+import net.minecraft.resources.ResourceLocation;
+
+/**
+ * The fabric-specific entrypoint for ComputerCraft's API.
+ */
+public final class FabricComputerCraftAPI {
+    private static ComputerCraftAPIFabricService getInstance() {
+        return ComputerCraftAPIFabricService.get();
+    }
+
+    /**
+     * Registers a turtle upgrade serializer.
+     *
+     * @param key Turtle upgrade serializer ID.
+     * @param serializer Turtle upgrade serializer for registration
+     * @see ITurtleUpgrade
+     * @see TurtleUpgradeSerialiser
+     */
+    public <T extends ITurtleUpgrade> TurtleUpgradeSerialiser<T> registerTurtleSerializer(ResourceLocation key, TurtleUpgradeSerialiser<T> serializer) {
+        return getInstance().registerTurtleSerializer(key, serializer);
+    }
+    /**
+     * Registers a pocket upgrade serializer.
+     *
+     * @param key Pocket upgrade serializer ID.
+     * @param serializer Pocket upgrade serializer for registration
+     * @see IPocketUpgrade
+     * @see PocketUpgradeSerialiser
+     */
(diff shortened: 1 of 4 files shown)


SirEdvin:
So, locally test passing, I am not sure why they failing in CI, probably related to #1373 or I just running test incorrect locally

SquidDev:
Thank you for looking at this! We hit this in Plethora a while back, and [I bodged around it](https://github.com/SwitchCraftCC/Plethora-Fabric/blob/91a64b3cf9f428227425e06bbbc8aa6e9a416bee/build.gradle.kts#L171) and forgot to look into handling this properly. CCing @Lemmmy here, as this change will probably affect you too.

So the original intent with the API was that by just exposing a `ResourceKey`, it would be easy to use with any registry system (vanilla's, Forge's `DeferredRegistry`, whatever multi-loader abstraction people are using).

I do feel a little eeeeeh about this change, as now it means the API is now different between mod loaders. There are some possible alternatives I can think of:

 - Replace `TurtleUpgradeSerialiser.REGISTRY_ID` with a `TurtleUpgradeSerialiser.registryId()` method. This behaves the same as the field access, but ensures the registry exists. I feel a little uneasy about this, as it means you're performing side effects in class initialisers (or getters), but plenty of other mods do that, so it's probably not the end of the world.

 - Add a custom Fabric entry point that other mods can hook into (much like JEI or modmenu do). 
This will be run during CC:T's init, so guarantees the registry exists.

   This could optionally be used for registering all CC-related functionality (so mods don't need to do `if(isModLoaded("computercraft")) registerComputerCraft()`), but that's less relevant for dedicated CC add-on mods :).

I don't really have any strong thoughts here - I don't maintain any CC add-ons after all, so happy to go with whatever folks think will be easiest, be that this PR or one of the other options!

[^1]: I actually thought it used a topo-sort in production, but apparently I misremembered. Moot, as in dev the order is random.

SirEdvin:
I am not sure, if I did this correct or not, but even now registration for serializers is a little different between platforms, because as I understand, common approach for use registries in forge is to create mod-specific ones, like [here](https://github.com/SirEdvin/Turtlematic/blob/1.19.4/projects/forge/src/main/kotlin/site/siredvin/turtlematic/ForgeTurtlematic.kt#L29), so I personally don't see any issue here.

So I still using reference for `TurtleUpgradeSerialiser.REGISTRY_ID`, but for multiloading env code is pretty much different and requires some xplat proxy wrapper.

SquidDev:
Yeah, I think that depends on how you deal with abstracting over registries in general. In CC:T, I use a `DeferredRegistry`-style pattern for  both loaders, so turtle serialisers are just registered the same as any other block/item:

https://github.com/cc-tweaked/CC-Tweaked/blob/e153839a9824d858347a91beb20b6906d061ead7/projects/common/src/main/java/dan200/computercraft/shared/ModRegistry.java#L257-L270

SirEdvin:
I suppose, that for exposing `registry()` a lot of cross-platform internal code of CC:T need to be exposed as part of API, which doesn't seem to me as a good solution. Also, still, for Forge version, modder will still use `DefferedRegistry` not `registry()` from Serializer. 

But for being able to mimic DefferedRegistry registry API pattern in common code I can suggest to make `TURTLE_UPGRADE_SERIALISERS` and `POCKET_UPGRADE_SERIALISERS` accessible from FabricAPI, so modder can decide which approach they what to use - registration functions or mimic registry wrapping with some extra logic. 
In this case, registration function can be just removed.

SquidDev:
> Add a custom Fabric entry point that other mods can hook into (much like JEI or modmenu do).
This will be run during CC:T's init, so guarantees the registry exists.

I had a look at this, but I'm not convinced about this model. This means then you're calling the CC entrypoint before some mods have been initialised. This is is _probably_ fine (most Fabric mods do half their work in static initialisers), but feels a little ugly too.

It is a bit sad really. I feel Fabric normally has a nicer API design than Forge, but I do prefer Forge's explicitness here.

> Replace `TurtleUpgradeSerialiser.REGISTRY_ID` with a `TurtleUpgradeSerialiser.registryId()` method. This behaves the same as the field access, but ensures the registry exists.

This is the approach I ended up going with in f0abb83f6e49c2f63ea26792e0063df7674f5712. This should mean you can replace `REGISTRY_ID` with `.registryId()` and things will Just Work:tm:.

SirEdvin:
Ah, this what you mean by `registryId()` function. Again my reading issues. 
Thanks for resolving this!

SquidDev:
Yeah, I don't think I explained that very well! I had a very clear image of what it'd look like in my head, but then explained the implementation rather than how it'd actually help.

PR #1461, opened 2023-05-30
HTTP proxy (closes #369)

### LINKED ISSUE #369: global network proxy
sometimes such as `raw.githubusercontent.com`, can't be connected in China, many install scripts use this
If computer have commands like `setproxy http://127.0.0.1:1080` or `setproxy socks://127.0.0.1:1080`, it will be useful when connect github and download scripts.

图片

Lemmmy:
This PR implements in-mod tunnelling for HTTP and websocket connections. This provides an alternative to applying a proxy to the entire JVM, which may not always be desirable. The implementation is loosely based on what SwitchCraft is using, though with less bells and whistles (no proxy pooling, no healthchecks, no failover, and no strict enforcement).

New settings were added to `computercraft-server.toml`. Globally, there is one proxy configuration in `http.proxy`:

```toml
	#Tunnels HTTP and websocket requests through a proxy server. Only affects HTTP
	#rules with "use_proxy" set to true (off by default).
	#If authentication is required for the proxy, create a file in the same
	#directory as "computercraft-server.toml" called "computercraft-proxy.pw",
	#containing the username and password separated by a colon, e.g.
	#"myuser:mypassword". For SOCKS4 proxies only the username is required.
	[http.proxy]
		#The type of proxy to use.
		#Allowed Values: HTTP, HTTPS, SOCKS4, SOCKS5
		type = "HTTP"
		#The hostname or IP address of the proxy server.
		host = ""
		#The port of the proxy server.
		#Range: 1 ~ 65536
		port = 8080
```

And per-rule, there is a `use_proxy` boolean, which defaults to false:

```toml
	[[http.rules]]
		host = "*"
		#Enable use of the HTTP/SOCKS proxy if it is configured.
		use_proxy = false
```

Finally, if the proxy requires basic authentication, a separate file must be made in `world/serverconfig/computercraft-proxy.pw`. The file must be of the form `username:password`, the format of an unhashed htpasswd file, such as a Prometheus password file. It is kept separate since Forge may sync the config to the client.

HTTP, HTTPS, SOCKS4 and SOCKS5 proxies are supported. All four support authentication, though SOCKS4 does not support password authentication (per protocol). In this case the `computercraft-proxy.pw` file should just contain the username. I have not fully tested HTTPS proxies, but I believe they are only appropriate if the proxy server has a TLS certificate for the hostname you connect to it with (e.g. more for corporate proxies than for free open proxies).

Two new dependencies were added: netty-codec-socks and netty-handler-proxy. Hopefully I've added them in all the correct places.

Note that this does not proxy DNS in any way. `http.checkURL` and the DNS lookups involved in initiating requests (including to the proxy) will still be done by the system/JVM.

### TODO
- I'm not entirely happy with how the password file is loaded. I needed access to the server folder or an existing config path, and put it near where each loader already loads configuration. It is also not reloadable.
- ~~Currently `http.proxy.type` is completely ignored on Fabric - something seems to be wrong with correcting `defineEnum` in the config spec. This affects `monitor_renderer` in the client config too. The value appears in the config spec itself, but gets completely skipped over when validating.~~ Fixed by d3b39be
- Does this functionality require documentation in `doc/guides`?

Fixes #369.


projects/forge/src/main/java/dan200/computercraft/ComputerCraft.java
```diff
@@ -95,6 +96,7 @@ private static void syncConfig(ModConfig config) {
         if (!config.getModId().equals(ComputerCraftAPI.MOD_ID)) return;
 
         if (config.getType() == ModConfig.Type.SERVER && ((ForgeConfigFile) ConfigSpec.serverSpec).spec().isLoaded()) {
+            ProxyPasswordConfig.init(config.getFullPath().resolveSibling(ComputerCraftAPI.MOD_ID + "-proxy.pw"));
```
SquidDev: Could we pass the config path into `ConfigSpec.syncServer`, and then call ` ProxyPasswordConfig.init` in there instead?

Also, do you know what happens when a client connects to a dedicated server here? I fear this will crash, as the path is only available on the server. I wonder if we need something like `config.getConfigData() instanceof FileConfig fileConfig ? fileConfig.getNioPath() : null`.

projects/core/src/main/java/dan200/computercraft/core/apis/http/NetworkUtils.java
```diff
@@ -150,19 +209,27 @@ public static Options getOptions(String host, InetSocketAddress address) throws
...
      */
-    public static void initChannel(SocketChannel ch, URI uri, InetSocketAddress socketAddress, @Nullable SslContext sslContext, int timeout) {
+    public static void initChannel(SocketChannel ch, URI uri, InetSocketAddress socketAddress, @Nullable SslContext sslContext, @Nullable ProxyHandler proxy, int timeout) {
         if (timeout > 0) ch.config().setConnectTimeoutMillis(timeout);
 
         var p = ch.pipeline();
         p.addLast(SHAPING_HANDLER);
 
+        if (proxy != null) {
+            if (CoreConfig.httpProxyType == ProxyType.HTTPS && sslContext != null) {
```
SquidDev: `sslContext` will be null when connecting to non-secure/HTTP servers, which is probably not what we want here.

I wonder if we could make `getProxyHandler` return a `Consumer<ChannelPipeline>` instead (where each lambda is just `p -> p.addList(...)`). Then in the HTTPS case we fetch `getSslContext` outside the lambda, and set up the pipeline inside. IDK, I can't think of a cleaner way, but also appreciate it's not very nice.

SquidDev (changes requested):
:heart:

projects/core/src/main/java/dan200/computercraft/core/apis/http/NetworkUtils.java
Lemmmy: Hopefully I've understood correctly and implemented this properly in 1f5476fa69f2d2689ce828ad28d7c42cd52edd5c. The SSL context is now added to the pipeline inside the lambda in `getProxyHandler`, but it also needs the `SocketChannel` and `timeout` for this.

SquidDev (approved):
Thank you!

I'd quite like to write some tests for this, but it looks like actually implementing a HTTP proxy from Netty is quite a pain :/. Can probably wait for now.

Issue #903, opened 2021-08-23
Improved interface for storing upgrade data


SquidDev:
This is effectively copied from my comments in #873. This issue attempts to address its needs and those in #847:

 - Capabilities on turtles. Aside, do we want capabilities on turtle item stacks too? It might be cool if turtles could, for instance, act as tanks in item form.
 - Persisting data when turtles are broken/placed (#847).
 - Handling additional NBT data on the upgrade `ItemStack` (we also need to handle doing this in a crafting grid).

I think the first two of these are relatively easy to solve in a cohesive manner by adding the following:

 - `interface TurtleUpgradeData extends ICapabilityProvider, INBTSerializable<Tag> { }`
 - `public TurtleUpgradeData ITurtleUpgrade.createData(ITurtleAccess, TurtleSide)`.
 - `public TurtleUpgradeData ITurtleAccess.getUpgradeData(TurtleSide)`.

I think we probably also need a generic `ITurtleAccess.setChanged()` method which just calls `TileEntity.setChanged()`.

Basically upgrades can define a class which implements `TurtleUpgradeData` and stores data about their upgrade (such as a fluid tank's contents or a list of waypoints). Those can then be persisted via `INBTSerializable<Tag>` on both the turtle block entity and item stack.

`createData` will be called before `createPeripheral`, so you can then access the constructed object when constructing your peripheral, rather than accessing it every method call.

`ICapabilityProvider` also allows this interface to expose capabilities if it wants, though those are not required.

As far as saving/reading to the upgrade item, perhaps we add two additional methods to `TurtleUpgradeData` which are basically `serializeToStack(ItemStack)` (or `ItemStack serializeToStack()`) and `deserializeFromStack(ItemStack)`?

## Example

```java
class FluidData implements TurtleUpgradeData {
  private final ITurtleAccess turtle;
  private final FluidTank current = new FluidTank();

  @Override
  public deserializeFromStack(ItemStack stack) {
    current.readFromNBT(stack.getCompoundTag().get(FLUID_TANK_TAG)); // TODO: more error checking please
  }

  // TODO: Storage and capabilities
}

class FluidUpgrade implements ITurtleUpgrade {
  @Override
  public TurtleUpgradeData createUpgradeData(ITurtleAccess turtle, TurtleSide side) {
    return new FluidData(turtle);
  }

  @Override
  public IPeripheral createPeripheral(ITurtleAccess turtle, TurtleSide side) {
[...]

PR #1465, opened 2023-06-03
Allow passing NBT information from and to item stacks

### LINKED ISSUE #847: Turtle NBT upgrade data are loss when you break turtle

SirEdvin:
So, this it mostly exists to partially implement #903 and fix #847. For now the goal of this PR is mostly discussing API how it should be implemented, because #903  design for more complex, but I guess this design was mostly about capabilities. Due to multi-load version of current mod, capabilities idea is obsolete right now, so I suggest to focus on NBT data passing from and to item stacks, that correspond updates. 

How is this NBT data can be used:

- Creating new turtle tools upgrades, that will use enchantments
- Storing important stuff inside turtle upgrade, to not lose it, for example, [experience](https://docs.siredvin.site/API/experience/), fluids, [upgrade configurations](https://docs.siredvin.site/API/warp/), [captures entities](https://docs.siredvin.site/API/capture/), consumed items by upgrade, like [another upgrades](https://docs.siredvin.site/UnlimitedPeripheralWorks/peripherals/peripheralium_hub/)

TODO list:
- [x] Make sure that turtle upgrade information passes when turtle destroys (I don't remember correctly, but there is wasn't a case or just my memory isn't correct)
- [x] Introduce render method, that would relay only on `CompoundTag data` parameter, rather then ITurtleAccess, to provide ability to render tweaks inside inventory
- [x] Add some test item and some tests for getting/settings upgrade for turtles.


SirEdvin:
So, I guess now it is ready for at least first look. 

Main gap in implementation is probably an assumption, that all upgrades need to have a default NBT values, which should be empty, but I guess this is not so bad.

SirEndii:
*cough* gonna upvote this *cough*

SquidDev (commented):
Thank you for looking into this, and apologies for the delay!

Rather than passing around separate `ITurtleUpgrade` and `CompoundTag`s everywhere, I think it probably makes more sense to introduce a `record UpgradeData<T>(T upgrade, CompoundTag data)` class. Open to alternative names! Only other thing I can think of is `UpgradeInstance`, which IMO is even worse!

Then we can do the following:
 - `TurtleItem.create`/`PocketItem.create` accept an `UpgradeData<_>` instead of separate `IXyzUpgrade` and `CompoundTag`s.
 - `UpgradeManager.get(ItemStack)` returns an `UpgradeData<T>` instead of a raw `T`. 
 - Add a `ITurtleUpgrade.setUpgrade(TurtleSide, @Nullable UpgradeData<ITurtleSide>)`. Possibly also want a getter here.

I think this will make code a little cleaner, as this gives us a guarantee that either both upgrade and data are set, or neither are.

Some other thoughts here:
 - We probably want to expose the current upgrade data to the upgrade modeller. I think adding an `TransformedModel getModel(T upgrade, CompoundTag data, TurtleSide side);` overload would work, which defaults to calling `getModel(upgrade, (ITurtleAccess)null, side)`.

   It might make sense to deprecate the `ITurtleAccess` version, I don't know what sort of thing people are doing in their modellers.

 - Some upgrades use turtle data for sending state to the client (for instance modems use it for sending the on/off state). We don't want to save this data when breaking the turtle - it doesn't make much sense for a turtle to have an "on" modem when in item form!

   In #903 I proposed making this an entirely separate concept. That's probably too complex a system, maybe there just needs to be a `CompoundTag idkwhattocallthis(CompoundTag)` (or `CompoundTag idk(ITurtleAccess, TurtleSide)`) which is invoked when creating the item.

SirEdvin:
Okey. I will try to change this behavior. 

About filter storing information after breaking, I can suggest only another methods for ITurtleUpgrade that process breaking and placing turtle.

So we will have four methods:
```java
default @Nonnull ItemStack produceCraftingItem(@Nonnull CompoundTag upgradeData) {
      return getCraftingItem();
  }

  default @Nonnull CompoundTag produceUpgradeData(@Nonnull ItemStack stack) {
      return new CompoundTag()
  }

default @Nonnull CompoundTag load(@Nonnull CompoundTag data) {
  return new CompoundTag()
}

default @Nonnull CompoundTag save(@Nonnull CompoundTag data) {
  return new CompoundTag()
}
```

I am not sure, will it be CompoundTag or UpgradeData, I will probably check if wrapping will be required in most cases, it will be UpgradeData, or I will just add this method for UpgradeData as proxy methods. 

wdyt?

SquidDev:
I'm not sure `save`/`load` make much sense, as we need the item-form and BlockEntity-form NBT to be consistent, as we could call `produceCraftingItem`/`produceUpgradeData` on both. Instead I think let's do the following:

```java
default ItemStack getUpgradeItem(CompoundTag upgradeData) { return getCraftingItem(); }
default CompoundTag getUpgradeData(ItemStack craftingStack) { return new CompoundTag(); }
default CompoundTag getPersistedData(CompoundTag upgradeData) { return upgradeData; }
```

Does that sound sensible?

SirEdvin:
Yep, only one suggestion about `getPersistedData`, I propose this method signature:

```java
default @Nullable CompoundTag getPersistedData(CompoundTag upgradeData) { return null; }
```

To be backward compatible and allow most of upgrades just ignore persisted data if they don't need it. Or instead of `Nullable` you would prefer empty CompoundTag?

SquidDev:
Probably an empty tag. Let's avoid null where possible. I think as long as we make this clear that classes may need to override this method in the javadoc it's probably fine :).

SirEdvin:
And ... partially done.
I have pushed I guess, most if not all changes :) 

Returning to `getModel(T upgrade, CompoundTag data, TurtleSide side);` I personally not sure if this is good idea or not, because I can imagine using different information about turtle used for rendering, especially information about another upgrade in case, when you place upgrade on top of turtle, like @JheffersonMarques with Solar Turtle, this can help avoid rendering solar panel on turtle two times, which probably can lead to graphical glitches. 


So, my next goal with this PR actually implement some netherite sword peripheral using all of this methods and test it finally :)

SirEdvin:
I will probably would need to review code once more after several days, because seems two desings overlap  and now we have a strange inconsistency between `setUpgrade` but `getUpgradeData`, probably method names should be consistent :)

SquidDev:
> `Returning to getModel(T upgrade, CompoundTag data, TurtleSide side);`  I personally not sure if this is good idea or not, 

So we want a way to change the model based on the current upgrade data when `ITurtleAccess` is not available (i.e when rendering as an item). Though yes, deprecating the `ITurtleAccess` version probably isn't needed - though maybe next breaking change we can shuffle those methods around a bit.

Will need to take a look at how the turtle item model cache works though, as we'll probably need some eviction there.

SquidDev:
We're going to need to be careful here about how `CompoundTag`s are passed around - we probably need to do some defensive copying in a couple of places to make sure we don't accidentally share tags between multiple items.

projects/common/src/main/java/dan200/computercraft/shared/pocket/core/PocketServerComputer.java
```diff
@@ -112,20 +113,28 @@ public Map<ResourceLocation, IPeripheral> getUpgrades() {
...
     /**
      * Set the upgrade for this pocket computer, also updating the item stack.
      * <p>
      * Note this method is not thread safe - it must be called from the server thread.
      *
      * @param upgrade The new upgrade to set it to, may be {@code null}.
      */
-    public void setUpgrade(@Nullable IPocketUpgrade upgrade) {
-        if (this.upgrade == upgrade) return;
+    public void setUpgrade(@Nullable UpgradeData<IPocketUpgrade> upgrade) {
```
SirEdvin, line 124: So, just for context. This method is not exposed as part of API, and because of this, it was replaced.

SirEdvin:
And another cleanup on the way, also with some start code for implementing test tool with enchant ability (not working now, mostly because I probably need to introduce separate registry for test mod).

> We're going to need to be careful here about how CompoundTags are passed around - we probably need to do some defensive copying in a couple of places to make sure we don't accidentally share tags between multiple items.

Hm ... I am not quite sure how it this possible? You mean in cases when some other upgrade will return constants instead of building new tags? 

UPD: I also will update todo with missed client logic part

SirEdvin:
Okey, render method is introduced, and some manual test was performed. 
Of course, UpgradeContainer somehow (?) introduces random nbt caches here and there and also damage value propagation. Not sure if damage values is UpgradeContainer fault, but I guess, I will solve this anyway.

SquidDev:
> Hm ... I am not quite sure how it this possible? You mean in cases when some other upgrade will return constants instead of building new tags?

Oh, also that I guess! I was mostly thinking if you store the upgrade data directly (i.e. `itemStack.setCompoundTag(upgradeData)`) in your `getUpgradeItem` method, then if the turtle later changes its upgrade data, it shouldn't update already-existing item stacks.

SirEdvin:
Yep, this is also possible now with existance of UpgradeContainer. And probably this is a source of my issue. I will try to make some copies here and there.

Probably, what I can do is make sure that each new instance of UpgradeData will have a copy of nbt data, rather then actual one. This probably an overkill, but this also can protect code from having to many links to NBT data, that should be ideally hidden inside turtle.

SirEdvin:
Okey, now everything seems to work fine, I just need to port this changes to forge version and add some autotests :)

projects/common-api/src/client/java/dan200/computercraft/api/client/turtle/TurtleUpgradeModeller.java
```diff
@@ -34,6 +37,20 @@ public interface TurtleUpgradeModeller<T extends ITurtleUpgrade> {
...
+     * Obtain the model to be used when rendering a turtle peripheral.
+     * <p>
+     * Used when turtle access object doesn't exist. For compatibility reasons, by default this method call getModel
+     * that depend on turtle.
+     *
+     * @param data  Upgrade data instance for current turtle side
+     * @param side    Which side of the turtle (left or right) the upgrade resides on.
+     * @return The model that you wish to be used to render your upgrade.
+     */
+    default TransformedModel getModel(@Nonnull UpgradeData<T> data, TurtleSide side) {
```
SquidDev: I think we should probably unpack the `UpgradeData` here and make this `T upgrade, CompoundTag tag,` to match the other method.

projects/common/src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
```diff
@@ -105,19 +109,19 @@ private static void storeItem(Player player, ItemStack stack) {
         }
     }
 
-    private static @Nullable IPocketUpgrade findUpgrade(NonNullList<ItemStack> inv, int start, @Nullable IPocketUpgrade previous) {
+    private static @Nullable ItemStack findUpgrade(NonNullList<ItemStack> inv, int start, @Nullable UpgradeData<IPocketUpgrade> previous) {
```
SquidDev: Can we make this return an `UpgradeData<IPocketUpgrade>` instead? This would mean we don't need to do the lookup again in `equipBack`.

SquidDev (changes requested):

projects/common-api/src/main/java/dan200/computercraft/api/upgrades/UpgradeBase.java
```diff
@@ -50,6 +52,45 @@ public interface UpgradeBase {
...
+
+    /**
+     * Specific hook, that allow you to filter data, that should be stored when turtle was broken.
+     * Use this in cases when you don't need to store all upgrade data by default, and you want to change this
+     * behavior, for example, as modem do.
+     * @param upgradeData NBT data that currently stored for this upgrade
+     * @return filtered version of this data, by default just all of it
+     */
+    default @Nonnull CompoundTag getPersistedData(CompoundTag upgradeData) {
+        return upgradeData;
```
SquidDev: Did we want to make this return a new compound tag? Happy to go with either, just wasn't sure what the conclusion of our discussion was.

SirEdvin: So, your initial suggestion was to return same tag. At first, I was thinking about returning empty new Compoung Tag, but I think now that returning same tag is much intuitive way to work this.

SirEdvin: I have updated ModemUpgrade code to return empty CompundTag, so it will not store modem state on turtle break

projects/common/src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
SirEdvin: Sure, done

projects/common-api/src/client/java/dan200/computercraft/api/client/turtle/TurtleUpgradeModeller.java
SirEdvin: Done

SirEdvin:
Okey, should I try to add some automatic tests also? I have tested turtle items manually and it seems to works perfect.

SquidDev:
> Okey, should I try to add some automatic tests also?

If you're happy to, that would be great, but don't feel like you have to!

SirEdvin:
Eh, unfortunately, I would rather avoid doing this, because I am a little lazy :) In any case, I guess, I am planning to use this functionally a lot, so probably if I found any problem, I would introduce fixes and tests for it.

Test seems passing locally, so I am not sure, that I can do something with this failed test.

SquidDev:
I had a couple of small/boring tweaks I was going to make locally, and then things got away from me 😅. I'm going to push what I've done soonish, and then please tell me if it ruins everything or not!

SquidDev (approved):
I'm happy with this if you're okay with my changes! Thank you so much for you work on this, looks really good and excited to do some things with it. :heart: :tada:

SirEndii:
I'm also excited to use this! ❤️‍🔥

SirEdvin:
Hello, thanks for your changes! I am a bit worried about more mature protect solution that just copy every time, but I guess, it is much better anyway. And also extra thanks for documentation, which is not my strong side :)

I am pretty fine with this changes, thanks again for your time!

(Which enchantable tools CC:T addons gonna loose like half of its popularity, which is fine by me :) )

SquidDev:
Awesome!

PR #1469, opened 2023-06-07
Add toggleModem method to Turtle API

merlinlikethewizard:
I've been wanting a way to activate wired modem blocks via turtles, and I'm hoping I've found an elegant solution.

I think it would add a lot of creative ways to automate turtle wired networks if turtles could place and use wired modem blocks without needing a player to come right-click on them. I believe this was attempted here https://github.com/cc-tweaked/CC-Tweaked/pull/510, but apparently never finished.

My first approach was to add a "toggle" method to a wrapped wired modem block peripheral, but this ran into the issue of running in the wrong thread and not being able to access level block entities. Then I tried adding a new Turtle API method "toggleModem," and this seemed to be the best option, with the smallest footprint in terms of code changes.

I'm new to modding, so I'm definitely open to any feedback on this, but I really hope some version of this functionality can be added in. Thanks!

### DIFF
--- a/projects/common/src/main/java/dan200/computercraft/shared/peripheral/modem/wired/WiredModemFullBlockEntity.java
+++ b/projects/common/src/main/java/dan200/computercraft/shared/peripheral/modem/wired/WiredModemFullBlockEntity.java
@@ -141,6 +141,10 @@ public class WiredModemFullBlockEntity extends BlockEntity {
         return InteractionResult.CONSUME;
     }
 
+    public void use() {
+        togglePeripheralAccess();
+    }
+
     private static void sendPeripheralChanges(Player player, String kind, Collection<String> peripherals) {
         if (peripherals.isEmpty()) return;
 
--- a/projects/common/src/main/java/dan200/computercraft/shared/turtle/apis/TurtleAPI.java
+++ b/projects/common/src/main/java/dan200/computercraft/shared/turtle/apis/TurtleAPI.java
@@ -768,6 +768,18 @@ public class TurtleAPI implements ILuaAPI {
     }
 
 
+    /**
+     * Toggle a wired modem block in front of the turtle to allow / disallow peripheral access.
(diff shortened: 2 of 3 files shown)


fatboychummy:
My only thought would be to add `toggleModemUp` and `toggleModemDown` like the other turtle methods have.

I'll also mention that this is something that I've requested earlier as well ( #264 ). Unfortunately I never got to learning Java myself to implement it... I must thank you for doing this!

SquidDev:
Thank you for putting this together!

I'm afraid I'm not going to merge this. While I understand the use-case, it feels strange to me to add a method to the turtle API for interacting with a single block.

An alternative approach here would be allow right clicking modems via `turtle.place()`. This isn't something I want to add to CC:T (see #1305 for some discussion on this), but you can do it via a datapack: create a `data/computercraft/tags/blocks/turtle_can_use.json` file with the following contents:

```json
{"values": ["#computercraft:wired_modem"]}
```

See the [datapack-example](https://github.com/cc-tweaked/datapack-example/blob/dev/data/computercraft/tags/blocks/turtle_can_use.json) for a concrete example.

merlinlikethewizard:
Hi @SquidDev! I totally understand your point, and I had no idea this was possible with a data pack (thank you for letting me know, I searched all over for a way to do this).

I'll only make one plea for the tag being added by default, which is that if this was a part of the base mod it would open turtles up to a whole world of fully automated behavior with wired networks that there is no way for them to have otherwise. You said in #1305 that you don't want turtles using items and blocks arbitrarily, which I agree with because it allows for game-play where players need to get more inventive with their solutions. For wired modems however, there is no work around solution for fully automated wired networks built by turtles, and there's a whole continent of creative possibility that's sadly, and just barely, out of reach.

Obviously it's totally up to you and either way thank you for taking the time to review this and keep up the good work :)

PR #1485, opened 2023-06-17
Allow changing turtle upgrades from the GUI

SquidDev:
This exposes a couple of slots on the right hand side of the turtle UI, which display the currently equipped upgrades. Items can be added/removed to swap them out as upgrades.

## Screenshots
With no upgrades:

A screenshot of a ComputerCraft turtle's UI. The UI is composed of a computer terminal, with the player's and turtle's inventories below. Sticking out the right hand side of the UI is two additional slots for the turtle's upgrades. Each slot is empty and has a small pictogram of a turtle, with a line on the left or right indicating which side of the turtle it represents.

With both upgrades:

A screenshot identical to the above, however the two upgrade slots have an ender modem and diamond sword in respectively.

## Comments/questions?

 - I've put these upgrade slots on the right hand side as that's where the turtle's inventory is, and it feels good to group the two. However, maybe there's an argument for putting it on the left, as that would match the computer sidebar.
 - The little turtle upgrade icons are meant to [look like turtles from behind](https://github.com/cc-tweaked/CC-Tweaked/assets/4346137/88f4653f-d1b4-4164-bab5-bd6cce09b97b), but I'm not sure I've done a great job there. Ways of making this look more obvious are welcome!


Lupus590:
> The little turtle upgrade icons are meant to look like turtles from behind, but I'm not sure I've done a great job there

It seemed obvious to me, it may take someone a few attempts to figure out that these are the equip slots, but I am sure that most people will realise what effect these slots have when leaving the GUI.

> However, maybe there's an argument for putting it on the left, as that would match the computer sidebar.

I feel like that might lower the association with the turtle's inventory which I think might delay the discovery of these being equip slots for some people.

PR #1489, opened 2023-06-18
Add ability to serialize Unicode strings to JSON

MCJack123:
This PR adds an option to `textutils.serializeJSON` that tells it to interpret strings as UTF-8 sequences instead of plain ANSI strings. This allows sending Unicode strings to services that take in JSON data. The driving use case is to be able to send Unicode messages to Discord - currently, the UTF-8 sequences get converted to independent Unicode characters, breaking the output.

To make this work, I had to adjust the definition of `serializeJSON` to also take a table as a second argument instead of a boolean. This was already done in other functions, so it should be fine here, and it also modernizes the parameters to match the other functions.


projects/core/src/test/resources/test-rom/spec/apis/textutils_spec.lua
```diff
@@ -174,6 +176,17 @@ describe("The textutils library", function()
...
+
+        it("serializes NBT style", function()
+            expect(textutils.serializeJSON({ test = 2 }, { nbt_style = true })):eq('{test:2}')
+            expect(textutils.serializeJSON({ test = 2 }, true)):eq('{test:2}') -- old style
+        end)
+
+        it("serializes Unicode strings", function()
+            expect(textutils.serializeJSON("\xe3\x81\x93\xe3\x82\x93\xe3\x81\xab\xe3\x81\xa1\xe3\x81\xaf", { unicode_strings = true })):eq('"\\u3053\\u3093\\u306B\\u3061\\u306F"')
+            expect(textutils.serializeJSON("\xf0\x9f\x98\xaf", { unicode_strings = true })):eq('"\\uD83D\\uDE2F"')
+            expect(textutils.serializeJSON("\\\"\xc3\xbf\n\"", { unicode_strings = true })):eq('"\\\\\\"\\u00FF\\n\\""')
```
SquidDev: This might be a little cleaner if we use unicode escapes and long strings, as we can remove a bunch of escapes:

```suggestion
            expect(textutils.serializeJSON("\u{3053}\u{3093}\u{306B}\u{3061}\u{306F}", { unicode_strings = true })):eq([["\u3053\u3093\u306B\u3061\u306F"]])
            expect(textutils.serializeJSON("\u{1f62f}", { unicode_strings = true })):eq([["\uD83D\uDE2F"]])
            expect(textutils.serializeJSON("\\\"\u{00ff}\n\"", { unicode_strings = true })):eq('"\\\\\\"\\u00FF\\n\\""')
```

projects/core/src/main/resources/data/computercraft/lua/rom/apis/textutils.lua
```diff
@@ -824,7 +844,13 @@ unserialise = unserialize -- GB version
 --
 -- @param t The value to serialise. Like @{textutils.serialise}, this should not
 -- contain recursive tables or functions.
--- @tparam[opt] boolean bNBTStyle Whether to produce NBT-style JSON (non-quoted keys)
+-- @tparam[opt] { nbt_style? = boolean, unicode_strings? = boolean } options Options for serialization.
+--  - `nbt_style`: Whether to produce NBT-style JSON (non-quoted keys) instead of standard JSON.
+--  - `unicode_strings`: Whether to parse strings as containing UTF-8 characters instead of
+--    using the default 8-bit ANSI character set.
```
SquidDev: I think it's probably a little confusing to describe this as parsing (it is, but probably best reserved to the parsing `unserialiseJSON` does).

```suggestion
--  - `unicode_strings`: Whether to treat strings as containing UTF-8 characters instead of
--    using the default 8-bit character set.
```

SquidDev (changes requested):
Thank you! A couple of wee nitpicks, but otherwise looks good!

I'm a little eh about `unicode_strings`, but don't really have any better suggestions. `encode_unicode` and `encode_utf8` aren't really any better.

SquidDev:
Hah, guess I need to add support for this to illuaminate :flushed:.

SquidDev (approved):

LinearNoodle:
Hey guys, I'm getting the error "textutils.lua:446: malformed pattern (missing ']')" when using the textutils.lua from this version when calling serializeJSON on non-unicode strings. It works great on unicode strings and does exactly what its supposed to, but with normal strings it gives that error.

LinearNoodle:
The issue is due to the \0 character on line 446

SquidDev:
That pattern is the same as what what we had before - it's just been indented. Can you revert to the previous version and check whether that worked or not?

How are you testing this change? Are you running this in PUC Lua 5.1 - that has a bug with their handling of `\0`, which is fixed in later versions (and in Cobalt, which CC uses).

LinearNoodle:
I am using CraftOS-PC Accelerated, so that might be the reason then, my bad
I tested it by just downloading textutils.lua and loading it through os.loadAPI

SquidDev:
Right yes, CraftOS PC uses `%z` instead of `\0`.

https://github.com/MCJack123/craftos2-rom/blob/7e84ee2ace1ae740e2a33a8b6c5e93213a192205/rom/apis/textutils.lua#L424

**Edit:** @MCJack123, does CraftOS-Lua not handle `\0` in patterns? You're [getting the length in `gsub`](https://github.com/MCJack123/craftos2-lua/blob/445025c53cb7cb61d74c675c35bfcb4a28cf4126/src/lstrlib.c#L683), so not sure what's going on here.

**Edit 2:** _Oooh_, Accelerated. Right, LuaJIT won't have this patch.

MCJack123:
Yep, I had a patch for this issue in LuaJIT in my own fork of the ROM, so it'll be fixed once this is brought over there.

Issue #1408, opened 2023-04-09
textutils.unserializeJSON throwing 'Too long without yielding'


SammyForReal:
### Description
I noticed some behaviour when parsing large **.json** files. The file I used here is [this one](https://github.com/json-iterator/test-data/blob/master/large-file.json), with about 25MB of size.

When trying to unserialize the data of the file with `textutils.unserializeJSON`, the following message appears:
`...rom/apis/textutils.lua:562: Too long without yielding`

### How to reproduce
This screenshot shows the steps I took to reproduce this error:
![image](https://user-images.githubusercontent.com/53350357/230801739-31ec1eab-7304-43d0-be55-cc1d54606aea.png)
> _MC Version: 1.19.X; Mod version: 1.104.0_
This also happens in emulators like CraftOS-PC or CCEmuX

### Expected behaviour
Expected behaviour would be not to get this error instead of the result, even if it takes some time.
As compromise, maybe some kind of precheck can be included to make sure the given data is not _too_ large?

Of course, the chance of someone parsing such large files is not that high. On the other hand, I did not test this with _less larger_ files, that may be more realistic.

Either way, what the current behaviour is does not seem to be a proper solution to me.
And even though it seems unrealistic to parse such large files, it is not impossible.

I could not find an already existing issue, which is why I though I should open this one to make sure the community is aware of this.

DevWolf31:
This is caused by the coroutine library. It's something related to the coroutine timeout.

SquidDev:
I don't think we'd ever aim to support parsing JSON files larger than 1MB, which should currently parse fine.

That said, there are probably some performance improvements which could be made here. I suspect the main bottleneck is `parse_string`, which we could probably improve a bit by using `string.find` rather than iterating over each character (much like we do for the Lua lexer).

migeyel:
Another option would be to let the caller pass a reader function and behave like `load` does. The reader function can yield every once in a while if the input is too long.

PR #1500, opened 2023-06-25
Added sleep() to textutils.serializeJSON and textutils.unserializeJSON

WindClan:
Fixes a bug where giving a large enough JSON to textutils.unserializeJSON made it give a "Too long without yielding" error.


SquidDev:
Thank you for your PR!

I'm afraid I'm not going to merge this. I'm definitely happy to accept optimisations to the JSON parser (see discussion in #1408 and 0e48ac1dfe9896bc62bcfb02218adc4d97e8a6c6), but I don't want to allow parsing arbitrarily large JSON. The current parser should be good for up to a few MBs, but after that I don't think loading the whole thing into memory in one go is something we want to encourage.

PR #1510, opened 2023-07-02
docs: Fix turtle.refuel usage code

penguinencounter:
For usage example no. 1:
line 2 uses "new_level" before it was defined; changed to "level" instead

- the code *should* error when the fuel level is "unlimited", but never does (`nil == "unlimited"` is false)


SquidDev:
Good spot, thank you!

PR #1523, opened 2023-07-17
Add alternative API for retriving peripheral object of pocket computer

SirEdvin:
Method `getUpgrades()` for Pocket Computers now deprecated and should not be used anymore, but in some cases getting peripheral object from Pocket Computer can be quite useful (in my case I need access to it to extract information about internal upgrades, that stored inside this one), so this PR adds another method to solve this issue.


SquidDev:
What cases do you need this? Isn't the current `IPeripheral` passed around to all the same places as `IPocketAccess`?

SirEdvin:
So, general idea was to use this for retrieving internal peripherals from upgrade in pocket access when I have only pocket access. :) But it seems I am just copied turtle api aproach for this, so I probably come up, if I would have real use case for this.

PR #1531, opened 2023-07-22
Automatically bake models from turtle modellers

SquidDev:
This avoids us having to list a bunch of model locations twice, and hopefully makes things a little cleaner for add-on authors too

Putting this up as a PR for slightly better visibility, in case this any anybody who uses the API has thoughts.


SirEdvin:
Oh, this is a great change, thanks you for this!

PR #1532, opened 2023-07-22
Render enchanted tools (and other upgrades) with a glint

SquidDev:
- [x] Item models
 - [x] Block entities - I think this is just using platform-specific code to render the various models.
 - [x] Profile this - I _think_ it's fine, but it's possible we'll need some caching in the BER and and the Forge `FoilModel`.


SquidDev:
A screenshot of four turtles at four corners of a square. The bottom left and top right are upside down. The two left ones carry diamond pickaxes, and the right two carry (enchanted) netherite picks. The fact that they're enchanted is the exciting bit - we didn't use to support that!

Client-side game tests were worth it for this.

Issue #1530, opened 2023-07-21
GPS return nil value at some coordinate


kulapa2000:
### Minecraft Version

1.18.x

### Version

1.101.2

### Details

![2023-07-21_20 56 00](https://github.com/cc-tweaked/CC-Tweaked/assets/119069453/ab41e80e-6e16-4ff0-a104-32092d6eff10)

I was using the ender modem to get the gps coordinates, and as the screenshot shows, gps works normally at (1437,8,1148) but it returns nil after it went one step forward, which is at (1437,8,1149)

SquidDev:
Thank you for the report! Just a couple of questions:
 -  What are the positions of your GPS hosts? Just so I can attempt to reproduce this locally!
 - Could you run `gps.locate(2, true)` as well, and screenshot the result - this should include some additional debugging information, which will help explain why it cannot find the position.

kulapa2000:
> 

I was using it on a forge server, but when i log into the server today, gps works normally at that coordinate. But please don't close this issue for now, because this bug has showed up for a several times, I will follow up next time it shows up.

kulapa2000:
> 

I got a new error coordinate this time.
The gps return a nil value at (1411,70,100)

The host position are as follows:
(1557,69,948) 
(1557,69,953)
(1562,69,948)
(1557,73,948)

And when I run the gps.locate(2,true) it says ambiguous position
![image](https://github.com/cc-tweaked/CC-Tweaked/assets/119069453/1108a3b1-684d-414d-a620-c21d77885354)

SquidDev:
Thank you, that's very helpful!

The problem here is that you're very far away from your GPS array, which causes accuracy to suffer. In this particular case, we can't disambiguate on the y axis between $y = 68$ and $y = 70$. If we compare the distances to the GPS host at $(1557, 73, 948)$, we can see why:

$$
\begin{align}
\text{distance}(y = 68) &= \sqrt{146^2 + 5^2 + 848^2} \approx 860.492 \\
\text{distance}(y = 70) &= \sqrt{146^2 + 3^2 + 848^2} \approx 860.483 \\
\end{align}
$$

Because the two distances are less than 0.01 away from each other, the GPS library decides that's not accurate enough to guarantee the correct position.

The easiest fix here is just to move the GPS hosts further apart. Just shifting the top computer up a few more blocks should help here!

[...]

PR #1542, opened 2023-07-30
GPS Guide, add note about the accuracy problem

Lupus590:
Related #1530


Lupus590:
> Execution failed for task ':fabric:runGametest'.

I don't know how to fix that or how I managed to make a change that would break it in the first place.

SquidDev:
Yeah, I suspect that's another instance of #1373.

PR #1549, opened 2023-08-03
Typo `instealled` -> `installed` and some grammar "improvements"

SirEndii:
I've fixed a typo where `installed` was spelled incorrectly. I've also fixed some grammar issues and changed some styles. I'm not going to list them all up, the file comparison in GitHub should show them well 😅 

Maybe you do not agree with every change, just let me know what you don't like, and I'm gonna change it back


SquidDev:
Thanks!

PR #1554, opened 2023-08-09
Change tools to also be peripherals

### LINKED ISSUE #964: Turtles should be able to detect Equipped Pickaxes
Turtles currently have no way of know if they have pickaxe equipped.

My current workaround is unequipping and detecting if anything appeared in the turtles inventory

It would be nice if turtles, could use the peripheral.Find to detect if theres a pickaxe.

would save me like 2 headaches.

ToxicGLaDOS:
This fixes #964 by making pickaxes (and all tools) peripherals in addition to tools. You can use `peripherals.getType(side)` and get the type of tool.

This is just a draft for now to ask if this is a reasonable approach. I haven't added any test cases for the new functionality like `peripherals.call("left", "dig")`, nor have I fixed the style fixes for now. If we want to accept this I'll make those changes.

I'd appreciate any recommendations.


Wojbie:
I believe this solution was already deemed to sadly break things in https://github.com/cc-tweaked/CC-Tweaked/issues/964#issuecomment-979309073

PR #1558, opened 2023-08-14
Add dark mode styling for recipes

MineRobber9000:
Exactly what it says on the tin.

An example of the dark mode recipe styling in action, using the recipes from the modem page.


SquidDev:
Thank you!

PR #1564, opened 2023-08-17
Fixed documentation: Turtle API in case of dropDown method

vehsamrak:
Docblock for dropDown() was similar to drop(), so I changed it to clearify difference.
Not sure how to add it to all supported versions, but ready for doing that.


SquidDev:
Thank you!

PR #1569, opened 2023-08-23
Remove compression from terminal/monitor packets

SquidDev:
Minecraft already adds its own compression, so there's not much reason to do this. Well, I _think_ - I need to test the effects of this change on a server.

### DIFF
--- a/projects/common/src/main/java/dan200/computercraft/shared/computer/terminal/TerminalState.java
+++ b/projects/common/src/main/java/dan200/computercraft/shared/computer/terminal/TerminalState.java
@@ -5,18 +5,10 @@
 package dan200.computercraft.shared.computer.terminal;
 
 import io.netty.buffer.ByteBuf;
-import io.netty.buffer.ByteBufInputStream;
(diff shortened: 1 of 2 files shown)


SquidDev:
Okay, so set up a small testing grid here with 14 lots of 8x6 monitors, each at 0.5 times terminal size. The uncompressed packet here is a little over 26KiB.

## With compression
The sent packets end up being compressed to 6KiB, which at 20Hz, would put as at around 1.6MiB/s.

The actual numbers I end up measuring from btop and nethogs are closer to 1MiB/s, so not entirely clear what's going on here. Unfortunately Minecraft doesn't expose any stats for compression by default, and I think adding them is going to be too much of a pain here.

## Without compression
The sent packets are not compressed, and so we can see in JFR they take up 26KiB as expected. Compression clearly kicks in here, and measured network usage appears to be about 1.3MiB/s.

This is about 30% worse than the current state of things (i.e. with compression), but I think given the reduced CPU usage (no compression on the main thread) and complexity I think it's worth it.

Issue #1269, opened 2022-12-24
Add DFPWM play function to speaker


alexhorner:
Based on https://github.com/cc-tweaked/CC-Tweaked/issues/1229 and after looking at https://github.com/cc-tweaked/CC-Tweaked/blob/c3fe9f00d4e35554c00f7228ab11eefa77d21f02/projects/common/src/main/java/dan200/computercraft/shared/peripheral/speaker/SpeakerPeripheral.java#L330

Looks like the speaker `playAudio` function takes in raw 8 bit signed mono PCM samples but then immediately converts them to DFPWM. Having read the discussion, this makes perfect sense, because otherwise the audio bandwidth for servers would be quite unreasonable.

I propose that a new function `playDfpwm` be added, which allows you to pass DFPWM samples directly to the speaker. This cuts down on processing costs in game, which could help servers, as we won't have to needlessly convert audio from DFPWM to PCM and then back again.

In addition, it might be nice to add a config option, disabled by default and commented to note bandwidth implications, that allows the `playAudio` function to actually play raw PCM samples directly, for those singleplayer users and willing multiplayer servers who would prefer the higher bandwidth but better quality where they can support it. This would be a non-breaking change in "userspace".

Kan18:
I also suggested this in [a comment on 1229](https://github.com/cc-tweaked/CC-Tweaked/issues/1229#issuecomment-1328377565) but got no response. Speakers' initial development implementation [*did* accept DFPWM](https://github.com/cc-tweaked/CC-Tweaked/pull/982), but, after a [Twitter comment](https://twitter.com/cbrzeszczot/status/1467979132017561608), it was [later changed](https://github.com/cc-tweaked/CC-Tweaked/pull/982#issuecomment-987337735) to accepting PCM from Lua code because it "possibly feels more natural than our speaker 'hardware' consuming DFPWM data", and it being "invisible to players" (I don't exactly get this part, since (human) players still have to encode their music to DFPWM - did this mean music player programs?)

SquidDev:
> since (human) players still have to encode their music to DFPWM 

They don't _have_ to. It's the native format supported by the `speaker` program, but it's perfectly fine to play raw PCM audio (for instance, from WAV files).

Really, the use of DFPWM on the Java/network side is an implementation detail. As mentioned in #1229, its there as an aesthetic choice (I want the audio to be a little tinny), but there's other ways that could be implemented.

[...]

PR #1574, opened 2023-08-28
Add config option to use PCM audio over network

MCJack123:
This PR adds a `speakers_use_pcm` config option to the server config, which allows users to choose to use full PCM audio on the backend in `speaker.playAudio`, rather than encoding to DFPWM first. This allows a choice of better quality audio, at the cost of higher network bandwidth. It defaults to `false` to keep prior behavior and reduce bandwidth requirements. It works by adding a single byte flag to the network message, which indicates the encoding type. If set, the client will decode the audio as unsigned 8-bit PCM; otherwise, it'll decode as DFPWM.


SquidDev:
Thanks for the PR! I'm afraid I'm not going to merge this - the crunch of DFPWM is an intentional aesthetic choice, and I don't really want to remove that, even behind a config option.

As mentioned in #1269, I'm definitely open to changing the audio encoding used for transport to something else, but I do want to have something which feels slightly rough around the edges.

TheRedXD:
This should be reopened and pushed. Many people have been wanting this feature, and a subjective aesthetic choice that is so easy to make configurable should not be restricted. It is a bigger benefit to push this as an option in the main repository than try to spin up a new repository which needs constant maintenance just to keep a singular change.

PR #1579, opened 2023-09-03
Use method handles in our generated Lua methods 

SquidDev:
When the target method is in a different class loader to CC, our generated method fails, as it cannot find the target class. To get around that, we create a `MethodHandle` to the target method, and then inject that into the generated class (with Java's new dynamic constant
system). We can then invoke the `MethodHandle` in our generated code, avoiding any references to the target class/method.

This should have equivalent performance to the original code. I've included some benchmarks in this PR (though have reverted them before merging), which might be useful in the future.


SquidDev:
A couple of other things possibly worth noting here:

 1. We could theoretically push more of the generation to `MethodHandle`s. For instance, use a terrible mixture of `dropArguments` and `foldArguments` to extract each argument, and then `filterReturn` to wrap the result into a `MethodResult`.
 2. Now that the generated class doesn't reference the target method, we could reuse the generated bytes for other methods with the same signature. If we took 1. to the extreme, all classes would reuse the same bytes, just varying the class data we initialise the class with

It doesn't feel worth doing either right now - the current code is relatively stable, and I don't think alternative approaches will be any simpler.

Issue #1246, opened 2022-12-12
`loadfile` should read files with binary mode


walksanatora:
this small change would bring CC-T a bit closer to the real lua intepreter because modules and scripts keep their binary data instead of replacing it all with `?`

just replace
`r` with `rb` in [bios.lua](https://github.com/cc-tweaked/CC-Tweaked/blob/mc-1.19.x/projects/core/src/main/resources/data/computercraft/lua/bios.lua#L467)

walksanatora:
interesting
even opening in rb mode is not consistent with lua
because `0x00` is replaced with \000
nvm that was `cc.pretty` formatting it since nul has no charachter

walksanatora:
but after running a few CraftOS programs with this small change
it would appear nothing breaks
(though a full test should probally be conducted)

SquidDev:
Unfortunately, this breaks program behaviour whenever a program uses a non-ASCII character. For instance, here's a program which creates an `image.lua` file which contains some of the drawing characters:

```lua
local out = fs.open("image.lua", "w")
out.write([[print("]])
out.write("\x87\x83\x8b")
out.write([[")]])
out.close()

local function run(path, mode)
  local h = fs.open(path, mode)
  local contents = h.readAll()
  h.close()
  local ok, err = load(contents, "@" .. path)
  if ok then ok, err = pcall(ok) end
  if not ok then printError(err) end
end

run("image.lua", "r")
run("image.lua", "rb")
```

When running the resulting program under text and binary mode, we get the following output:

![A screenshot of a ComputerCraft terminal running the above program. The first line displays a continuous line of drawing characters, while the second line is interspersed with UTF decoding errors](https://user-images.githubusercontent.com/4346137/207011520-800f1010-ee82-4f1e-a9d7-75b044980f48.png).

I've vaguely mentioned this in #1233, but I think the correct solution here really is just to remove text mode entirely, and have binary mode be the only mode. This is pretty much how Lua works on UNIX systems anyway. But that is a breaking change, and I'm not sure how large the repercussions of that would be, so unlikely to happen any time soon.

walksanatora:
write in text mode, read in binary mode
write in binary mode, read in text mode
(the source of all pains of the world)

walksanatora:
also by modifying the example you gave (replacing `w` with `wb`)
the `r` prints `???` and `rb` prints the properly formatted string

walksanatora:
[...]

Issue #1233, opened 2022-11-27
HTTP Get seems to strip trailing whitespace when getting files


tehbiga:
### Minecraft Version

1.18.x

### Version

1.101.0

### Details

`http.get` seems to be stripping trailing whitespace / newlines from response bodies. I was able to compare the responses side by side with a browser.

***Story:***
Was writing a program that connects to my home computer to transfer files with a generated install script which checks crc8 values and noticed the crc values were _always_ wrong. I realized after a bunch of testing that if I call `http.get` on my file and dump the contents with `res.readAll()` the final new line was missing. If I request the same url from a browser and save the file the newline is there.

_Computer `http.get`:_
https://i.imgur.com/Kqj18S4.png

_File downloaded with browser:_
https://i.imgur.com/DXthola.png

SquidDev:
If you want an exact match, you probably need to use [http.get's binary mode](https://tweaked.cc/module/http.html#v:get). The default text mode does a whole bunch of extra stuff, including normalising line endings and dealing with unicode.

tehbiga:
> If you want an exact match, you probably need to use [http.get's binary mode](https://tweaked.cc/module/http.html#v:get). The default text mode does a whole bunch of extra stuff, including normalising line endings and dealing with unicode.

`wget` uses binary mode and has the same problem.

SquidDev:
Hrmr, can you give me a URL to test with and your code? They appear to be preserved for me.

![A screenshot of CCEmuX, downloading https://squiddev.cc/r/demo.lua. The downloaded file contains two trailing new lines.](https://user-images.githubusercontent.com/4346137/204132338-e8c7dde1-662b-455b-b397-93f34431a8d3.png)

tehbiga:
I'll do some more testing tonight and figure it out.

tehbiga:
@SquidDev I figured out where all of the confusion is coming from:

Reading the file with mode `r` is stripping the last newline character.
Reading with mode `rb` shows the missing newline.

I can switch to using binary reads _(and I should have from the beginning)_, but this also means that the `edit` program is not reading in binary mode and the editor view is hiding the newline.
This is where all of my confusion was coming from not being able to verify with the built-in editor. When I'm editing files on my real computer I obviously see the newline and in-game I do not, causing the confusion.

[...]

PR #1598, opened 2023-10-01
Add binary option to WebSockets

MCJack123:
Currently, handling of UTF-8 codes in WebSocket text messages is handled entirely on the Java side. This is understandable for most uses, but some users may want to interact with the bytes directly (e.g. with `utf8`) without Java meddling with codepoints outside CC's range.

This PR adds a binary flag to `http.websocket` that disables UTF-8 decoding of text messages on the Java side. When enabled, the binary flag of each message remains intact, but the data received will always be interpreted as binary data - no more `?`s replacing out-of-range codes. Sending text messages on a binary handle will cause Java to reinterpret the bytes as UTF-8 before sending, preserving manually-encoded codepoints. (This is an implementation detail, since `TextWebSocketFrame` takes a UTF-16 `String`, so we need to convert it from UTF-8 ourselves.) If that fails for some reason, the sender will fall back to normal bytewise strings.

The syntax for this flag matches `http.get`'s argument list, so there should be no confusion between the two lists, and no compatibility issues should arise - WebSockets still default to normal/text handles. I've added a test suite as well (which duplicates the original WS test) that ensures that UTF-8 bytes are preserved through the socket.

image

### DIFF
--- a/projects/core/src/main/java/dan200/computercraft/core/apis/http/websocket/WebsocketHandle.java
+++ b/projects/core/src/main/java/dan200/computercraft/core/apis/http/websocket/WebsocketHandle.java
@@ -90,7 +90,10 @@ public class WebsocketHandle {
                     var bytes = new byte[buf.capacity()];
                     buf.get(bytes);
                     data = new String(bytes, StandardCharsets.UTF_8);
-                } catch (UnsupportedCharsetException ignored) {}
+                } catch (UnsupportedCharsetException ignored) {
+                    // Suppress warnings.
+                    data = text;
+                }
             }
             websocket.sendText(data);
         }
--- a/projects/core/src/main/java/dan200/computercraft/core/apis/http/websocket/WebsocketHandle.java
+++ b/projects/core/src/main/java/dan200/computercraft/core/apis/http/websocket/WebsocketHandle.java
@@ -9,6 +9,8 @@ import dan200.computercraft.api.lua.*;
 import dan200.computercraft.core.apis.IAPIEnvironment;
 import dan200.computercraft.core.apis.http.options.Options;
 
+import java.nio.charset.StandardCharsets;
+import java.nio.charset.UnsupportedCharsetException;
 import java.util.Arrays;
 import java.util.Optional;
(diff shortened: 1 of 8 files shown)


MCJack123:
This should be passing tests, but it seems the in-game tests are randomly failing in areas I didn't touch - please disregard the failures.

SquidDev:
Thanks for the PR! Issues like this are always a pain, because the current behaviour is incorrect, and really sending/receiving raw UTF-8 is what we should have done from the start.

I have mentioned this in a couple of places (#1246, #1233) and I am tempted to just remove text mode across the whole mod, always dealing with the raw bytes. Websockets are a little different, as they mandate UTF-8, but we can verify that when sending. Obviously this would break things, but I don't know how much. Maybe an experiment to run on SwitchCraft :).


> This should be passing tests, but it seems the in-game tests are randomly failing in areas I didn't touch.

Yeah, game tests are notoriously flaky in CI. I'm not quite sure why the windows one timed out - normally that's pretty reliable. I've kick-started both.

projects/core/src/main/java/dan200/computercraft/core/apis/http/websocket/WebsocketHandle.java
```diff
@@ -79,7 +81,21 @@ public final void send(Coerced<String> message, Optional<Boolean> binary) throws
...
-            websocket.sendText(text);
+            var data = text;
+            if (websocket.isBinary()) {
+                // Try to convert the string from UTF-8 bytes to UTF-16 codepoints.
+                // If this fails, fall back to normal ANSI.
+                try {
+                    var buf = LuaValues.encode(text);
+                    var bytes = new byte[buf.capacity()];
+                    buf.get(bytes);
+                    data = new String(bytes, StandardCharsets.UTF_8);
```
SquidDev, line 92: It might be better to use a `CharsetDecoder` here instead, and then fail on `CharacterCodingException` errors, instead of replacing invalid characters with U+FFFD. Not sure, leave as is, will have a think about it.

PR #1615, opened 2023-10-25
Remove text mode, update Cobalt

### LINKED ISSUE #768: Update to Lua 5.2
As mentioned in [this discussion](https://github.com/SquidDev-CC/CC-Tweaked/discussions/757), there's been some fantastic work to update CC:T to Lua 5.2.

Obviously we want to get these changes integrated, as I want `goto` as much as the next person. However, getting this user-ready is probably going to require a fair amount of other work (see also, the other issues in this project).

 - [ ] Merge SquidDev/Cobalt#50.
 - [x] Drop our custom version of `load` - this has been handled within the Lua VM for a while now.
 - [ ] Test with some of the most popular programs (well, those which still work on latest CC D:).

### LINKED ISSUE #1246: `loadfile` should read files with binary mode

### LINKED ISSUE #1478: io.open(<path>, 'r+b') does not allow writing
### Minecraft Version

1.18.x

### Version

1.101.2

### Details

_Issue:_
`io.open(<path>, 'r+b')` does not allow writing as standard Lua does preventing me from updating chunks of files.

According to the documentation seeking should be available for binary file handles, but trying various combinations of `rb`, `wb`, and `ab` with and without `+` in the middle, or at the end, does not allow me to modify a file without losing contents or update at an arbitrary seek position.

Not sure if this is intentional or if I've missed something in the docs. I know there are differences between official Lua interpreters and the mod's, but I'm assuming this was missed since everything else seems to work as expected.

_In-game test:_
https://i.imgur.com/v83tUmP.png

_Expected:_
Here is an ilua example I did on my desktop to show standard functionality:
```lua
> f = io.open('test.txt', 'wb')
> f:write('Hello world')
true
> f:close()
true

> f = io.open('test.txt', 'r+b')
> f:seek('set', 6)
6
> f:write('there')
true
> f:close()
true

> f = io.open('test.txt', 'rb')
> f:read('*a')
'Hello there'
```

### LINKED ISSUE #1534: HTTP API corrupts request payload
The HTTP API implementation reinterprets `POST` payloads as text regardless of the binary setting. I ran into it when trying to `POST` binary data to a server and getting garbage on the other side.

Also, the default headers are `"Content-Type": "application/x-www-form-urlencoded; charset=utf-8"` and `"Accept-Charset": "UTF-8"` irrespective of the `binary` flag.

https://github.com/cc-tweaked/CC-Tweaked/blob/1b88213eca6e36c13fd1f37b3c09a0d5fe9c571e/projects/core/src/main/java/dan200/computercraft/core/apis/http/request/HttpRequest.java#L65-L69

Test with
```lua
local h = http.post {
  url = "https://httpbin.org/anything",
  body = string.rep("\254", 7),
  binary = true,
}

local response = h.readAll()
h.close()

-- expected: 7
-- actual: 14
print(textutils.unserialiseJSON(response).headers["Content-Length"])
```

Or simply [`pastebin run uSfZyzi8`](https://pastebin.com/uSfZyzi8).

_Originally posted by @viluon in https://github.com/cc-tweaked/CC-Tweaked/discussions/1526_

### LINKED ISSUE #1605: WGet run adds wierd characters to unicode
### Minecraft Version

1.19.x

### Version

1.101.3

### Details

When running code directly with WGet run, wierd characters appear near unicode characters. Note: They do not appear if the file is downloaded by wget and then ran from the shell.

Notice the chat.

SquidDev:
This is an experimental PR containing a number of breaking changes. Hopefully most programs will not be affected by these changes, though I have not done any testing to confirm this.

Any help with testing these changes would be much appreciated - please let me know what does/doesn't work in the comments: <sub><sup>(and don't forget to like and subscribe)</sup></sub>

[Forge 1.20.1 Download](https://squiddev.cc/maven/cc/tweaked/cc-tweaked-1.20.1-forge/1.109.0-SNAPSHOT/cc-tweaked-1.20.1-forge-1.109.0-20231025.094704-1.jar) | [Fabric 1.20.1 Download](https://squiddev.cc/maven/cc/tweaked/cc-tweaked-1.20.1-fabric/1.109.0-SNAPSHOT/cc-tweaked-1.20.1-fabric-1.109.0-20231025.094704-1.jar)

There is no guarantee that we're going to go ahead with any or all of these changes - I want to do some more testing and see what the fallout is first.

## Changes
 - [x] Remove text mode from file and HTTP handles, always returning a binary handle (fixes #1605, fixes #1534, fixes #1246).
 - [x] Send/receive the exact text in non-binary websocket messages (closes #1598).
 - [x] Add support for update (`r+`, `w+`) modes (fixes #1478).
 - [x] Update to Cobalt 0.8 (fixes #768). Tracking progress for this at https://github.com/SquidDev/Cobalt/pull/77.
 - [x] Update our Lua parser/lexer to handle `goto` and labels.
 - [x] Update documentation.

## Known breaking changes
 - The file encoding used when reading/writing files has changed. See #1246 for some discussion of possible cases where this may cause issues.
 - Bytecode dumping/loading has been removed.
 - `setfenv`/`getfenv` are now emulated by using the `_ENV` upvalue. This means that `getfenv`/`setfenv` will not work on non-Lua functions (for instance coroutines or native/library functions).

    The changes to the environments also means that `load` always uses the global environment rather than that of the current coroutine.


MCJack123:
Looks great! I'm happy to see some of these ancient quirks finally being scrapped ~~and more work being made for me~~. I'll test it out with some of my program library - I'm most curious to see how this affects Phoenix and Herobrine.

I'm a bit concerned about the removal of bytecode loading though - I have a number of legitimate (and illegitimate) uses for it, including for sending functions over modems, optimizing table generation, and compiling code. Is this done out of security concerns? I'd think that being written in Java would significantly reduce the possibility for exploitation, but I understand if this is necessary - I had a huge hole in CraftOS-PC for a long time that allowed arbitrary code execution, made possible by a bug in PUC Lua. It's fine to break anything that depends on 5.1 bytecode, but I'd appreciate it if we could still use 5.2 bytecode (with appropriate `load` mode parameters).

Also, was my work on 5.2 Cobalt at all valuable? It's become super stale since you rewrote half the compiler, but I hope it was able to save you a little bit of time while writing 0.8.

SquidDev:
> I'm a bit concerned about the removal of bytecode loading though

So I have gone around in circles a lot here, however I think this the best option in terms of future proofing. The problem here is that bytecode (as it is right now) is an inherently leaky thing. Everything about it is tied to how the interpreter works internally, which means if the interpreter changes, the bytecode format does too.

There are definitely options here which I may consider in the future (we'll need something to handle upgrading bytecode formats if/when persistence happens) but it is not a problem I want to solve now, and this is the safest option which gives room to change in the future.

>  I have a number of legitimate uses

:cowboy_hat_face: 🇽

In all seriousness, this does kill off [JVML-JIT](https://github.com/Team-CC-Corp/JVML-JIT), which remains to this day the coolest CC project out there. However, in most cases there are alternatives to generating bytecode (the presence of `goto` makes compiling to Lua source code much more practical, even if a bit slower), so I'm not going to shed any tears.

SquidDev:
> Also, was my work on 5.2 Cobalt at all valuable? It's become super stale since you rewrote half the compiler, but I hope it was able to save you a little bit of time while writing 0.8.

Oh, very much so! Thank you for it, and I'm sorry it never got merged. The parser work was still a useful reference, even if that code has changed a lot since.

SquidDev:
Bug report: writeable handles are missing the `writeLine` method:

A screenshot of the result of running `fs.open("foo", "w")`. The returned table has `close`, `flush`, `seek`, `write` methods, but no `writeLine`

Additionally, there's a slight change in behaviour between `readAll` - it now includes the trailing new line where previously it didn't. This breaks [Grin-Get](http://www.computercraft.info/forums2/index.php?/topic/20561-grin-get-a-grin-based-package-manager/), as its base64 decoder chokes on the newline.

MCJack123:
Just tested a few things:
- First thing I noticed is that `_VERSION` is `Lua 5.1`. If we're going full 5.2 compatibility, we should switch that to `Lua 5.2` so things that are optimized for it will function. Which is relevant...
- ...when [Phoenix](https://phoenix.madefor.cc) failed to boot because of a combination of the wrong version + `string.dump` returning unexpected results. Now that `string.dump` returns a text description, it's completely out of line with any of the documented outputs, no matter the Lua version. This caused an assumption on Lua 5.1's bytecode output (which wouldn't have been used if `_VERSION` was correct), which took one byte as the size of an integer, to fail, causing an error while building some bytecode. (If `_VERSION` is `Lua 5.2`, it uses a different path that doesn't use integer sizes, which would have succeeded to build but failed to load, falling back to the slow route.) Otherwise, Phoenix worked fine, besides some fixes required around the aforementioned `writeLine` bug.
- [Herobrine](https://github.com/MCJack123/Herobrine) worked just fine, no issues to report despite using some file encoding hacks (it encodes the file data in binary mode, but does a check and re-runs the file if it's in text mode).

I'll see what else I can test later, though it's looking promising for the most part. (Please update `_VERSION` though!)

---------

> So I have gone around in circles a lot here, however I think this the best option in terms of future proofing. The problem here is that bytecode (as it is right now) is an inherently leaky thing. Everything about it is tied to how the interpreter works internally, which means if the interpreter changes, the bytecode format does too.
> 
> There are definitely options here which I may consider in the future (we'll need something to handle upgrading bytecode formats if/when persistence happens) but it is not a problem I want to solve now, and this is the safest option which gives room to change in the future.

Personally, I think the current solution of outputting the disassembly from `string.dump` is half-baked, because it now rewrites a standard function to do non-standard things - the docs specifically say that it returns a string "so that a later `load` on this string returns a copy of the function." It interferes with absolutely everything that touches `string.dump` at all, making it essentially useless as-is besides debugging, and causing the standard library to once again stray from standards (which caused Phoenix to not boot at all, even with its protection around bytecode loading!).

If the main concern is the changing nature of the bytecode format, then I'd rather just have it work as expected, but very clearly say that it's not standard, like LuaJIT does. Something like making the header `\033Cob` to indicate it's not PUC Lua bytecode is enough, and it would keep the code that uses it normally stable while allowing the flexibility of changing bytecode as often as necessary. Keeping bytecode patchers happy isn't necessary - they can continue to do their cursed trickery with the added burden of decoding the formats themselves from Cobalt source, and they can complain to `/dev/null` when the format abruptly changes.

EmmaKnijn:
I'd be willing to work on the documentation when needed

SquidDev:
Bug report: should have noticed this earlier, but a zero-args `read()` behaves differently between text and binary handles (gaaah, this is entirely on me).

It wouldn't be too hard to preserve this behaviour - I kinda don't want to, but we probably should.

```lua
fs.open("gawr.txt", "r").read()  -- "a"
fs.open("gawr.txt", "rb").read() -- 97
```

SquidDev:
There's also differences between writable text and binary handles when doing `.write(65)`. Text mode writes `65` and binary mode writes `A`. I've written some tests for this in c30bffbd0fe8b32295f48eda5a8b0972dd40c666, which should help us prevent these sorts of regressions.

Might make sense to do #1478 as part of this PR too. I'm not quite sure how all these different file handle types should compose, so probably a good way to experiment.

walksanatora:
so wait are we lua5.2 now?

SquidDev:
There's definitely some more work to be done here (mostly testing, would like to tweak some docs), but think it's better to get this merged now rather than making it any bigger.

> so wait are we lua5.2 now?

Yep.

viluon:
I would like to second @MCJack123's stance that the removal of bytecode dumps and loading is a great loss. I think this is a regression in expressiveness that the vast majority of users won't even notice but that will hurt roughly the coolest 0.1% of programs.

> If the main concern is the changing nature of the bytecode format, then I'd rather just have it work as expected, but very clearly say that it's not standard, like LuaJIT does. Something like making the header `\033Cob` to indicate it's not PUC Lua bytecode is enough, and it would keep the code that uses it normally stable while allowing the flexibility of changing bytecode as often as necessary. Keeping bytecode patchers happy isn't necessary - they can continue to do their cursed trickery with the added burden of decoding the formats themselves from Cobalt source, and they can complain to `/dev/null` when the format abruptly changes.

I wholeheartedly agree. I think it's safe to assume that the people who mess with bytecode don't need a standard interface, they just need _an_ interface. After all, PUC Lua 5.1 and Cobalt have differences in how they generate bytecode and expose intermediate values as locals via the `debug` library, so in a way, Cobalt is already malcompliant. There is no standard for Lua bytecode; the best resources are reverse-engineered.

> > I have a number of legitimate uses
> 
> 🤠 🇽
> 
> In all seriousness, this does kill off [JVML-JIT](https://github.com/Team-CC-Corp/JVML-JIT), which remains to this day the coolest CC project out there. However, in most cases there are alternatives to generating bytecode (the presence of `goto` makes compiling to Lua source code much more practical, even if a bit slower), so I'm not going to shed any tears.

I don't think many people realise that general program persistence is achievable in software alone, but requires bytecode modifications to work efficiently. Let me outline a theoretical solution, which I plan to implement when I get to it.

1. Run a program in a controlled environment. Wrap any side-effecting functions in proxies that yield directly after a side-effect, and perform the necessary work in `coroutine.yield` and `coroutine.resume` to hide this fact from user code. Map coroutines to closures in the background.
2. Replace `load` with a proxy that patches the resulting bytecode, such that line info for a given function assigns a unique number to every instruction. Perform the necessary work in `error` and user-facing `debug.traceback` to hide this fact from user code.
3. On every yield, scan and persist the heap up to the environmental boundaries. Retrieve PCs across the call stack with a `debug.traceback` call. This data constitutes a complete program checkpoint.

To resume a program from a checkpoint,
1. Re-establish the heap.
2. Recreate the call stack: copy every closure on the saved call stack and edit it such that it begins with a branch on a call to a unique global identifier `setup`. Within the true branch, include code for setting up the stack, followed by a jump to the target PC for the given stack frame. Make `setup` recognise the caller and set up its locals accordingly. Return a truthy value from the first call, falsy values thereafter.
3. Begin execution by calling the patched closure corresponding to the lowest frame on the saved call stack.

That, modulo some more sandboxing work and top-level event logging, should be enough to achieve something like checkpoint/restore in userspace. Nothing in this approach should prevent it from working in nested instances and should be considerably faster than a nested Lua VM. A working CRIU library would enable sending running programs over the network, capturing deep and efficient recordings for record & replay debugging, time travel in arbitrary applications, or skipping complicated initialisation steps given enough disk space.

I don't see a way of adapting this approach to a bytecode-less Cobalt, as it relies on rewriting debug info. Maybe if an API for reading PCs directly were introduced?

SquidDev:
Amy and I spent quite a lot of time thinking about this a few years back. Restoring the Lua VM state within Lua _is_ theoretically possible (though none of us have code to show for it, so this whole discussion is moot), but there's a lot of awkward things one needs to handle (varargs, upvalues, `OP_CONCAT`). I don't think bytecode is the panacea you think it is.

You definitely some expressiveness lost by generating source code instead of bytecode, but I don't think it's going to be the limiting factor in what you're trying to do.

---

I was going to write another paragraph, but instead I refer you back to my original message: https://github.com/cc-tweaked/CC-Tweaked/pull/1615#issuecomment-1779859760. There's definitely lots of options here, but I would like you to trust me that I have thought about this, and decided this is the best one for my own sanity.

Issue #955, opened 2021-11-20
Rebalancing Ender Modems


SquidDev:
**TLDR:** Ender modems are OP.

CC 1.6 introduced the `repeat` program, which allows rednet to relay messages between computers. The intent here was to allow more long-distance wireless communication through the use of repeater
stations. However, 18 months later CC introduced ender modems, which allow for infinite range communication, making both wired modems and the `repeat` system obsolete.

Nowadays there's very little reason to craft a normal wireless modem, aside from very early game. This has several side effects:

-   One can build a single GPS tower at spawn and cater to the whole server.
-   Wireless networking is entirely trivialised, as everyone (well, every turtle) gets a wireless modem.

This has undoubtedly made things easier and more accessible, and so I don't think there's any harm leaving things as is. However, modded Minecraft (tech mods especially) is basically a big infrastructure management game. It makes me sad that CC trivialised one of its hardest problems, and it would be nice to rebalance things.

I think I'd propose the following changes:

-   Remove ender modem turtle and pocket computer upgrades. Ender Modems can now only be used as in-world peripherals. We might want to remodel and rename them so they more obviously distinct from the wireless modem[^1].

-   Ender modems now only have infinite range when talking to other ender modems. When talking to wireless modems, they use the distance of the wireless modem instead. We might want to adjust the max distances of wireless modems to account for this.

Basically what I'm trying to encourage is having to think about (and build) infrastructure around your turtle setups. You can't just plonk down a few turtles and set your mining swarm loose. You need to place down an ender modem to communicate with them over a long distance, need to make sure they stay within range of the repeater station. Maybe you need a GPS tower too?

I don't know, maybe this is just adding complexity for the sake of it. I'm very wary that what I'm suggesting here is making things actively harder for the player, maybe it's just too late to change things. Thoughts?

[^1]: While the [BitNet Communication Tower](https://github.com/theoriginalbit/MoarPeripherals/wiki/BitNet-Communication-Tower) is not the aesthetic I'm going for, I do really like the concept and mechanics - definitely an inspiration here.

RubenHetKonijn:
I do agree that this should happen, would also add some extra challenges to write code for. Only concern I have is world compatibility. Big servers have a ton of ender modems and that should be accounted for when rebalancing this.

SquidDev:
[...]

PR #1627, opened 2023-11-11
GPS - Use distance the server received.

fatboychummy:
I had a very random thought at 2 am today.

The GPS hosts receive a distance from the client, but then the hosts all send *separate* messages to the client. These separate messages may be sent at different times, meaning the player, turtle, or whatever, might have moved a slight bit in that time.

Thus, my thoughts are as follows:

1. Each host receives a distance at the same time, assumedly originating from the same location.
2. Each host can send that distance back to the client.
3. The client can use that distance instead to calculate its position. It *should* be more accurate, since the distances come from one message sent to all servers at the exact same time.

This is a really half-baked idea and fueled by way too much caffeine, but I thought I'd get the thought out there before I forgot about it. I have no idea how I'd test if this actually is happening (or if these changes even mitigate anything), but I recall having minor accuracy issues with GPS in the past. *In theory* (if my theory is right), this would resolve some accuracy issues.

### DIFF
--- a/projects/core/src/main/resources/data/computercraft/lua/rom/apis/gps.lua
+++ b/projects/core/src/main/resources/data/computercraft/lua/rom/apis/gps.lua
@@ -149,7 +149,7 @@ function locate(_nTimeout, _bDebug)
             if sSide == sModemSide and sChannel == CHANNEL_GPS and sReplyChannel == CHANNEL_GPS and nDistance then
                 -- Received the correct message from the correct modem: use it to determine position
                 if type(tMessage) == "table" and #tMessage == 4 and tonumber(tMessage[1]) and tonumber(tMessage[2]) and tonumber(tMessage[3]) and tonumber(tMessage[4]) then
-                    local tFix = { vPosition = vector.new(tMessage[1], tMessage[2], tMessage[3]), nDistance = tMessage[4] }
+                    local tFix = { vPosition = vector.new(tMessage[1], tMessage[2], tMessage[3]), nDistance = tonumber(tMessage[4]) and tMessage[4] or nDistance }
                     if _bDebug then
                         print(tFix.nDistance .. " metres from " .. tostring(tFix.vPosition))
                     end
--- a/projects/core/src/main/resources/data/computercraft/lua/rom/apis/gps.lua
+++ b/projects/core/src/main/resources/data/computercraft/lua/rom/apis/gps.lua
@@ -148,8 +148,8 @@ function locate(_nTimeout, _bDebug)
             local sSide, sChannel, sReplyChannel, tMessage, nDistance = p1, p2, p3, p4, p5
             if sSide == sModemSide and sChannel == CHANNEL_GPS and sReplyChannel == CHANNEL_GPS and nDistance then
                 -- Received the correct message from the correct modem: use it to determine position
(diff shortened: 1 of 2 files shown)


Wojbie:
I feel like this is something that would need to be well tested.. Like is there actually meaningful difference between distance that GPS host gest vs distance that GPS locate call gets?

SquidDev:
> Like is there actually meaningful difference between distance that GPS host gets vs distance that GPS locate call gets?

So there is potentially some delay between the original message being sent and the response being received. On most systems I'd expect this to occur in the same tick ([see some of the comments in `ComputerThread`](https://github.com/cc-tweaked/CC-Tweaked/blob/76968f2f28435ddd8cf1efe5dc7330410e87cc09/projects/core/src/main/java/dan200/computercraft/core/computer/computerthread/ComputerThread.java#L73-L87) about latency), but on a very over-burdened system it might be different.

That said, I still don't entirely understand the behaviour of GPS when the querying computer is moving. See also the discussion in #901, where I'm still very confused why `queueEvent(); pullEvent()` doesn't work to clear the queue.

Speaking of which, it might be interesting to do similar experiments like in #901 and see how it compares.

SquidDev:
I do have some concerns here about how this now requires a unique message for each requesting computer[^1], but that's probably niche enough it doesn't really matter.

[^1]: In the past, I have had built GPS servers which broadcast a packet once a tick, rather than in response to a request. This does help when you've got a lot of computers making GPS requests every tick.

fatboychummy:
> I do have some concerns here about how this now requires a unique message for each requesting computer[1](#user-content-fn-1-5d2995dfc4fe9a6c41a903845e259332), but that's probably niche enough it doesn't really matter.

Ooh, yes. I did mean to account for that. In theory the IDs wouldn't need to be anything fancy. Just enough that a collision doesn't occur once per tick essentially -- a randomly generated number. Having an ID could also be beneficial so you don't receive GPS pongs meant for other requests.

<details open><summary>There's an easier solution to receiving pongs meant for other computers, though...</summary>
<pre>
os.queueEvent("clear")
os.pullEvent("clear")
local x, y, z = gps.locate()
</pre>
</details> 

> In the past, I have had built GPS servers which broadcast a packet once a tick, rather than in response to a request. This does help when you've got a lot of computers making GPS requests every tick.

I'm curious if this would be worth implementing into the GPS program. Something like `gps busyhost x y z`.

fatboychummy:
Hmm, update on that. Including an ID would be a breaking change. Currently `gps.locate()` just sends out the string `PING`, and the server responds with its location. To include an ID we would need to change what `gps.locate()` sends, which will break any custom GPS servers.

SquidDev:
Sorry, I think I missed your latest reply! This is one of those cases where I'm not actually too worried about breaking custom GPS servers - as long as protocol changes don't cause crashes it's probably fine.

I'm was a little wary about how this will interact with anonymised GPS, but it occurs to me that while the computer ID is hidden, it's still fairly easy to track players with the current system.

I have mentioned some changes to long-distance GPS in #955 which I don't think affect this (compliant GPS servers can still choose not to broadcast the distance component), but worth mentioning.

djmattyg007:
Perhaps it would be better to just create a `gps2` wth a bunch of breaking changes, rather than trying to shoehorn everything into the existing code?

SquidDev:

PR #1647, opened 2023-12-04
Add guide to using Rednet

MCJack123:
This PR adds a guide for using Rednet, describing setup, sending and receiving messages, using protocols, using host lookup, and security suggestions + usage info.

The security section may be too dense even for people interested - maybe it's better to be split off somewhere else? It also mentions third-party projects, which I don't know whether you'd want that on the website. Otherwise, this should be a pretty comprehensive tutorial.

### DIFF
--- a/doc/guides/rednet_guide.md
+++ b/doc/guides/rednet_guide.md
@@ -46,10 +46,17 @@ Before using Rednet, your program needs to open it. Opening Rednet will tell the
 modem to start listening for messages from other computers. The
 [`rednet.open`](https://tweaked.cc/module/rednet.html#v:open) function takes the
 side of the modem to open as a string. On pocket computers, this will always be
-`back`. For example, if your modem is on the left side, you would use this code:
+`back`. For example, if your modem is on the left side of the computer, you
+would use this code:
 
 ```lua
 rednet.open("left")
+```
+
+You can also use this code to open every attached modem at once:
+
+```lua
+peripheral.find("modem", rednet.open) -- (Ab)uses peripheral.find's filter parameter to run rednet.open for each modem found.
 ```
 
 Always run this function before you do anything with Rednet - otherwise, it may
--- a/doc/guides/rednet_guide.md
+++ b/doc/guides/rednet_guide.md
@@ -1,0 +1,381 @@
+---
+module: [kind=guide] rednet_guide
+see: rednet Send and receive messages over modems.
+---
+
+<!--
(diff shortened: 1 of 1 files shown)


Lupus590:
I feel like encryption (if included at all) should be its own page.

doc/guides/rednet_guide.md
```diff
@@ -0,0 +1,388 @@
...
+<!--
+SPDX-FileCopyrightText: 2023 The CC: Tweaked Developers
+
+SPDX-License-Identifier: MPL-2.0
+-->
+
+# Transferring information wirelessly using Rednet
+Being able to send data between computers is an important feature for various
+programs, including turtle controllers and distributed banking systems.
+[modem](https://tweaked.cc/peripheral/modem.html) peripheral enables computers
```
SquidDev, line 15: We use the same markdown processor as the rest of the docs, so it should be possible to use ``[`modem`]`` here instead.

SquidDev (changes requested):
Thank you for working on this!

I think the content here is pretty good, but I'm not sure the structure is quite there right now. The order the information is presented is a bit odd - for instance, we're talking about sending files or tables, before the reader has even received a basic string!

My general approach to these sorts of docs is to tell a sort of story. We introduce a simple idea and let the reader use it. Then in the next section, we highlight a deficiency and/or possible extension, and use that as a motivation for introducing the next idea.

I'm taking some inspiration from the [diataxis model for tutorials](https://diataxis.fr/tutorials/) here, though I'm definitely not disciplined enough to follow it to the letter.

I really like the turtle control program, and think that might be a good example to develop throughout the tutorial. Using that, we we could split the guide into several steps the reader can follow:

 1. Setting up a modem and calling `rednet.open` on a single computer.
 2. Setting up a turtle with a modem, and sending messages to the turtle.
 3. Introduce a basic turtle control program. Don't introduce any new ideas here, but just the basic program.
 4. Then use that as a motivation for other features. For instance:
     - Using tables to send the result back to the controlling computer.
     - Protocols and host lookup for selecting which turtle to control.
     - Broadcast to control multiple turtles at once.

Mostly using this structure as an example to explain what I'm trying to do. Definitely don't feel like you have to follow this at all!


I think the cryptography section needs to be in an entirely different document. Someone just starting with rednet doesn't need to know that "Curve25519 is an elliptic curve asymmetric cryptographic algorithm" :p.

SquidDev:

Issue #739, opened 2021-03-16
[Feature request] Add some pathing system


sewbacca:
## What we currently have
- with `shell.setPath()` you can only set the path during runtime
- same with `shell.setAlias()`
- we have no way for setting `package.path` except from within the running program

## My idea
I'd like to have some general way for setting those variables and some more, if we ever need to, permanently, without using the startup file. Maybe we could use `.settings` for those information. If we would have a lot of different packages installed via `packman` or any other packaging software, each package, that brings a new `bin` or `programs` folder doesn't have to write into the startup file. I'm more interested in generalizing `package.path`, since there is no clean way for setting the `package.path` variable for every program, except by overriding the `shell` API.

## Proposal
1. Add a setting for `shell.path` and `shell.package.path` and default them to whatever we use currently. And maybe do the same with `shell.aliases` as a map.
2. Add a new file `.paths` which contains for each usecase e.g. `shell.path` a list of paths. To manipulate those paths, maybe we could introduce a new api called `path` or `paths`, where you can add and remove any path for any usecase (so uninstalling gets easier, if we remove a certain path (e.g. `/myprograms/bin`)

## Drawbacks
1. Editing a setting as a string might be complicated if you'd want to remove a certain path. First you would have to split the string, find the path that you want to remove and combine it again.
2. Adding a new config file is a bad sign in my opinion. You could evade that, by storing those tables directly into the `.settings` file. Still, you would need to add a new path API for that.

Lupus590:
Related: https://github.com/SquidDev-CC/CC-Tweaked/pull/152

Lupus590:
> something like `shell.packagePath()` and `shell.setPackagePath()`

I believe that both of those have been suggested already.

SquidDev:
Typically all this discourse happened on Discord, which makes it much harder to link to :/. Anyway, the most useful discussion happened mid December 2020 ([this is the tail end of the discussion](https://discord.com/channels/477910221872824320/477911902152949771/783628659697123330)).

> My take is that programs should be the ones responsible for managing their package environment, not the shell.
> 
> I'm not especially happy with where things stand right now, but I don't think everyone bunging things into the path is the solution either.
> What a lot of modern(tm) package managers (IRL, not in CC) do is just generate wrapper scripts which set up the environment and then launch the main program.

PR #1662, opened 2023-12-31
Add `shell.package_path` setting for `package.path` defaults

MCJack123:
This PR simply adds a new setting called `shell.package_path`, which stores the default `package.path` value. This allows users and programs to easily add global library paths without having to override `cc.require`. For instance, this can be used by package managers to add a central directory for shared libraries, instead of requiring programs to have their own copies of common software. The default settings will remain the same, and `package.path` changes in programs won't affect anything else globally.


Wojbie:
Related discussion where this was already denied #739

MCJack123:
I'm thinking of this more from a user-oriented perspective, where the user gets to determine what their own path is (and hopefully a program/package manager would respect the user's wishes), rather than as something for programs to abuse for their own paths. Of course, you can't force programs to not do bad things, but the way things are right now makes it very difficult to have common libraries for your own programs, and I think this is the simplest and most minimally invasive way to fix it.

----------

OTOH, maybe the reason CI failed is exactly why this is a bad idea. 🙃

If a global setting is a bad idea, then the next best way to resolve this issue is to add inheritance. This keeps the idea of using wrapper programs, but makes it a bit more flexible by allowing the original program to be run with `shell.run` as usual. It would resolve an annoying sticking point I had while developing Herobrine, where I needed to `shell.run` each program to get a new environment, but because `package.path` wasn't inherited, I had to add a line to append the library directory to every single program - this caught me when trying to run it uncompressed from a disk drive, as I had to change the path in every file.

Lupus590:
> If a global setting is a bad idea, then the next best way to resolve this issue is to add inheritance. This keeps the idea of using wrapper programs, but makes it a bit more flexible by allowing the original program to be run with shell.run as usual. It would resolve an annoying sticking point I had while developing Herobrine, where I needed to shell.run each program to get a new environment, but because package.path wasn't inherited, I had to add a line to append the library directory to every single program - this caught me when trying to run it uncompressed from a disk drive, as I had to change the path in every file.

Would `os.run` be better here?

PR #1678, opened 2024-01-10
Fix incorrect Lua list indexes

lonevox:
Without this change, tag lists in Lua look like this (key/value pairs):
1 first_value
2 second_value
3 third_value
0 fourth_value

This means that the last value isn't reached with ipairs().

With this change, they look like this and can be iterated with ipairs():
1 first_value
2 second_value
3 third_value
4 fourth_value


SquidDev:
Thank you!

PR #1679, opened 2024-01-10
fix: breaking_changes flattening link incorrect

znepb:
The link for "The Flattening" at https://tweaked.cc/reference/breaking_changes.html#mc-1.13 was incorrect, so I fixed it.


SquidDev:
Thanks!

Issue #596, opened 2020-12-04
Player parameter to monitor touch event


terickson001:
## Useful information to include:
 - Add an extra parameter to the `monitor_touch` giving the entity (Or player name) that triggered the event
 - Allows for more possibilities when multiple players are interacting with the same monitor

SquidDev:
This is one of those issues which is awkward, because I get that they're useful, but I don't think they fit with one of CC's main design choices - "don't break the fourth wall".

I think it's hard in this case to argue that monitor would know who is touching it.

Lupus590:
All monitors have fingerprint scanners built into the screen? My phone has one, granted it doesn't cover the whole screen but it is in the screen.

terickson001:
Understandable, I was just thinking about it usability-wise, not design-wise

PR #1683, opened 2024-01-15
feat: add player name to monitor_touch events

ryanrossiter:
Adds a player name argument to the `monitor_touch` event so that different player interactions can be distinguished.

I'm not sure if `player.getName().getString()` is affected by server nicknames, ideally it would return the real player name. Alternatively it might be better to use the player UUID?

### DIFF
--- a/doc/events/monitor_touch.md
+++ b/doc/events/monitor_touch.md
@@ -15,12 +15,13 @@ The [`monitor_touch`] event is fired when an adjacent or networked Advanced Moni
 2. [`string`]: The side or network ID of the monitor that was touched.
 3. [`number`]: The X coordinate of the touch, in characters.
 4. [`number`]: The Y coordinate of the touch, in characters.
+5. [`string`]: The name of the player who touched the monitor.
 
 ## Example
 Prints a message when a monitor is touched:
 ```lua
 while true do
-  local event, side, x, y = os.pullEvent("monitor_touch")
-  print("The monitor on side " .. side .. " was touched at (" .. x .. ", " .. y .. ")")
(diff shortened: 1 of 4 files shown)


SquidDev:
Thank you for the contribution! I'm afraid I'm not going to merge it; ComputerCraft/CC:T intentionally doesn't expose player names in any of its event (key, mouse, monitor interaction), as it feels like it veers too close to breaking the fourth-wall (see also #596).

There are some workarounds you can do here using an entity sensor peripheral (such as from [Plethora](https://modrinth.com/mod/plethora-peripherals) or [Advanced Peripherals](https://modrinth.com/mod/advancedperipherals)) - you can determine the look vector of nearby players, map it to a monitor coordinate, and check that corresponds to whoever has clicked the computer. I know there's some code kicking around to do this, I'll try to find it.

**Edit:** One example of this can be found [here](https://github.com/emmachase/bj3/blob/c32d97c66af75acdbe1f68c05719e11e627d33c3/modules/auth.lua), along with the [calibration script](https://github.com/emmachase/bj3/blob/c32d97c66af75acdbe1f68c05719e11e627d33c3/calibrateAuth.lua).

ryanrossiter:
Thanks for the prompt response (and maintaining this great mod)! That's a fair ethos, I understand why this may not be desirable.

Those mods look awesome I'll dig into them, thanks for sharing 👀

In the interest of adding a similar functionality to core CC:T, what would you think about adding a "Stylus" item which could be named and then the stylus name would be available on monitor touch events? I feel like that could be a happy medium.

PR #1684, opened 2024-01-16
Rewrite turtle upgrade modeller registration API

### LINKED ISSUE #1626: Crash when rendering turtle upgrades
### Minecraft Version

1.20.1

### Version

cc-tweaked-1.20.1-forge-1.108.1.jar

### Details

 I am playing Prominence 2 on a server with some friends and when looking up stuff on jei the game crashes, ive narrowed it down to this mod, i get exit code -1, ive already tried increasing the memory allocated to the game
 
heres the crash report:

[crashlog.txt](https://github.com/cc-tweaked/CC-Tweaked/files/13324634/crashlog.txt)

this is the crash recreated using prominence 2 v1.4 i tried updating cc to the latest version and i still get the same error

help a gal out pls

SquidDev:
Opening this as a PR for greater visibility. This is an API change (though the existing method is still around for backwards compatibility reasons), so mod makers will need to be aware of this.

---

Originally we exposed a single `registerTurtleUpgradeModellermethod` which could be called from both Fabric (during a mod's client init) and Forge (during `FMLClientSetupEvent`).

This was fine until we allowed upgrades to specify model dependencies, which would then automatically loaded, as this means model loading now depends on upgrade modellers being loaded. Unknown to me, this is not guaranteed to be the case on Forge - mod setup happens at the same time as resource reloading!

Unfortunately there's not really a salvageable way of fixing this with the current API. Forge now uses a registration event-based system, meaning we can guarantee all modellers are loaded before models are baked.

### DIFF
--- a/projects/common-api/src/client/java/dan200/computercraft/api/client/turtle/RegisterTurtleUpgradeModeller.java
+++ b/projects/common-api/src/client/java/dan200/computercraft/api/client/turtle/RegisterTurtleUpgradeModeller.java
@@ -1,0 +1,26 @@
+// SPDX-FileCopyrightText: 2023 The CC: Tweaked Developers
+//
+// SPDX-License-Identifier: MPL-2.0
+
+package dan200.computercraft.api.client.turtle;
+
+import dan200.computercraft.api.turtle.ITurtleUpgrade;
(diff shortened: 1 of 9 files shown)


laboriens:
model problem correct with the latest version

Issue #1688, opened 2024-01-19
Add wireless modem power control


hugeblank:
The premise is fairly simple, add a couple of additional methods onto the wireless modem that get/set a radius, in blocks, where messages can be transmitted. This could be done in a couple of ways:
1. Percentage, where 100% is the maximum distance the modem can transmit, and 0% is the singular block that the modem is in.
2. Radius, where the maximum is the longest range the modem can transmit without restriction (eg. 64 blocks), and the minimum is 0, again being the single block area where the modem exists.

Why? My primary use-case is inventory-contained networking for things like pocket computers, or plethora's neural interfaces. To be able to set the transmit radius to 0 would give wired modem-like functionality to all pocket computers. However, I think this could be useful beyond that, like in a base within a heavily populated spawn area. Limiting the modem's power could reduce the amount of computers that end up processing the message considerably.

This request partially depends on #955, unfortunately. Ender modems simply should not have the ability to capture a transmission from a wireless modem that can't even transmit more than a block away. The idea of a local in-inventory network also gets broken due to this.

hugeblank:
This report is a bit subversive, as my goal is in-inventory networking, and not necessarily to refactor/break modems. An entirely new peripheral could also fulfill the need I have and I would be just as happy. The way proposed above is the one that makes the most sense to me, given the confines of CC and the peripherals within.

Lupus590:
Maybe pocket computers should be able to talk to other pocket computers in the same inventory even without a wireless modem.
With some creative usage, one could simulate having pocket computers with multiple peripherals.

Wojbie:
I kinda question usefulness of 0 power messages? Like it would allow you to have 2 pockets talk with each other but then you have 2 pockets with wireless modems.. And that don't really increase ability of things you can do? Anything you could do with 2 wireless pockets you can do with just one wireless pocket.

And without ^ i kinda fail to see usefulness of reducing amount of computers that will see and then ignore your message. Feels like most users would just keep transmitting at full power and this would have low amount of actual use.

hugeblank:
> Anything you could do with 2 wireless pockets you can do with just one wireless pocket.

You're right. This should be a functionality that's inherent to pocket computers/inventory devices. That way one can do something like have 2 connected pocket computers: one with a speaker, and one with a modem. (Seemingly the same conclusion that was arrived at by #1406)

SquidDev:
[...]

Issue #1406, opened 2023-04-08
Pocket computers do not have many usecases without network access


fatboychummy:
# Prelude

Pocket computers, on their own, are extremely useless within Minecraft. They do have some extremely niche use-cases, but 9 out of 10 times it requires either a connection to the internet, or a wireless modem (or both). Because of this, I rarely see pocket computers being used with speakers or other peripherals. After all, what is the point of using a speaker if that means you cannot communicate with anything else? Why equip a block scanner if you can't transmit scan data to anything?

I understand people listen to music using pocket computers, and that is a valid use-case for pocket computers. However, it's very niche, and again requires network access (ATM-7/8 anyone?) to listen to anything more than a single song (if you can even manage to fit it onto the pocket computer, that is).

Without access to the `http` library, the amount you can do on a pocket computer is severely limited. In fact, even with internet access I think the only reason I've ever built a pocket computer is so I can build a neural headset with it. They just... Aren't worth it. Nothing cool or fun I can do with them without first setting up some "middleman" http server.

# Suggestion A
Pocket computers should have wireless capability by default, or the ability to equip two modules like a turtle can. Being able to equip a wireless modem and a speaker would be amazing as it would allow you to receive auditory alerts, or play some sort of "CC Radio", or more! You could remote-scan locations with a block scanner (or AP's geoscanner). Pocket computers would be overall much more useful.

# Suggestion B
Pocket computers should be able to communicate with other pocket computers in your inventory (maybe call it "NFC" or something?). This would allow you to have a wireless pocket computer that can receive and send messages, as well as a second pocket computer which controls some sort of device on your person.

You would be able to do all of the things from Suggestion A here, but it'd be a bit more inelegant and painful to set up.

9551-Dev:
very epic suggestion chummy.

DevWolf31:
Epic. Very good idea.

SirEndii:
Agreed

scmcgowen:
“Cannot reproduce”
It’s a suggestion my man

SirEdvin:
Returning to this issue, I can suggest "Suggestion C", I suppose.

## Suggestion C

New pocket upgrade (or reusing some of old ones) that can equip any other pocket upgrades and provide access to them.  Of course, with some limitation (like, not more then 3 or 5 equipped upgrades) and upgrades probably should be unique.

SurmanPP:
[...]

PR #1689, opened 2024-01-19
Basic NFC Modem functionality for Pocket Computer

hugeblank:
Attaches a 0-block radius "NFC" wireless modem to the top of the pocket computer. This is done to address the core issue presented in both #1406 and #1688.

TODO:
- [ ] ~~Rename peripheral type to "nfc"~~
- [x] Use per-player wireless network instead of the global one to prevent ender modem sniffing of NFC transmissions.
- [ ] ~~Rename transmission events to "nfc_message"~~

Feedback would be greatly appreciated on this to do list, specifically whether any/all of the items should actually be done.

### DIFF
--- a/projects/common/src/main/java/dan200/computercraft/shared/pocket/peripherals/PocketNFCPeripheral.java
+++ b/projects/common/src/main/java/dan200/computercraft/shared/pocket/peripherals/PocketNFCPeripheral.java
@@ -1,0 +1,30 @@
+package dan200.computercraft.shared.pocket.peripherals;
+
+import dan200.computercraft.api.peripheral.IPeripheral;
+import dan200.computercraft.api.pocket.IPocketAccess;
+
+import javax.annotation.Nullable;
+
+public class PocketNFCPeripheral extends PocketModemPeripheral {
+
+    public PocketNFCPeripheral(IPocketAccess access) {
+        super(false, access);
+    }
+
+    public void update(IPocketAccess access) {
+        setLocation(access);
+
+        var state = getModemState();
+        if (state.pollChanged()) access.setLight(state.isOpen() ? 0xf39d20 : -1);
+    }
+
+    @Override
+    public double getRange() {
+        return 0;
+    }
+
+    @Override
+    public boolean equals(@Nullable IPeripheral other) {
+        return other instanceof PocketNFCPeripheral;
+    }
+}
--- a/projects/common/src/main/java/dan200/computercraft/shared/pocket/peripherals/PocketNFCPeripheral.java
+++ b/projects/common/src/main/java/dan200/computercraft/shared/pocket/peripherals/PocketNFCPeripheral.java
@@ -1,11 +1,21 @@
+// SPDX-FileCopyrightText: 2017 The CC: Tweaked Developers
+//
+// SPDX-License-Identifier: MPL-2.0
+
 package dan200.computercraft.shared.pocket.peripherals;
 
+import dan200.computercraft.api.ComputerCraftAPI;
+import dan200.computercraft.api.network.PacketNetwork;
 import dan200.computercraft.api.peripheral.IPeripheral;
 import dan200.computercraft.api.pocket.IPocketAccess;
+import dan200.computercraft.core.util.Nullability;
+import net.minecraft.world.entity.Entity;
 
 import javax.annotation.Nullable;
 
 public class PocketNFCPeripheral extends PocketModemPeripheral {
+
+    private @Nullable Entity entity = null;
 
     public PocketNFCPeripheral(IPocketAccess access) {
         super(false, access);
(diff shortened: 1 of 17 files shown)


Wojbie:
#### Feedback

>  Rename peripheral type to "nfc"
>  Rename transmission events to "nfc_message"

Does it really need to implement as separate logic? I mean if this was just a normal wireless modem it would work with all already written code (and rednet).

> Use per-player wireless network instead of the global one to prevent ender modem sniffing of NFC transmissions.

We could achieve same effect just by not allowing messages from this modem to be heard on range over 1 block. Not sure how feasible that would be to invert this bit of message networking logic but it might be less complicated then creating completely new system.

#### Suggestion

I would suggest increasing range to 1 block to allow near range communication
This way while adding ability to network inside player inventory we also add ability to communicate between 2 players close to each other (like in a boat) [Sneakernet anyone?] or between player and wireless modem right next to it (commonly used for door locks in CC:T).

projects/common/src/main/java/dan200/computercraft/shared/pocket/peripherals/PocketNFCPeripheral.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public PocketNFCPeripheral(IPocketAccess access) {
+        super(false, access);
+    }
+
+    public void update(IPocketAccess access) {
+        setLocation(access);
+
+        var state = getModemState();
+        if (state.pollChanged()) access.setLight(state.isOpen() ? 0xf39d20 : -1);
```
Wojbie: Traditionally only main peripherals set light status. This might look odd if this one and back peripheral will both try to set different colours.

hugeblank: Good point, not sure how to address it though. Color blending? Blinking between the two colors?

hugeblank:
> Does it really need to implement as separate logic? I mean if this was just a normal wireless modem it would work with all already written code (and rednet).

This is true, in that case I won't implement list entries 1, and 3. 

> We could achieve same effect just by not allowing messages from this modem to be heard on range over 1 block. Not sure how feasible that would be to invert this bit of message networking logic but it might be less complicated then creating completely new system.

IMO, the objective here is strictly to have a means to transmit messages contained within the player's inventory. If it's desired to have out-of-inventory communication, then that's when you'd attach a wireless modem.

fatboychummy:
> I would suggest increasing range to 1 block to allow near range communication

Personally I'd say 2 blocks. Depending on where the range is checked from, you may end up in situations like the following: (assuming it checks from the feet)

SmartSelect_20240119_152448_Samsung Notes

Where gray is range of 1, purple is range of 2. If it checks from the feet, a computer at head-height beside the player would be out of range with a range of 1, even if you stood right beside it. Having a 2 block range would fix that.

(Please ignore my great drawing skills, not at a PC at the moment so all I have is my phone's notes app)

Wojbie:
> IMO, the objective here is strictly to have a means to transmit messages contained within the player's inventory. If it's desired to have out-of-inventory communication, then that's when you'd attach a wireless modem.

I see your point. And in the end if someone really wanted long range modem communication they could just carry a wireless pocket that would repeat messages between wireless modem and this inner modem. I just think that combining this with short range (1-block) mode from #1688 would provide more wide game play possibilities. Like sneakernet or pokemon style pocket gaming.

> Personally I'd say 2 blocks. Depending on where the range is checked from, you may end up in situations like the following: (assuming it checks from the feet)

If i remember right CC modems transmit from eyes of player.

hugeblank:
> I see your point. And in the end if someone really wanted long range modem communication they could just carry a wireless pocket that would repeat messages between wireless modem and this inner modem. I just think that combining this with short range (1-block) mode from #1688 would provide more wide game play possibilities. Like sneakernet or pokemon style pocket gaming.

I agree, I just don't think it's in the scope of this PR, and more-so something that should be added onto wireless modems like proposed in the referenced issue.

hugeblank:
Latest commit isolates networks on a per-entity basis, preventing sniffing from ender modems.

projects/common/src/main/java/dan200/computercraft/shared/pocket/peripherals/PocketNFCPeripheral.java
hugeblank: 5c67b6c is my attempt at addressing this, the light slowly cycles between the two active colors

hugeblank:
Demonstration of the light cycle effect:

https://github.com/cc-tweaked/CC-Tweaked/assets/16393543/298d10df-0044-40f2-aedb-331b0402f9e3

This is a bit extreme, but I thought it would be interesting to see what it looks like and field discussion from there.

PR #1692, opened 2024-01-23
Peripheral tools

### LINKED ISSUE #964: Turtles should be able to detect Equipped Pickaxes

ToxicGLaDOS:
This is an update to a previously closed pr I had to `mc-1.19.x`.

This fixes https://github.com/cc-tweaked/CC-Tweaked/issues/964 by making pickaxes (and all tools) peripherals in addition to tools. You can use peripheral.getType(side) and get the type of tool.

I haven't added any test cases for the new functionality like peripheral.call("left", "dig"), but I can do that if this seems like a reasonable approach.

Any comments or improvements are welcome!

### DIFF
--- a/projects/common-api/src/main/java/dan200/computercraft/api/turtle/TurtleUpgradeDataProvider.java
+++ b/projects/common-api/src/main/java/dan200/computercraft/api/turtle/TurtleUpgradeDataProvider.java
@@ -63,6 +63,7 @@ public abstract class TurtleUpgradeDataProvider extends UpgradeDataProvider<ITur
         private @Nullable Item craftingItem;
         private @Nullable Float damageMultiplier = null;
         private @Nullable TagKey<Block> breakable;
+        private String peripheralType = "";
         private boolean allowEnchantments = false;
         private TurtleToolDurability consumeDurability = TurtleToolDurability.NEVER;
 
@@ -93,6 +94,11 @@ public abstract class TurtleUpgradeDataProvider extends UpgradeDataProvider<ITur
          */
         public ToolBuilder craftingItem(Item craftingItem) {
             this.craftingItem = craftingItem;
+            return this;
+        }
+
+        public ToolBuilder peripheralType(String peripheralType) {
+            this.peripheralType = peripheralType;
             return this;
         }
 
@@ -162,6 +168,7 @@ public abstract class TurtleUpgradeDataProvider extends UpgradeDataProvider<ITur
                 if (consumeDurability != TurtleToolDurability.NEVER) {
                     s.addProperty("consumeDurability", consumeDurability.getSerializedName());
                 }
+                s.addProperty("peripheralType", peripheralType);
             }));
         }
     }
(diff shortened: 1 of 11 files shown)


Wojbie:
I believe this solution was already deemed to sadly break things in https://github.com/cc-tweaked/CC-Tweaked/issues/964#issuecomment-979312010 have you considered how this might break existing code that expects tool side to be able to connect to standalone peripherals?

ToxicGLaDOS:
> I believe this solution was already deemed to sadly break things in [#964 (comment)](https://github.com/cc-tweaked/CC-Tweaked/issues/964#issuecomment-979312010) have you considered how this might break existing code that expects tool side to be able to connect to standalone peripherals?

Just to make sure I understand; The issue is that if a turtle has a tool in it's left hand, and a peripheral block, like a monitor, to it's left it's not clear which should be returned by `peripheral.getType("left")`? Or put another way, `peripheral.getType("left")` could return "pickaxe" or "monitor" depending on what's next to the turtle, even if there's a pickaxe in it's left hand.

Wojbie:
I mean that currently we don't consider tools to be peripherals. Which means even if turtle has tool equipped at that side it allows one to still access normal peripherals at that side. 
143471839-d36e9b82-3202-4e5d-b7bc-16b5f6582065
See here - turtle is able to access left monitor cause it has tool on left side but on right side it has workbench so it can't access right monitor.
With your change it would also be unable to access monitor on left side too. What i am trying to say is that there has to be some consideration made on how this change may affect other parts of mod.

ToxicGLaDOS:
Right, I think that's what I was trying to get at. So I think there's a few options.
1. I think it makes sense for held items to act consistently and block access to peripherals on their sides. However, I realize this would be a pretty big change that would likely break a lot of existing scripts out there, so this might not be ideal.
2. We could check for a block on the side first and return that peripheral instead of the held item, this has similar issues to the first as it would break scripts that used to work if the turtle happened to pass a peripheral block.
3. We could have an entirely separate function for getting an equipped peripheral rather than a block peripheral. This seems like it might maintain compatibility best as we can keep the normal function how it is and simply add another one that only cares about equipped peripherals, but that interface is kind of confusing. Someone might ask "why does peripheral.getType("left") return "workbench", but I need to use the special peripheral.getEquipped("left") for a pickaxe?"
4. We could have a function which just works on tools. So rather than peripheral.getType("left") returning "pickaxe" we'd have a new API for tools and use something like tool.getType("left"). This should maintain compatibility and show that there's a distinction between tools and peripherals if that's something we'd like to keep.

This issue has been bugging me for a while, so I'd like to see a solution, but I'm not sure which of these (or another) solution  makes the most sense, but any of them would be an improvement for me. If we can decide on an approach I'd be happy to take a crack at implementing it and opening another pr!

fatboychummy:
Honestly, since tools/peripherals now have dedicated item slots, I'd much prefer if we could just `turtle.getLeftSlot()` and `turtle.getRightSlot()` or something. That way, especially when datapacks are loaded that allow for things like enchanted pickaxes to be equipped, we can get more info about the items we have equipped without having to unequip them.

I'll note this is just personal preference though, I'm not sure how this solution would compare to some other things. I just think having tools abstracted away into peripherals would make it much harder to get detailed information about the tool. Granted, it could make it better as well (if you exposed methods to give more information)...

ToxicGLaDOS:
@fatboychummy I like that idea, but I think that could run into some of the same issues that the original thread was talking about. Namely, in https://github.com/cc-tweaked/CC-Tweaked/issues/964#issuecomment-979090637 where @SquidDev says

> I'm not opposed to this in principle, but I do want to ensure people don't end up writing code like if `turtle.getLeftUpgrade() ~= "minecraft:diamond_pickaxe" then error("Need a pick") end`. Otherwise it's just going to be incompatible with people using other pick upgrades (such as those from other mods or added via CraftTweaker).

I think we need some way in the interface to know the capabilities of the tool. I'm not exactly sure how that would work though.

PR #1747, opened 2024-03-13
`speaker sound` command

fatboychummy:
## A quick checklist
   - No issues that I have seen.
   - I believe this is useful for some people who just want to test what certain sounds will sound like, without having to write their own wrapper program to do it. We already have a `speaker` program, why should we need to write a wrapper program or open the lua prompt?
   - No tests currently exist for the `speaker` program, and I am unsure exactly how I'd implement testing of it (see below!)

## Changes
1. Added `speaker sound <name> [volume] [pitch] [speaker]` command, which just plays the builtin sounds.
2. Fixed a small issue where not supplying a filename/url to `speaker play` would throw an `attempt to index nil value` error, instead it throws an error stating the usage.

## Thoughts/Questions
1. I originally wanted to split `speaker play` into `speaker play <audio/sound> ...`, but I am unsure if this is a good way to go, since everyone is likely used to the original `speaker play` command. `speaker sound` kind of sounds odd to me, but I couldn't think of a better subcommand that isn't already in use. Perhaps we could make it `speaker play sound` (just testing if the third argument is exactly `"sound"`) while otherwise keeping `speaker play` the same?
3. I debated on putting the optional `speaker` argument before `volume` and `pitch`, but eventually decided on putting it last, as I assume most people would use this program just to test what certain sounds sound like, and changing volume or pitch is more likely than selecting a specific speaker. Should this be changed?
4. If we want to make tests for `speaker.lua`, how should it be done? I could write easy mocks for `speaker sound` and `speaker stop`, but I am unsure how I'd test the raw audio processing in `speaker play`.


Wojbie:
It is kinda sad you can't add resource names auto complete cause CC:T don't know what resources are loaded. Or if it played correctly for that matter.

SquidDev:
Thank you!

PR #1763, opened 2024-03-24
Fix cc.image.nft.draw signature

cyberbit:
This corrects duplicate `xPos` parameters in the documentation for the `draw` function in `cc.image.nft`.


piprett (approved):

SquidDev (approved):
Haha, good spot. Thanks!

PR #1793, opened 2024-04-18
Update to 1.20.5

SquidDev:
A screenshot of Minecraft. There is a wolf wearing armour, demonstrating that this is 1.20.5. Above is a ComputerCraft monitor. The contents of the monitor is floating several blocks to the right, demonstrating that this is a long way from being stable.

I can get in-game and the tests pass, so must be ready to ship!

Notable changes:
 - Item NBT is gone, and replaced with data components. This is is a really good change, but also incredibly wide-reaching, and there will be bugs as a result.

   I have registered datafixers for ComputerCraft's items, which should allow worlds *just* using CC:T to migrate across (with some exceptions — we don't preserve tool data such as enchantments). Mods typically don't register datafixers, so in the general case you should expect worlds to break.

 - Accordingly, upgrades now store data using a `DataComponentPatch` rather than a `CompoundTag`. Existing upgrade data is stored in the `minecraft:custom_data` component.

   There's now a method to set new upgrade data (`setUpgradeData(DataComponentPatch)`), rather than a "mark dirty" method.

   I'm not entirely sure of the ergonomics of this change, especially when working with multiple values — it might be better to use a `DataComponentMap` here instead.

 - Networking is now done (mostly) with `StreamCodec`s. I love `StreamCodec`.

Still to do:
 - [x] Remaining datafixers (disks, printouts, treasure disks). Maybe also datafixers for turtle tools - I don't know if we can do this aside from poking the upgrade data and hoping it looks close-enough to a tool.

 - [ ] Migrate upgrades to codecs. I had a look at this in the 1.20.4 update and decided against it, but with the arrival of `StreamCodec`, it would make things easier! The trouble is we need some way to track the upgrade's id, which really implies a `Holder` like object.

   Vanilla has support for reloadable registries now (though not sure if they're synced to the client), so that might be worth looking at.

 - [x] Monitors.
 - [x] > All AbstractBlock public methods to be overridden were made protected and no longer marked as @Deprecated.

   We should apply this change to CC itself.

 - [ ] Tests for placing and picking computers and turtles, checking they preserve NBT.
 - [x] Fix client tests.
 - [ ] Sort out `ComputerConvertRecipe`. Honestly, looking at this and the overlay, and clear colour recipe, it feels like it would be nice to have a data-driven ingredient function system. But maybe a bit of a scope creep.


projects/common/src/client/java/dan200/computercraft/client/gui/ClientInputHandler.java
```diff
@@ -49,31 +49,31 @@ public void queueEvent(String event, @Nullable Object[] arguments) {
 
     @Override
     public void keyDown(int key, boolean repeat) {
-        ClientNetworking.sendToServer(new KeyEventServerMessage(menu, repeat ? KeyEventServerMessage.TYPE_REPEAT : KeyEventServerMessage.TYPE_DOWN, key));
```
SquidDev: We should cherry-pick these changes to 1.20.1

projects/common/src/main/java/dan200/computercraft/shared/common/AbstractContainerBlockEntity.java
```diff
@@ -26,9 +23,4 @@ protected AbstractContainerBlockEntity(BlockEntityType<?> type, BlockPos pos, Bl
     protected final Component getDefaultName() {
         return Component.translatable(getBlockState().getBlock().getDescriptionId());
     }
-
-    @Override
-    public boolean stillValid(Player player) {
-        return BlockEntityHelpers.isUsable(this, player, BlockEntityHelpers.DEFAULT_INTERACT_RANGE);
```
SquidDev: Reminder to double-check why we removed this. Possible it's just implemented in `BaseContainerBlockEntity` now? Is this really not the case in 1.20.1?

SquidDev:
Some other comments while I'm going through this

 - [x] Can we simplify any of the computer loot table stuff, now we have the `copy_components` loot function?
 - [ ] Do another pass on this PR and document all the new things (especially components).

> Vanilla has support for reloadable registries now (though not sure if they're synced to the client), so that might be worth looking at.

I think this is tricky to do without mixins. We *might* be able to do something with vanilla's dynamic registries. These still allow using datapacks (hopefully we can avoid the "experimental" warning), just not `/reload`ing them — you need to restart the world.

Issue #911, opened 2021-08-28
User made shell launch issue (bios improvement)


MAGGen-hub:
Why we always need to start working from the shell or multishell? Thats weird! It looks like... well... you can launch Windows and Linux (custom user shell) but to launch it you need to launch DOS(shell.lua / multishell.lua) first. I want this part of bios.lua to be changed. My solution is:
Insert this to bios.lua 939 line
```lua
settings.define("bios.custom_shell", {
    default = nill,
    description = "Allow user to launch his own shell or program that will run as core. Be carefull with this!",
    type = "string",
})
```

And change shell run like this:
```lua
-- Run the shell
local ok, err = pcall(parallel.waitForAny,
    function()
        local sShell
        local bSuccess = true
        if settings.get("bios.custom_shell")  then
			sShell = settings.get("bios.custom_shell")
        else if term.isColour() and settings.get("bios.use_multishell") then
            sShell = "rom/programs/advanced/multishell.lua"
        else
            sShell = "rom/programs/shell.lua"
        end
        bSuccess = bSuccess and os.run({}, sShell)
        bSuccess = bSuccess and os.run({}, "rom/programs/shutdown.lua") -- in fact I want to remove this too and replase with my custom shutdown with GUI "Goodbuy" (but i just use os.shutdown to disable this)
        
        -- Whops! Something went wrong (posible termination) and our custom shell errored! Better not to launch it again until player remove the problem.
        if not bSuccess then
        settings.unset("bios.use_multishell")
        end
    end,
    rednet.run
)
```

It also wold be cool if we can access bios settings and other with some key (like holding Del or F10) but I think it's to much for ingame computer)))

So can you impove bios in next versions of your modification or it's imposible for some reason? 
If thats imposible, please tell me how I can change it (with recourse pack) without messing up your mod code (I mean I don't want to change bios.lua inside your mod I need to "override" it but don't know how)?

Thank you for your attantion. Good luck in development!
P.S. (I also want to see terminals and monitors with higer resolution (like graphic monitors in OpenComputers) and unicode support (Want to see Russian lang in PC), but I know that it extrimly hard to made that, so... I just whant to know you plan to do them or not, or how can I do this by my own?)

SquidDev:
[...]

Issue #710, opened 2021-02-06
Early bios / startup hook | custom bios


aNickzz:
From what I've read, even with resource packs, changing the `bios.lua` is only possible by modifying the mod .jar file.

I'd really like to roll my own bios / os, and in the original CC I was "kinda" able to do so by [overwriting the `os.shutdown` function](https://github.com/aNickzz/dr-roboto/blob/814db6c703784db6dce5eab46ffad3932d437b86/computerRoot/startup#L74) that gets run at the end of `bios.lua` however its kinda hacky and it annoyed me a little bit knowing that `bios.lua:883` would always be the last stack frame in my stack traces.  So much so that I [checked for and hid](https://github.com/aNickzz/dr-roboto/blob/814db6c703784db6dce5eab46ffad3932d437b86/computerRoot/roboto/component/debug.lua#L31) that frame from my stack trace function.

There's also the issue of `bios.lua` irreversibly [hiding some native functions](https://github.com/SquidDev-CC/CC-Tweaked/blob/e85cdacbc58dedacb7fcbb119efd2b44db4bcdd6/src/main/resources/assets/computercraft/lua/bios.lua#L8).

> And as a side note, I noticed that neither the original nor the tweaked versions of CC make us of the `'mode'` parameter in the `load` function.

## How the feature/change should work

There are two feasible options that I can think of right now;

1. Much like the way the shell checks for `startup` or `startup.lua`, the computer could check for a custom `bios.lua` and load that instead of the built-in `bios.lua`
2. The same thing as option 1. but put the check as the first thing in the built-in `bios.lua` (before it loads the `expect` module).

The second option would end up similar to my hack of overwriting `os.shutdown` but it wouldn't overwrite the built-in types like `load` and its variants.

## Rationale

I'd like to implement my own top-level coroutine handler and operating system. Admittedly, much of what I'd like to do is already possible, however, something like this would make it a bit nicer and easier, as my previous implementation requires that you don't touch the normal "startup" file and instead use a `mystartup.lua`.

Wojbie:
> There's also the issue of `bios.lua` irreversibly [hiding some native functions](https://github.com/SquidDev-CC/CC-Tweaked/blob/e85cdacbc58dedacb7fcbb119efd2b44db4bcdd6/src/main/resources/assets/computercraft/lua/bios.lua#L8).

[...]

PR #1795, opened 2024-04-19
Add setting bios.shell_path

hugeblank:
Allows the user to directly set the path the bios uses to launch the shell. Setting `bios.shell_path` takes precedent over `bios.use_multishell`. Essentially a custom shell may choose to implement `shell` and `multishell` APIs, independent of whether use_multishell is set. This is loosely reflected in the description of the setting. Also some additional status checking has been implemented to allow for shell errors to be easily read before computer shutdown. 

My justification for adding this is that while [unbios](https://gist.github.com/MCJack123/42bc69d3757226c966da752df80437dc) exists, it is the logical extreme of what I imagine most players want. This PR is the spirit of unbios in a way that makes sense for CC, that advanced players can take advantage of (if I'm not mistaken mbs could benefit from this too, no?) Also servers presumably would ever so slightly benefit from computers that don't *need* a whole shell having the ability to start without one.

### DIFF
--- a/projects/core/src/main/resources/data/computercraft/lua/bios.lua
+++ b/projects/core/src/main/resources/data/computercraft/lua/bios.lua
@@ -697,6 +697,20 @@ if term.isColour() then
         type = "boolean",
     })
 end
+
+local sShell
+if term.isColour() and settings.get("bios.use_multishell") then
+    sShell = "rom/programs/advanced/multishell.lua"
+else
+    sShell = "rom/programs/shell.lua"
+end
+
+settings.define("bios.shell_path", {
+    default = sShell,
+    description = "The path the bios executes as the shell. This program is responsible for implementing the shell and multishell API, handling user input, and program execution.",
+    type = "string"
+})
+
 if _CC_DEFAULT_SETTINGS then
     for sPair in string.gmatch(_CC_DEFAULT_SETTINGS, "[^,]+") do
         local sName, sValue = string.match(sPair, "([^=]*)=(.*)")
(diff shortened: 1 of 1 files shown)


projects/core/src/main/resources/data/computercraft/lua/bios.lua
```diff
@@ -728,24 +742,24 @@ if fs.exists(".settings") then
...
+local shellOk
 local ok, err = pcall(parallel.waitForAny,
     function()
-        local sShell
-        if term.isColour() and settings.get("bios.use_multishell") then
-            sShell = "rom/programs/advanced/multishell.lua"
-        else
-            sShell = "rom/programs/shell.lua"
+        shellOk = os.run({}, settings.get("bios.shell_path"))
+        if shellOk then
```
Wojbie, line 750: Is there a reason this part of logic is changed and extra variable is checked?

hugeblank: yes, when the shell errors os.run emits false, and does not error. this was never caught before and so after 1 second of the error displaying, the computer shuts down. The only reason we see the error in the first place is because we're afforded that 1 second by the shutdown rom program saying goodbye. My change allows a user to see the error at their leisure, where the "press any key" is.

hugeblank: This actually runs a little deeper, in that we don't know on what terminal the error from os.run was rendered on. Ideally we'd redirect to the native one, like we do when the whole parallel loop errors. I almost wish os.run gave the error string like pcall does so we could just `printError` it like we do with the parallel loop.

edit - os.run is defined in bios, lucky me!

SquidDev:
This was previously discussed in #710 and #911, and it's still not clear to me if this is actually useful. I understand the instinctual desire for this, but I don't actually think it ends up solving any practical problem.

hugeblank:
The utility for me is that I don't have to deal with my own shell (or programs that don't require a shell) being layered atop an existing shell. It just feels like an extra step that should have the ability to be removed, especially when it's dead simple to implement, does not break compatibility, and is entirely up to the user to take advantage of.

Wojbie:
How does the dead shell impacts anything? Assuming multishell is off and you called `shell.exit()` in startup code that then initiated your shell it no longer does anything and will exit once your code exits.

EDIT: I guess it is there in call stack but does that really impact anything ?

hugeblank:
> Assuming multishell is off and you called `shell.exit()` in startup code that then initiated your shell it no longer does anything and will exit once your code exits.

Startup programs wouldn't even be ran, and the shell API would not exist. Those are both things the shell does. Basically any `shell.<...>` setting would become something the user may (or may not) implement in their custom shell.

EDIT:
#1796 posed an interesting problem regarding startup files that I spent far too long thinking up a couple solutions for. Like I mention in this comment before the edit, startup programs wouldn't even be ran, and as such players wouldn't even be faced with having to deal with the limitations and quirks of the vanilla shell. This is a benefit to the option of having a clean slate for a custom shell.

projects/core/src/main/resources/data/computercraft/lua/bios.lua
hugeblank: latest commit extracts os.run logic so we're able to print the error on the correct terminal

PR #1823, opened 2024-05-08
Changed Heart Character (-3 pixels)

avigailxo:
Suggestion for changing the heart ("\3") character in the default CC font.
Here are the changes I made:
Heart Character Rework Large


SquidDev:
Sorry for the delay in looking at this! This is a bit hard to decide — most of the font is just taken from Minecraft itself, including the heart icon, which I'm reluctant to diverge from. However, I definitely agree that this is a nicer design.

Let's merge this for now, and see if we get any feedback one way or another :).

SquidDev:
Hrmr, not sure what happened here, but it looks like the font got all moved around when you were making this change for now. Have reverted for now in e4236824d7c375b79772fe23abfe66e5a6e71bc0, but will take another look a this when I have some time.

PR #1860, opened 2024-06-13
Update to Minecraft 1.21

SquidDev:
I forgot about the breeze, so instead there's a much more boring screenshot:

A screenshot of a CC terminal in a superflat Minecraft world. `_HOST` as been entered in the Lua REPL, showing it is running in MC 1.21.

## Still to do:
 - [ ] NeoForge support. I think this should be pretty straightforward, but I want to tear out NeoGradle first, and replace it with an alternative Gradle plugin, which might require a bit more work.
 - [ ] `turtle.attack`. This whole code is commented out right now, and not sure quite how to handle this without just copying `Player.attack` again :/.
 - [ ] Testing. Especially around rendering and `turtle.craft` — the changes there are pretty simple, but easy to get wrong.

## API changes
 - `IMedia.getAudioTitle`/`IMedia.getAudio` were merged into a single method returning a `Holder<JukeboxSong>`.

### DIFF
--- a/buildSrc/src/main/kotlin/cc/tweaked/gradle/ForgeExtensions.kt
+++ b/buildSrc/src/main/kotlin/cc/tweaked/gradle/ForgeExtensions.kt
@@ -31,9 +31,9 @@ fun JavaExec.setRunConfig(config: Run) {
     environment(config.environmentVariables.get())
     systemProperties(config.systemProperties.get())
 
-    config.modSources.get().forEach { classpath(it.runtimeClasspath) }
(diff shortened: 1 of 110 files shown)


vico93:
Was the `turtle.attack` broken for 1.21? Or the not copying `Player.attack` a choice for that version?

SquidDev:
Player attacks got a massive rewrite in 1.21, to account for the new enchantments. We need to update the turtle's attacking code to match that.

SquidDev:
Pushed to [`mc-1.21.x`](https://github.com/cc-tweaked/CC-Tweaked/tree/mc-1.21.x) branch.

PR #1888, opened 2024-07-02
Transfer item data when adding upgrade to pocket computer by crafting (i.e. enchantments)

CaptainSqrBeard:
## About
This PR now makes `PocketComputerUpgradeRecipe` transfer item data to crafted pocket computer by cloning `ItemStack` and adding upgrade to it.

## Why
There is an issue that removes enchantments from pocket computers when upgrade is attached through crafting.
Pocket API also have functions that equip and unequip upgrades that doesn't remove enchantments.

### How pocket computer can be enchanted
- Pocket computer can be enchanted with custom enchantment that can apply on any item. 
- Any enchantment to pocket computer can also be applied with anvil while being in creative mod
- Commands like `/give`

## Image
image

### DIFF
--- a/projects/common/src/main/java/dan200/computercraft/shared/pocket/recipes/PocketComputerUpgradeRecipe.java
+++ b/projects/common/src/main/java/dan200/computercraft/shared/pocket/recipes/PocketComputerUpgradeRecipe.java
@@ -84,7 +84,9 @@ public final class PocketComputerUpgradeRecipe extends CustomRecipe {
         var computerID = itemComputer.getComputerID(computer);
         var label = itemComputer.getLabel(computer);
         var colour = itemComputer.getColour(computer);
-        return itemComputer.create(computerID, label, colour, upgrade);
+        var newItemComputer = itemComputer.create(computerID, label, colour, upgrade);
+        newItemComputer.addTagElement("Enchantments", computer.getEnchantmentTags());
+        return newItemComputer;
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


Wojbie:
This is a bit confusing. CC:T Pockets are not enchantable. Why would those recipes need to care about this?

CaptainSqrBeard:
> This is a bit confusing. CC:T Pockets are not enchantable. Why would those recipes need to care about this?

I explained that in PR description

Wojbie:
Sounds like having enchantments on pocket computer is a bug that should be fixed then.

SquidDev:
I think there are definitely some enchantments it makes to have on CC items (for instance Holding(?), which causes the item to stay around after death).

That said, rather than special-casing the `Enchantment` tag, I think the better thing to do is just clone the whole item stack and then call `PocketComputerItem.setUpgrade` (it's probably worth auditing all usages of the various `TurtleItem.create`/`PocketComputerItem.create` methods, to see if others can be replaced, e.g. in turtle upgrade or overlay methods)

This is what we do on MC 1.21 ([example](https://github.com/cc-tweaked/CC-Tweaked/blob/0e4710a9566235b8066f9fddb519229509b32da0/projects/common/src/main/java/dan200/computercraft/shared/turtle/recipes/TurtleUpgradeRecipe.java#L122-L125)), so should be fairly easy to backport. Will probably end up a bit of a merge nightmare, but we'll just have to live with that :).

CaptainSqrBeard:
Thanks for the reference! I'll change my code later

CaptainSqrBeard:
Done!

SquidDev:
Thank you!

PR #1906, opened 2024-07-20
Fix `cc.strings.wrap` Hangs Program If 0 Width

### LINKED ISSUE #1905: `cc.strings.wrap` Hangs Program If 0 Width
### Minecraft Version

1.20.1

### Version

1.109.3

### Details

Submitted the issue before I was done typing accidentally, whoops. The `wrap` function in `cc.strings` causes a too long without yielding if you pass 0 to it. One of my programs did this accidentally in development. Expected behavior would be either an assertion, return the string as provided, or return an empty string. 

image

Lupus590:
closes https://github.com/cc-tweaked/CC-Tweaked/issues/1905

We now require the width to be 1 or greater


SquidDev:
Thank you! Have cherry-picked this in manually, as I made a couple of other changes to `cc.strings`.

PR #1923, opened 2024-08-01
Add normal modem to all pockets

Wojbie:
Allow me to preface that this is a draft because this is most likely not a proper way to implement this. This is just created as showcase/experimentation space to allow interested parties to see how this would feel.

image

### What this is:
This basic bodge permanently attaches basic modem to top slot of all pockets. For base mod it means that one can attach second basic modem to double amount of channels they can open, ender modem to allow better range/interdimensionality or a speaker to add sound while still having basic functionality of basic modem installed. With other peripheral mods it means one can use peripherals from those mods on pocket without loosing network access. This does keep pockets as one swappable peripheral, it just add a unswapable internal top modem that increased functionality of pockets.

This is example of Suggestion A from https://github.com/cc-tweaked/CC-Tweaked/issues/1406 in same vein that https://github.com/cc-tweaked/CC-Tweaked/pull/1689 is a example of Suggestion B.

Despite fact that this is not proper way to do it it seems to be working correctly from tests i have preformed. 
image
image

As a side note that highly amuses me:
Turns out [colony4cc](https://github.com/uecasm/colony4cc/blob/trunk/src/main/java/nz/co/mirality/colony4cc/mixin/MixinPocketServerComputer.java) have been doing something similiar for last 4 years so this is not exactly unexplored solution.

Additionally i would like to thank SquidDev in https://github.com/cc-tweaked/CC-Tweaked/discussions/1916 for pointing out to me this is possible to bodge in this simple way. 😄 

EDIT: And if any of you wanna to test how this PR works you can use jar from this PR checks to try it out. https://github.com/cc-tweaked/CC-Tweaked/actions/runs/10193099406/artifacts/1763405706

### DIFF
--- a/projects/common/src/main/java/dan200/computercraft/shared/pocket/core/PocketBrain.java
+++ b/projects/common/src/main/java/dan200/computercraft/shared/pocket/core/PocketBrain.java
@@ -14,6 +14,7 @@ import dan200.computercraft.shared.computer.core.ComputerFamily;
 import dan200.computercraft.shared.network.client.PocketComputerDataMessage;
 import dan200.computercraft.shared.network.server.ServerNetworking;
 import dan200.computercraft.shared.pocket.items.PocketComputerItem;
+import dan200.computercraft.shared.pocket.peripherals.PocketModem;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.resources.ResourceLocation;
 import net.minecraft.server.level.ServerLevel;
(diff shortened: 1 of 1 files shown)


uecasm:
Hi, author of colony4cc here.  My evil hack doesn't add a modem to all pocket computers; it applies only when a specific "composite" peripheral is attached.  Basically I made a specific upgrade that's intended to behave identically to a modem plus my regular peripheral, which it does by attaching the modem to the back as normal and my peripheral to the bottom.  (There's another evil hack to prevent the modem from hijacking the pocket computer's light.)

As such, it's sort of like Suggestion A or C in #1406.  I'd definitely prefer a less-hacky solution for this sort of use case. 😀 

----

One caveat with the implementation in this particular PR is that it's only a standard modem, not an ender modem; whereas I wanted to allow both options.  You could perhaps base that on the computer type?  Though the ender modem is more expensive than an advanced computer.

Wojbie:
> One caveat with the implementation in this particular PR is that it's only a standard modem, not an ender modem; whereas I wanted to allow both options.

That is explicit choice on my part as i just wanted to add most basic standard wireless networking option as baseline so anything else is a upgrade choice by user. If this idea somehow gets actually implemented i feel it would make most sense to keep it this way. As whole point was just to add basic networking capability.

PR #1924, opened 2024-08-04
Fix getNextVersion

viluon:
Fix `getNextVersion` to work properly with stuff like `1.2.3-SNAPSHOT`. It would previously crash, trying to parse `3-SNAPSHOT` as an integer.


SquidDev:
Thanks!

Issue #825, opened 2021-06-12
Correctly scale computer GUIs with custom sizes


SquidDev:
Since #112 was closed, we've had the ability to reconfigure the dimensions of computer GUIs using the config file. However, it's always be a little janky, as computers do not scale, and so too large sizes do not fit on the screen.

We should automatically resize the computer GUI where needed. I suspect we'll need to make sure fonts are still readable (and not scaled badly) when doing so.

SquidDev:
Taking this off the 1.97.0 milestone - will hopefully look another day.

My initial approach was just to scale the terminal by 0.5x, and keep the GUI the same ratio. While this works, it ends up looking rather ugly. I think we need to scale the whole GUI when doing this, and somehow deal with all the scaling issues.

Or alternatively, just don't fix it and hope people don't use the config options on the small GUI mode :p.

PR #2000, opened 2024-10-24
Render pocket computers in tooltips

SquidDev:
I was looking at the 1.21.3's bundle code, and realised it'd be quite easy to render a preview of a pocket computer in the tooltip, much like bundle previews do.

I'm not sure this is a *good* feature — I think it's cute and neat, but I'm not actually sure it's useful.  It definitely takes up more screen space than I'd like.

In the current code, the computer renders at half size:

A screenshot of the Minecraft inventory GUI. The user is hovering over a pocket computer, and the tooltip shows the a preview of the pocket computer's terminal (playing the redirection game).

We could also try at quarter size:

A similar screenshot to the above, but the pocket computer preview is half the size. The screen shows a paint program, where I've drawn another pocket computer running the paint program (and so on one more time). I'm so funny.

My main concern here is the same issue that's blocking #825. When going down to fractional scales (e.g. when using a small GUI scale), you get all sorts of sampling issues, leading to unreadable text and weird Moiré patterns. Given the main purpose of this would be to read the text of the computer, it probably is a blocker.

The same screenshot as above, but at a smaller GUI scale. The text is unreadable, and there's Moiré patterns on the paint program's background. A similar version to the previous one. The paint program's "Press Ctrl to open menu" is entirely unreadable.

### DIFF
--- a/projects/common/src/client/java/dan200/computercraft/client/pocket/ClientPocketComputers.java
+++ b/projects/common/src/client/java/dan200/computercraft/client/pocket/ClientPocketComputers.java
@@ -56,4 +56,8 @@ public final class ClientPocketComputers {
         var id = PocketComputerItem.getInstanceID(stack);
         return id == null ? null : instances.get(id);
     }
+
+    static @Nullable PocketComputerData get(UUID id) {
+        return instances.get(id);
+    }
 }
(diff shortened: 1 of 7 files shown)


zyxkad:
Definitely should add some client configs for render this or not and use what scale to render it

Lupus590:
Maybe the computer should be able to change the text tooltip instead? Kind of like a pager display then.

vico93:
Really liked it

SquidDev:
> My main concern here is the same issue that's blocking #825.

I don't think there's really a good solution to this, so let's put this on hold for now.

PR #2002, opened 2024-10-26
Redstone relay

### LINKED ISSUE #1624: Add a separate redstone peripheral
- Let the wired modem be able to recognize redstone signals in the same way as the computer recognizes them.

By default, the computer has the ability to identify the redstone signal from each side, there are 6 outputs in total. Due to the features of minecraft redstone, the circuits turn out to be quite large and it is very inconvenient to connect them into one computer, and even if you connect them, it is only 6 signals, which may seem insufficient. Because of this, it is necessary to build complex systems with several computers that exchange data, instead it would be great to be able to connect a wired modem and use it as a "splitter" of redstone signals.

Ultimately, this will simplify interaction with redstone, and will avoid unnecessary use of additional computers, because in fact they are used only to identify redstone signals.

Example of use: Let's say we are building schemes on several floors of a building, whether it is a multi-storey farm or an architectural idea. Now we can get 4 signals for the current floor and only 1 signal for the upper and lower floors, which is definitely not enough. However, with this update, we will be able to conduct a wired modem for each desired floor and get additional redstone slots.

SquidDev:
This adds a separate redstone peripheral, as described in #1624.

There's still a lot to do here, but felt it was worth making this public and getting some feedback here.

 - The peripheral is currently called "Redstone Control". I'm not wild about the name, so here's some alternatives I considered:
   - Redstone Control/Redstone Controller: My main worry here is that this suggests it only outputs redstone signals, not that it can receive them too.
   - Redstone IO: this is what OC uses. It makes sense, but I don't just want to steal the name from another mod.
   - Redstone Integrator: This is what Plethora does (my mod so I can steal from it :p). However, Plethora intentionally went for kinda pretentious(?) names (i.e. Kinetic Augment, Introspection module) — I'm not sure this quite fits the vibes of CC itself though.

   Not sure, feedback welcome!

 - I'm using @TheStraying11's proposed textures from #1624. These originally came in an on/off variety for each face. I've decided against this for now, and the block is always "on", as doing the models for this is really awkward. We could just have the whole block be on/off, rather than each face. Not sure!

 - The block is directional and uses relative sides ("left", "right", etc...), like the `redstone` API. We use the same direction mapping as turtles, where left/right are relative to the direction the block is facing (like turtles), rather than the player facing the block (like computers).

## Still to do
 - [x] Tests
 - [x] Documentation


Karotte128:
I really like this feature!
Another possible name is "Redstone Port", but not sure if this name is good.

Wojbie:
One alternative naming scheme worth considering could be Redstone extender or Redstone relay? As it extends/relays computer redstone control over peripheral network.

SquidDev:
Oh, "Relay" is a good one!

MCJack123:
I know this was already merged, but could we get a side parameter on the `redstone` event? It can be frustrating to have to check every single input on all connected relays to see if one side changed, especially if there's a lot of inputs - it would be much simpler code-wise if the input change could be narrowed down to a single relay.

(For my specific use case, I'm wrapping peripherals with a driver that limits peripheral events to programs that request them. Not knowing which event corresponds to which peripheral means I have to figure it out myself through a lot of peripheral calls, or the event needs to be sent to all redstone devices, regardless of whether that peripheral changed.)

Wojbie:
Question is how its implemented.
Correct me if i am wrong but currently even if multiple changes happened at once, they would all generate only one `redstone` event right? Or would there be several of them?

- If its just one per all changes, then adding name/side argument would mean increasing amount of events generated on busy redstone control system. It may be problematic due to event queue size?
- But if its already generating multiple of them, then adding names/sides it would add some clarity.

I am about 90% sure its just one event for all changes.

fatboychummy:
Looking through the code, it looks to me like it is one event per relay. Which, if the relay included its own name (and assuming I've read things correctly), would not increase the number of events. However, this would decrease the amount of `getInput` calls required per event from `n*6` (`n=#relays`) to just `6`. This would keep the current limit of 256 redstone relays changing any number of sides per tick (1536 total redstone changes per tick).

Including the side of the relay that changed as well *would* increase the number of events received, but could then also include the previous and current redstone values, thus eliminating the need to even call `getInput` at all. This would, however, limit a network to having 256 redstone changes per tick, which is still a decent amount (in my opinion at least). However, since the computer would now be yielding for *every* change, I don't know if this would have any performance impacts for the increased resumes.

My overall thoughts? Having just the relay name in the event would be a significant upgrade, but going to per-change events may be overkill.

Lupus590:
I would also be surprised if (in most setups) any given restone relay has more one side being used. While it would make usercode more complex, one could map which side of each relay is being used and thus not need to check all 6 sides of that relay when its redstone event is pulled.

So yeah, having a second return value on redstone events for which relay queued that event would be useful.

zyxkad:
> My overall thoughts? Having just the relay name in the event would be a significant upgrade, but going to per-change events may be overkill.

I was also considering this issue when contributing AP and ends up with build a table with peripherals' names as the keys.
But I never see any CC:T's event defines a table, so it may not suitable with CC:T's coding style.

[SwarmEventDispatcher.java#L27](https://github.com/IntelligenceModding/AdvancedPeripherals/blob/0878c99744de192d7cbb01494c694e8d28e9ebf1/src/main/java/de/srendi/advancedperipherals/common/util/SwarmEventDispatcher.java#L27)
[RedstoneIntegratorEntity.java#L59](https://github.com/IntelligenceModding/AdvancedPeripherals/blob/0878c99744de192d7cbb01494c694e8d28e9ebf1/src/main/java/de/srendi/advancedperipherals/common/blocks/blockentities/RedstoneIntegratorEntity.java#L59)

PR #2037, opened 2024-12-21
Show error in wget on http.get error

tizu69:
This is the most minor change one could ask for, but removes the need to open a lua shell to check why a file download failed.


SquidDev:
Oh gosh, hadn't realised we didn't do this already. Thank you!

tizu69:
whar

Issue #1148, opened 2022-08-20
Some way to notify player through pocket computer with modem


IvoLeal72:
Basically a way for a pocket computer with a modem to call the attention of the player, to be able to make a notification system for when a farm stops working or a super smelter is finished, etc.

I have 2 possible solutions, either being able to add a speaker and a modem to the same pocket computer, or some special method for pocket computers that makes the item blink in toolbar or something like that, like what was discussed here http://www.computercraft.info/forums2/index.php?/topic/21910-pocket-computer-ring-or-push-notification/

Void-Seeker:
I personally think every computer at least should be able to beep, like the real ones do even without sound card. And Pockets should have speaker by default.

Lupus590:
What has been suggested on the forums is to open and close the modem channels, this will cause the front light to blink and thus proxy as a notification light. However, closing modem channels does mean that you may miss some messages, you can resolve this by having a computer acting as a message proxy, with all messages going to that and the pocket computer fetching them in a controlled manner, or you could set up message receipts so that the sender will resend their message to the pocket computer if the pocket computer doesn't confirm that it received the message.

IvoLeal72:
My idea would be something more integrated with the mod, than doing a weird workaround with the modem that WILL sooner or later lead to problems. The beep idea sounds good to me, but some way of it blinking in the toolbar, would also be cool.

IvoLeal72:
Hello, after thinking about this, the @Void-Seeker idea seems great to me, I have seen this beep possibility mentioned in various places in the community. Would resolve my issue.

walksanatora:
My personal solution would be to allow advanced pocket PC's to dual equip peripherals
Making it more worth the upgrade
Like the primary peripheral in back
And Mabey the other at top

IvoLeal72:
That would solve the issue too, but probably too much OP for a pocket

OttersMeep:
perhaps a basic whistle chip in the pocket computers, and regular computers that can play a tone at a fixed hertz (eg: 500 HZ)

walksanatora:
Give us os.beep(hz)
Add a POST beep to bios.lua clearly

archer2o2x:
Personally love the idea of being able to make a pocket computer beep, maybe with a fixed tone so that people have to get creative if they want multiple distinct alarms.

IvoLeal72:
[...]

PR #2047, opened 2025-01-03
Added notify function to pocket api

IvoLeal72:
pocket.notify() makes the led blink green and then return to previous color, and beeps 3 times, only to the player that is holding it and broadcast if no player is holding it

Possible solution for #1148

### DIFF
--- a/projects/common/src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
+++ b/projects/common/src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
@@ -10,9 +10,15 @@ import dan200.computercraft.api.pocket.IPocketAccess;
 import dan200.computercraft.api.pocket.IPocketUpgrade;
 import dan200.computercraft.api.upgrades.UpgradeData;
 import dan200.computercraft.impl.PocketUpgrades;
+import net.minecraft.core.Holder;
 import net.minecraft.core.NonNullList;
+import net.minecraft.network.protocol.game.ClientboundSoundPacket;
+import net.minecraft.sounds.SoundEvent;
+import net.minecraft.sounds.SoundSource;
+import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.block.state.properties.NoteBlockInstrument;
 
 import javax.annotation.Nullable;
 import java.util.Objects;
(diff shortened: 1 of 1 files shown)


Wojbie:
I do believe that pockets not in player inventory are supposed to shut down, is no-player logic supposed to be some kind of future proofing?

IvoLeal72:
I forgot about that haha. That is completely true, it's fun to make it notify and throw it, but kinda useless yeah, I can remove it

projects/common/src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
```diff
@@ -106,6 +112,46 @@ public final Object[] unequipBack() {
...
+
+
+    @LuaFunction("notify")
+    public final void doNotify(){
+        var previousColour = pocket.getLight();
+        try {
+            for (int i = 0; i < 3; i++) {
+                beep();
+                pocket.setLight(0x00CC00);
+                Thread.sleep(250);
```
zyxkad, line 146: Instead of block the thread, can you use timer/ticker?

zyxkad: Or just do this at client side

SquidDev:
Thank you for looking into this! The main thing stalling #1148 right now is that I'm not sure whether I actually want this feature (and if so, what the interface/functionality should be). I suspect the answer here is to go the #1406 route and add multiple upgrade slots, though that's blocked on a new Minecraft version.

Given the whole uncertainty around this, I'm afraid I'm going to close this.

IvoLeal72:
Out of curiosity, why is it blocked on a new mc version? Breaking change?

projects/common/src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
IvoLeal72: Hello, I know the PR is closed, but I'm curious on how to do this, can you give me some help, or forward me to some documentation/tutorial? I'm completely new to minecraft modding, though I have some years of experience with java

SquidDev:
> Out of curiosity, why is it blocked on a new mc version? Breaking changes? 

Yep! This'll require some changes to the Java API.

projects/common/src/main/java/dan200/computercraft/shared/pocket/apis/PocketAPI.java
zyxkad: Basiclly what I mean is async programming. You should register a listener when tick executes, and do actions at specific time.

IvoLeal72: I know how to do that with java standard timers, I'm asking about the forge standard way, both for async and the clientside, how can I make it run client side

SquidDev: So what I'd normally do instead is you have some state on the pocket computer (either client-side or server side) which counts the number of ticks since the last "notify" call, and then drive the light colour based on that.

So something like this:

```java
class PocketComputer { // PocketServerComputer on the server, or PocketComputerData on the client.
  private int notifyTimer = 0;

  public void notify() {
    notifyTimer = 60;
  }

  public void update() {
    if(notifyTimer > 0) notifyTimer--;
  }

  public int getLightColour() {
     var timer = notifyTimer / 12;
     return timer % 2 == 0 ? lightColour : 0x00CC00;
  }
}
```

The `update()` then would get called every tick (via Forge's `TickEvent.ServerTickEvent`/`TickEvent.ClientTickEvent`).

In order to do this on the client, you'd need to send a custom packet from the server to the client. This is a bit of a faff, so I probably wouldn't bother, and just do this server-side.

This would require making the original `pocket.notify()` method non-blocking, but I think that's fine.

PR #2097, opened 2025-02-09
Hints for runtime errors

SquidDev:
Only two years [after originally suggested](https://github.com/cc-tweaked/CC-Tweaked/pull/1320#issuecomment-1418290228), this adds hints for misspelled globals and table keys:

A CC computer. The user has tried to run redstone.getinput(), and we've printed a "Did you mean getInput" after the error.

This is heavily WIP. For one, the above screenshot is heavily cherry-picked. The suggestions are pretty broken in many cases — I'd forgotten how much I hate string distance metrics. We're currently using Jaro–Winkler, but we probably want to switch to (Damerau–)Levenshtein, and maybe play around with weights and whatnot.

Similar to the above, but redstone.getoutput suggests both getOutput and getAnalogOutput. Even worse, redstone.getouput has no suggestions.

Perhaps more importantly, I am incredibly unsure about the implementation here. The core logic is written in Java — we inspect (and partially evaluate) the Lua bytecode to find the failing operation and its operands.

There's a couple of things I'm not a fan of here:
 - It feels a bit magic to add a Java module for this. I'm not entirely adverse to it — this is never going to be part of the public API, so we can rip it out again if needed — but it does feel pretty ugly!
 - Lua bytecode doesn't store the full list of names in scope (only locals, and upvalues which are captured), so any `attempt to global "foo"` errors will have incomplete/incorrect suggestions.

The alternative here would be to parse the code in Lua, and evaluate the expression that way. However, this also has some issues:
 - This incurs a pretty significant performance. I'm particularly concerned about memory, as we'd have to realise the whole parse tree. I don't think there's a way we can (easily) lazily do this.
 - Anything more advanced then what we're doing right now would require proper semantic analysis of the code. But that's probably the case with either approach, so maybe that's fine!

### DIFF
--- a/projects/core/src/main/java/dan200/computercraft/core/lua/CobaltLuaMachine.java
+++ b/projects/core/src/main/java/dan200/computercraft/core/lua/CobaltLuaMachine.java
@@ -11,6 +11,7 @@ import dan200.computercraft.api.lua.ILuaFunction;
 import dan200.computercraft.core.CoreConfig;
 import dan200.computercraft.core.Logging;
 import dan200.computercraft.core.computer.TimeoutState;
+import dan200.computercraft.core.lua.errorinfo.ErrorInfoLib;
 import dan200.computercraft.core.methods.LuaMethod;
 import dan200.computercraft.core.methods.MethodSupplier;
 import dan200.computercraft.core.util.LuaUtil;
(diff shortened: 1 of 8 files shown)


SquidDev:
I'm still not wild about the implementation, but as mentioned in the commit message, I think it's worth getting in and seeing how well it does/doesn't work (and hoping it doesn't break!).

PR #2098, opened 2025-02-10
Pocket computer lecterns

### LINKED ISSUE #1912: Add an ability to place pocket computers on lecterns
Coming from create mod linked controller mechanic. Its would be nice to have this ability to use with the create contraptions and create: aeronautics ship controlls. So the idea is that you place PC on the lectern, and then if you press on it - it starts acting like normal PC in off-hand, but rendering screen on the lectern (accepting input from player, rendering on the lectern).

SquidDev:
This allows placing pocket computers on lecterns. The terminal is rendered to all nearby players, and right clicking opens the off-hand pocket GUI. Fixes #1912.

A screenshot of a pocket computer on a lectern.

I'd originally marked this as 1.21.4 only, as it requires some API changes (`getEntity` is now always nullable. This was always the case, but some mods relied on it being non-null when an upgrade's peripheral is created). I've changed my mind, and we're going to do it on 1.20.1 and deal with the consequences.

Currently marking as draft for a couple of reasons:
 - [x] It looks bad. There's no pixel consistency, and it's so flat!!!!! I don't know what to do about this — the correct thing would be to have a custom model (like we do for printouts), but we'd also need the model to be resizeable, and I'm not sure how feasible that is.

 - [x] We should only render the terminal when the player is close (16 blocks?, maybe make this configurable?). OC did a nice thing where it faded in the terminal, to avoid pop-in.

   We could also avoid broadcasting the terminal to distant players. However, this makes the player tracking in the pocket computer significantly more complicated, so I think we should skip for now, and deal with it if it becomes an issue.


Karotte128:
Cool feature! Does this also work with wired modems (attaching a wired modem at the bottom of the lectern) to connect the pocket computer to a wired network?

SquidDev:
Not currently. I'm not 100% sure of it as a feature, and I think that's going to be quite awkward to get working, and so want to get the core feature in first.

SquidDev:
Okay, switching to an actual model works pretty well. I think we're going to have to split the model up so the light is a separate texture, but otherwise I think this looks good?

Three pocket computers (normal, light blue, advanced) sat on a lectern. The model is actually 3D now, very similar to the item model.

I should note that this is a 32x32 picture, and so twice as detailed than the lectern. I think this is fine — it's not great, but it's at least a multiple of two.

A picture from above, showing a computer running paint. You can see the border of the pocket computer is half the width of a pixel on the lectern.

Issue #1585, opened 2023-09-12
Add a modem argument to gps.locate()


SDHKLF:
As far as I can tell, gps.locate cannot be used with a wireless modem connected to the computer over a wired network. Adding an argument to use a specific modem, or having gps.locate search for modems over the network would make this possible.

SquidDev:
Can you explain what you're trying to do here? If you've got a wireless modem connected to a wired network, I'd expect it to have a constant position, and so you don't need to use GPS at all.

SDHKLF:
Right, in this case the computer is on a create contraption, but because turtles can connect to wired networks it has more potential utility than just this instance.

SquidDev:
Right, that makes sense, thanks!

I think my main concern is that `gps.locate` returns the position of the adjacent block, which in this set up will be the wired modem rather than the calling computer. It sounds like sometimes that is useful, but definitely something we need to document when adding this.

PR #2135, opened 2025-03-10
Allow GPS to use modems connected through a hub

MCJack123:
This PR changes the `gps.locate` code to use `peripheral.find` to look for a modem, instead of manually enumerating through `rs.getSides()`, which is both more verbose than necessary and skips over networked modems.

### DIFF
--- a/projects/core/src/main/resources/data/computercraft/lua/rom/apis/gps.lua
+++ b/projects/core/src/main/resources/data/computercraft/lua/rom/apis/gps.lua
@@ -106,7 +106,7 @@ function locate(_nTimeout, _bDebug)
         return commands.getBlockPosition()
     end
 
-    -- Open GPS channel to listen for ping responses
+    -- Find a modem
     local modem = peripheral.find("modem", function(_, m) return m.isWireless() end)
     
     if modem == nil then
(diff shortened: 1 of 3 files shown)


fatboychummy:
Hmm, I wonder how useful this would be to also add this to `gps host`, then also allow passing in a peripheral name as 5th argument to override the `peripheral.find` and force it to use that one?

With this, we could create a complete GPS array with a single computer + multishell.

```lua
-- startup.lua
shell.run("bg gps host x1 y1 z1 modem_1")
shell.run("bg gps host x2 y2 z2 modem_2")
-- ...
```

Wojbie:
I believe gps is skipping networked modems on purpose, because networked modems return distance to full block modem they are networked from, not computer, so you would get position of that full block modem instead of computer itself. This feels like breaking change.

EDIT: #1585 adding modem argument would be indeed a way to implement this addition without risking breaking change, because that way only programs that explicitly opt in the new logic to get position of networked modem would get its position, and there is no risk of program confusing its position with modem one because default logic changed.

SquidDev:
Thanks for the PR! I definitely agree this makes sense to have, but not sure it should be by default. If a computer has both a local and networked modem attached (I'm not here to judge!), you probably want to prioritise the former. 

I wonder if a better approach here would be to do #1585 and just add a modem argument? 

--- 

RE `gps host`, I think it would be nice to have a `gps.host` function, which possibly supports multiple modems, but probably out-of-scope for this PR. 

> networked modems 

This is wireless modems connected via a wired one, rather than wired ones themselves.

SquidDev:

PR #2140, opened 2025-03-13
docs: specify valid types for settings.define

Lemmmy:
As per `valid_types` defined here:
https://github.com/cc-tweaked/CC-Tweaked/blob/676fb5fb53d14222bd90e60fef1f25f16495600b/projects/core/src/main/resources/data/computercraft/lua/rom/apis/settings.lua#L48


SquidDev:
Thank you!

PR #2164, opened 2025-03-26
Block MapColor metadata provider for block inspection

ShreksHellraiser:
This PR modifies `BlockDetails` to provide the MapColor of the block in question. Here is a demo program, which when run on a turtle, will change the color of the screen to the MapColor of the block in front of the turtle:
```lua
local function show(color)
    term.setBackgroundColor(colors.black)
    term.setPaletteColor(colors.black,color)
    term.clear()
    term.setCursorPos(1,1)
    term.write(color)
end

while true do
    local i, detail = turtle.inspect()
    if i then
        show(detail.color or 0)
    end
end
```

I don't personally have a usecase in mind for this yet, but someone in the discord asked if they could get the color of a block through `turtle.inspect`, and it peaked my curiosity.

This PR only adds this for Block inspection because `BlockState:getMapColor` takes in a `Level` and `BlockPos` as arguments. For experimentation's sake I passed `null, null` in and confirmed that works for getting the MapColor of blocks while they are in the inventory. However, the function args are not annotated as Nullable, and may theoretically be used in some places. Also the Fabric version refuses to build (because of the null checker). So for the sake of compatibility, this is only for blocks placed in-world.

I also feel a little weird directly accessing the `.col` member of `MapColor`, but the `MapColor:calculateRGBValue` function appears to or the RGB values with a magic number, resulting in the incorrect colors.


ShreksHellraiser:
Went back and gave the game code another lookover. Turns out the method `MapColor:calculateRGBValue` *actually* calculates the AGBR value instead. So while there is an alpha channel added, it also inverts the RGB color order. I suspect its because it is used for rendering.

Accessing the `.col` field is the most direct way, but if it is better for the long run, I can switch it to using calculateRGBValue and then swapping the values back around.

ShreksHellraiser:
I added this back to items via `EmptyBlockGetter.INSTANCE` and `BlockPos.ZERO`, if it's an item that does not have an associated block it returns Air and the color is black (0). Tested on both forge and fabric.

SquidDev:
Thanks for the PR, and sorry for the delay in looking at this!

PR #2194, opened 2025-05-04
Fix syntax highlighting for strings ending in an escaped backslash

LorneHyde:
Syntax highlighting does not work for strings that end in an escaped backslash. For instance, "\\" is a valid string which should be highlighted, as the double backslash means that we're _not_ escaping the end quotation mark. Any string ending in an even number of backslashes means that all backslashes are escaped.

Lua patterns are more limited than actual regexes, so I've switched to a different algorithm for syntax highlighting strings.

See the screenshot below to prove that this works and doesn't break existing syntax highlighting:
image

### DIFF
--- a/projects/core/src/main/resources/data/computercraft/lua/rom/programs/edit.lua
+++ b/projects/core/src/main/resources/data/computercraft/lua/rom/programs/edit.lua
@@ -215,35 +215,29 @@ local function tryWrite(sLine, regex, colour)
 end
 
 local function tryWriteString(sLine)
-    quotationChar = string.sub(sLine, 1, 1)
-    if quotationChar ~= '"' and quotation_char ~= "'" then
+    local quotationChar = string.sub(sLine, 1, 1)
+    if quotationChar ~= '"' and quotationChar ~= "'" then
         return nil
     end
 
-    -- Scan through the rest of the string, until we
-    -- find the closing quote. Ignore quotation marks
-    -- with an odd number of backslashes preceding them.
+    -- Scan through the rest of the string, skipping over escapes,
+    -- until we find the closing quote.
     local i = 2
     while i <= #sLine do
         local nextChar = string.sub(sLine, i, i)
-        if nextChar == quotationChar then
-            local j = i - 1
-            local backslashes = 0
-            while j > 0 and string.sub(sLine, j, j) == '\\' do
-                backslashes = backslashes + 1
-                j = j - 1
-            end
-            if backslashes % 2 == 0 then
-                match = string.sub(sLine, 1, i)
-                term.setTextColor(stringColour)
-                term.write(match)
-                term.setTextColor(textColour)
-                return string.sub(sLine, #match + 1)
-            end
+        if nextChar == "\\" then
+            i = i + 2 -- Skip over escapes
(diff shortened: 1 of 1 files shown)


projects/core/src/main/resources/data/computercraft/lua/rom/programs/edit.lua
```diff
@@ -214,15 +214,44 @@ local function tryWrite(sLine, regex, colour)
...
+        return nil
+    end
+
+    -- Scan through the rest of the string, until we
+    -- find the closing quote. Ignore quotation marks
+    -- with an odd number of backslashes preceding them.
+    local i = 2
+    while i <= #sLine do
+        local nextChar = string.sub(sLine, i, i)
+        if nextChar == quotationChar then
```
SquidDev: I wonder if you could make this a bit simpler by simply checking for `\\` here. So something like:


```lua
    local i = 2
    while i <= #sLine do
        local nextChar = string.sub(sLine, i, i)
        if nextChar == "\\" then
            i = i + 2 -- Skip over escapes
        elseif nextChar == quotationChar then
            break
        else
            i = i + 1
        end
    end

    -- By shifting it to the end, we also highlight incomplete strings, like `print("hello`.
    term.setTextColor(stringColour)
    term.write(string.sub(sLine, 1, i))
    term.setTextColor(textColour)
    return string.sub(sLine, i + 1)
```

SquidDev (changes requested):
Thanks for the PR and apologies for taking so long to look at this. I'm afraid CC:T is very much on the back burner for me right now, and so very much not getting the love and attention it probably deserves.

I've left one comment which I think would make this a bit cleaner, but otherwise happy to merge!

The long-term fix here is probably to switch `edit`'s syntax highlighter to use a proper lexer. CC:T [does already have one](https://github.com/cc-tweaked/CC-Tweaked/blob/b5c0c6e104c626fbfb6276b01bf0882603fb5d09/projects/core/src/main/resources/data/computercraft/lua/rom/modules/main/cc/internal/syntax/lexer.lua), which should be relatively reusable. I don't know if that's something you'd be interested in looking at? Definitely understand if not — I realise it's significantly more work, and this PR has sat unmerged for too long already!

projects/core/src/main/resources/data/computercraft/lua/rom/programs/edit.lua
LorneHybeat: Good idea, I've just implemented that :)

LorneHyde:
I've implemented your suggestions, could you please have another look? 

With regards to the highlighting of incomplete strings, one thing to keep in mind is the fact that `edit.lua` gives `textutils.complete` suggestions even when in a string. While this behaviour already feels a bit wrong, it does look worse when the code is already highglighted as a string. Examples of what I mean in screenshots below:

image
image

That being said, I still think highlighting incomplete strings is beneficial overall, I was just also bringing the (potential?) disadvantage to your attention :)

Below is a screenshot showing that my code works:
Screenshot 2025-06-04 232050

SquidDev:
Sorry, meant to say: thank you for working on this! Very much appreciated!

PR #2203, opened 2025-05-14
Add notes about minor changed file handle behavior in 1.109.0

MCJack123:
This PR adds some notes about minor behavior changes to file handles in 1.109.0 that may be important to users dealing with compatibility, and that I missed in that update and am definitely not adding notes for out of spite.


SquidDev:
Thank you! I genuinely thought I'd mentioned this in the changelog, but apparently not! It only shows up in tests :facepalm:

PR #2211, opened 2025-05-31
Fix parse_color

SpartanSf:
## Minecraft version
Universal
## Version
All since this code has existed
## Details
First discovered when attempting to run `term.setBackgroundColor(0)`, attempting to input 0 to a function that uses `parse_color` will crash the current program with an error that does not explain what went wrong. This only occurs on advanced computers.


SquidDev:
Thanks!

PR #2224, opened 2025-06-21
Miniscule typo fix in the shell.path() doc comment

matematikaadit:
Noticed this when reading the [shell API page](https://tweaked.cc/module/shell.html#v:path) in the wiki.


SquidDev:
Thanks!

PR #2238, opened 2025-07-10
Add capacity to fluid tank data (1.20.x)

### LINKED ISSUE #1629: Missing a way to get the capacity of a tank/fluid storage
I am looking for a way to get the maximum capacity of a tank / fluid_storage but the result from the .tanks() method on fluid_storage only gives the amount and name of the fluid.

Is there a technical limitation for this feature not being here ? I would assume forge tanks have an API to get the maximum capacity, is it not the case?

EmmaTheMartian:
Closes #1629

The Forge implementation feels pretty hacky. I considered trying to change `ComputerCraftAPIImpl#fluidStackDetails` to a `DetailRegistry<IFluidTank>` or similar, however it seemed like it would have been less compatible or would entail a more breaking change for addon developers.

The Fabric implementation was clean-cut.

I also made this change for 1.21.x and will make a separate PR for that.

### DIFF
--- a/projects/fabric/src/main/java/dan200/computercraft/shared/details/FluidDetails.java
+++ b/projects/fabric/src/main/java/dan200/computercraft/shared/details/FluidDetails.java
@@ -21,6 +21,7 @@ public class FluidDetails {
     public static void fillBasic(Map<? super String, Object> data, StorageView<FluidVariant> fluid) {
         data.put("name", DetailHelpers.getId(RegistryWrappers.FLUIDS, fluid.getResource().getFluid()));
         data.put("amount", fluid.getAmount());
+        data.put("capacity", fluid.getCapacity());
     }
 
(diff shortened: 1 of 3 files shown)


SquidDev:
Thanks for the PR! 

I'm not 100% sure about putting this in the result of `tanks()`. This method is somewhat meant to mirror inventory's `list()` method, and only return information about the fulid itself, while the capacity is much more a property of the tank itself[^1].

I wonder if a better approach here would be to have a `getTankCapacity(tank: number)`, like `getItemLimit`. Though then again, that won't work on Fabric, unless we want to constrain ourselves to slotted storages... 

I'm not sure. Thoughts welcome!

[^1]: The fluid peripheral is so badly designed. It's so tied to Forge's specific interface for fluid tasks :/.

PR #2265, opened 2025-08-22
Directional Support for turtle.place

### LINKED ISSUE #204: Directional block placement by turtles
I would love to see a way for turtles to have more control over the placement of stairs, torches, logs, chests and other blocks that are influenced by the direction they are placed in.

**How the feature/change should work**
Currently there is some base logic on the way these blocks are placed. Example for slabs:
``turtle.place()`` defaults to the bottom placement,
``turtle.placeUp()`` defaults to the bottom placement
``turtle.placeDown()`` defaults to the top placement.
When there is a block below the position the slab will be placed in it, the slab will be placed in the bottom
Otherwise when there is a block above the position the slab will be placed in it, the slab will be placed in the top

For stairs it also depends on the surrounding blocks, and maybe the direction of the turtle. In my opinion the way this works is really confusing. My suggestion would be to have the ``turtle.place*()`` functions allow for parameters ``x``, ``y`` and ``z``.
It should default to the way a turtle places at the moment so existing systems won't break down and things won't become more difficult for the beginners.

The effect of ``x``, ``y`` and ``z`` should be decided by _greater than 0_, _less than 0_ and _equals 0_. How these values should be interpreted is up for discussion (see below)

**Some rationale/use case for a feature.**
Having more control over the placement of blocks we can make turtles build more advanced structures without having to do strange movements, block placing/destroying/placing.

**Discussion points:**
_How should the values be interpreted?_

PeenDev:
Added support for the following parameters in the turtle.place function.

north, south, east, west these are self explanatory. 

face_up, and face_down are for things like dispensers, though you can also use it for something like a modem for example. 

left, right, up, down just calculate the cardinal direction relative to the turtles facing direction. 

top, bottom for things like slabs

ground for things like torches when you don't want to place them against a wall

upside_down for stairs 

you can pass multiple for example turtle.place("upside_down", "north")

if it cant resolve the orientation, cant apply them to a block, or no parameters are passed it falls back to the previous placement logic so nothing should change for existing programs. 

Directionality seems to be handled differently by different sets of blocks for some reason.

logs use axis, shulkers use facing, furnaces use horizontal facing, modems use their own enum, stairs use half, slabs use type
tried to account for all the variation, there may be a better way feel free to discuss. 

Happy to accept criticism or pointers! Thanks!


projects/common/src/main/java/dan200/computercraft/shared/turtle/core/TurtlePlaceCommand.java
```diff
@@ -245,8 +310,150 @@ private static void setSignText(Level world, SignBlockEntity sign, String messag
...
+        boolean isUpsideDown = false;
+        boolean isGroundAttachment = false; // for torches on ground vs wall
+        boolean foundOrientationParam = false;
+        Direction clickFace = null;
+
+        if (extraArguments != null && extraArguments.length > 0) {
+            // Process all string arguments as potential orientation parameters
+            for (var arg : extraArguments) {
+                if (arg instanceof String orientation) {
+                    switch (orientation.toLowerCase()) {
```
SquidDev: We should use the ROOT locale, to avoid weird system-specific issues.

```suggestion
                    switch (orientation.toLowerCase(Locale.ROOT)) {
```

I don't know if it's also worth checking unknown orientations, and rejecting invalid combinations (e.g. `"left", "right"`)?

projects/common/src/main/java/dan200/computercraft/shared/turtle/core/TurtlePlaceCommand.java
```diff
@@ -198,9 +261,11 @@ private static boolean deployOnBlock(
      * @param hit          Where the block we're placing against was clicked.
      * @param adjacent     If the block is directly adjacent to the turtle, and so can be interacted with via
      *                     {@link BlockState#use(Level, Player, InteractionHand, BlockHitResult)}.
+     * @param orientationParams The parsed orientation parameters for block placement
      * @return If this item was deployed.
      */
-    private static InteractionResult doDeployOnBlock(ItemStack stack, TurtlePlayer turtlePlayer, BlockHitResult hit, boolean adjacent) {
+    private static InteractionResult doDeployOnBlock(ItemStack stack, TurtlePlayer turtlePlayer, BlockHitResult hit, boolean adjacent,
```
SquidDev: I don't think the `OrientationParams` is used right now?

projects/common/src/main/java/dan200/computercraft/shared/turtle/core/TurtlePlaceCommand.java
```diff
@@ -245,8 +310,150 @@ private static void setSignText(Level world, SignBlockEntity sign, String messag
...
+
+        return newState;
+    }
+
     private static final class ErrorMessage {
         @Nullable
         String message;
     }
+
+    private static final class OrientationParameters {
```
SquidDev: This might be a bit cleaner as a record.

SquidDev (commented):
Thank you so much for PR! Especially all the tests — it's relaly, really appreciated! Apologies for taking so long to look at this — laptop died on me, so haven't been able to run Minecraft for the past few weeks :/.

When this was first discussed in #204, I did have some worries about how many edges cases we'd have to handle here. I think this is definitely better than I was expecting, but there are always going to be edge cases. I wonder if there should be an API method to register the block-state transforms (e.g. `ComputerCraftAPI.registerOrientationProvider(Block, BiFunction<BlockState, OrientationParameters, BlockState>`).

A couple of other smaller thoughts:
 
 - I'm in mixed minds about the cardinal direction specifiers (`north`, `south`, etc...). CC:T generally tries to avoid using any absolute positions. *However*, block states are a bit of a special case, as they're available with `turtle.inspect()`.

   My instinct is to remove them, but happy to discuss further!

 - The relative directions are sometimes flipped. Doing `turtle.place("left")` with a monitor, turtle or dispenser ends up placing a block facing to the right. Wired modems *do* face the right way, so I suspect they're inconsistent in some annoying way :/.

 - I wonder if we could merge some of the other specifiers? e.g. `upside_down` and `up`; `bottom` and `ground`.

 - There's some ambiguity here whether `turtle.place("left")` on a sign should place a sign with the text "left", or place a sign facing to the left. I wonder if it we should instead have `turtle.place` accept a table, so you do `turtle.place { facing = { "left" }` (also accepting a `sign_place` argument, so you can do `turtle.place { facing = { "left", "ground" }, text = "hello!" }`).

   As part of this, it probably makes sense to do the parsing when the `TurtleAPI` function is called, and just pass the orientation/text parameters to the place command.

PeenDev:
Hey thanks for the review! 

- Yeah I had initially done the cardinal direction placements and then thought it wouldn't be that helpful without already knowing the direction you're facing so I implemented the relative direction. I am happy to remove the cardinal direction placements- though I don't feel the same instinct about removing them, getting the turtle to resolve its direction is pretty simple and can be done multiple ways, but again happy to remove it!

- as far as the placement inconsistencies I think its a different understanding of "left", the way I interpreted it was not "facing" it was simulating as if the turtle had clicked the "left" block, this seemed like a more consistent way to handle it to me. (though the turtle placement is one that doesn't work how you would expect if you place it yourself in game)

- Yeah some of the specifiers can be merged, I think I just got caught up in all the edge cases I added more than needed, I'll clean that up. I do think that "ground" should stay though or we could flip it and have it default to placing on the ground and instead have a parameter that places it against the wall (toches, lightning rod, end rod)

- had not even considered a sign, thats a great callout I will make the appropriate changes!

PeenDev:
Just Bumping- Haven' taken a look at this in a while as there was no movement and just found some time to come back to it.

PR #2323, opened 2025-11-11
changelog: retroactively document changes to os.version()

tomodachi94:
This is useful information for people who want to know what changed along with the changes for these version numbers, if someone wants to write a backwards-compatible program that relies on behavior that changed between the two versions.

I had to do some archaeology to find this information. Here are the commits:
* 2c7ee1d7e75dc89f720e40feee430fbfb8cb525c
* d272a327c7d48e7001361697764687031872aa8d

An aside: it would be nice if this version number changed whenever a breaking change occured in CraftOS, but I digress.

### DIFF
--- a/doc/stub/global.lua
+++ b/doc/stub/global.lua
@@ -117,10 +117,14 @@ function read(replaceChar, history, completeFn, default) end
 -- Outside of Minecraft (for instance, in an emulator) [`_HOST`] will contain the
 -- emulator's version instead.
 --
+-- If you need to check for the presence of a feature, it is usually better to
+-- rely on feature detection, rather than comparing mod or Minecraft versions.
+--
 -- For example, `ComputerCraft 1.93.0 (Minecraft 1.15.2)`.
 -- @usage Print the current computer's environment.
 --
 --     print(_HOST)
+-- @see os.version
 -- @since 1.76
 _HOST = _HOST
 
--- a/doc/stub/os.lua
+++ b/doc/stub/os.lua
@@ -100,13 +100,12 @@ function sleep(time) end
 -- This is defined by `bios.lua`. For the current version of CC:Tweaked, this
 -- should return `CraftOS 1.9`.
 --
--- The version returned by this function is purely decorative.
--- It is usually better to rely on direct feature detection instead of
--- programmatically using the value of this string.
+-- If you need to check for the presence of a feature, it is usually better to
+-- rely on feature detection, rather than comparing CraftOS versions.
 --
 -- @treturn string The current CraftOS version.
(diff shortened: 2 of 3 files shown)


tomodachi94:
Another aside: I wonder if `os.version()` could provide a path for migrating pure-Lua APIs to `require`able modules, potentially allowing us to completely drop `os.loadAPI`. Updating `os.version()` would allow consumers that care about backwards-compatibility to account for the change[^1], and others could move on if they made that determination.

Or maybe this is a bit too grandiose of an aspiration :) please let me know. I'd be more than willing to try taking this on.

[^1]: On second thought, `local textutils = textutils or require("textutils")` should be sufficient for backwards compatibility

SquidDev:
Thanks for the PR!

As you mention later, it's much better to do feature detection rather than relying on the value of `os.version`. The version string is very much just window dressing. It might be worth updating `os.version`'s doc string to mention that. 

I don't know if you want to add this changelog entry for all version changes? The rest should be much easier, as they're just synced to CC version bumps (e.g. CC 1.60->1.70 would have changed os.version to CC 1.6->1.7).

> An aside: it would be nice if this version number changed whenever a breaking change occured in CraftOS, but I digress.

I think it should already — [the only breaking changes I'm aware of](https://tweaked.cc/reference/breaking_changes.html) are in 1.80pr0 and 1.109. Are there any I've missed?

tomodachi94:
> The version string is very much just window dressing. It might be worth updating `os.version`'s doc string to mention that.

I'll update that.

> I don't know if you want to add this changelog entry for all version changes? The rest should be much easier, as they're just synced to CC version bumps (e.g. CC 1.60->1.70 would have changed os.version to CC 1.6->1.7).

Yes, I totally can!

> I think it should already — [the only breaking changes I'm aware of](https://tweaked.cc/reference/breaking_changes.html) are in 1.80pr0 and 1.109. Are there any I've missed?

You're right, my bad. I don't think there are any.

tomodachi94:
I updated the documentation for `os.version` to describe it as "purely decorative" and encouraging feature detection instead. I also added a `@see _HOST` line in the hopes that people might use that instead if they insist on manipulating version strings.

tomodachi94:
Also: should I open a new issue about migrating some of the APIs to `require`?

SquidDev:
Thanks for the changes!

> Also: should I open a new issue about migrating some of the APIs to require?

CC:T tries to be as backwards compatible as possible, so we're never going to get rid of `os.loadAPI`. The most we could do here is allow `require`ing the existing APIs, and I'm not sure there's much point in that.

PR #2329, opened 2025-11-17
Made Pastebin uploads unlisted instead of public

Guedosha:
I made Pastebin uploads unlisted instead of public.
The majority of the time this isn't an issue but a player might upload something they shouldn't without realizing the pastes are public by default.


Rubrik25:
bro...
```lua
        "api_paste_format=lua&" ..
        "api_paste_private=1&"  -- where is ".."
        "api_paste_name=" .. 
```

Guedosha:
No idea how I forgot that... Just fixed it

PR #2333, opened 2025-11-22
Fix missing return value docs in shell.openTab

MCJack123:
This PR fixes missing documentation for `shell.openTab`'s return value, which returns the same as `multishell.launch` via tail call.


SquidDev:
Thanks!

PR #2367, opened 2026-01-16
Fix #2355

### LINKED ISSUE #2355: Inconsistant behavior with moving fluids from some mods and not others.
### Minecraft Version

1.20.1

### Version

1.116.2

### Details

(Logs not available, issue experienced in survival mode on someone else's server and not available to me).

- Minecraft 1.20.1
- Forge 47.4.13
- CC: Tweaked 1.116.2
- Create 6.0.8
- Advanced Peripherals 1.20.1-0.7.46r (but not using its API in this case).

At a high-level, I am attempting to use CC to automate the movement of fluids from the "Create Big Cannons - Chemical Weapons [1]) between Create basins and Create fluid tanks.  The CC API `pushFluid()` and `pullFluid()` are failing to move non-vanilla Minecraft fluids on a "large" (230+ peripherals) network when the fluid is named as the 3rd argument to the method call.  ALL fluid types do move when the name is omitted (`nil`), or when I move a vanilla fluid, like `minecraft:milk`.

When i tried to reproduce the issue (same server, adjacent chunks) with a minimal network (1 computer, 2 basins, 1 tank), the problem goes away: I can move any fluid from any mod by name.  This seems a bit baffling.

Lets start with the working example. A minimal network with the following attached to it

Image

- modem (a), cable and modem (b)
- computer on modem a
- `basin_9` (empty) on modem a
- `basin_10` (empty) on modem b
- `fluidTank_54` (empty) on modem b

Now let's put a bucket of `cbc_cw:oleic_acid` into `basin_9`, and run a little lua to verify initial state:

Image

Now lets attempt to move 250 millibuckets of `cbc_cw:oleic_acid`  from `basin_9` to `fluidTank_54`:

Image

Yay.  that works.  Ok, lets test this on the "big network".  The target tank can store > 8 buckets (its 6 blocks tall, so 48 buckets), and has 500 millibuckets of the acid already in it.

Image

Attempting to move 250mb **by name** fails, but when I don't specify the name, fluid movement works.

Image

The failure when specifying the fluid name on the large network is a problem.  Often a basin will contain multiple fluids, and I need to move only a specific amount of specific fluids.  The documented API says this should be possible, and it works on a "small" network, but not on a "big" one.

Further, I can move `minecraft:milk` from a different basin (`basin_8` ) to a different tank (`fluidTank_12`), all on the same network as the CBC stuff, by name.

I can move `createaddition:seed_oil` from `basin_7` to `fluidTank_13`.

Image

But I cannot move any CBC fluid (I've tried several) from any basin to any tank on a large network when I specify the fluid name to the API call.  All other modded fluids that i try work fine, and CBC fluids work on a small (attempted repro) network (so, i failed to repro at small scale).

This is quite... confusing.  And I am 100% certain that the destination tanks are EMPTY or contain the exact same fluid before testing.

The only other weird thing is that when I mouse-over a basin or tank holding a CBC fluid, the advanced tooltip (possibly from the Jade mod?) shows the tank contents as `fluid_type.cbc_cw.oleic_acid`, but only for CBC fluids.  Milk and Seed oil shows a non-item-string type name "Milk".  IDK if this matters.  Maybe is is a red herring.

I can move water, milk, lava, seed_oil and other fluids into and out of the tanks used in my testing (the exact same tanks that fail to move CBC fluids). 

[1] https://modrinth.com/mod/cbc-chemical-warfare

UQuark:
create's internal CombinedTankWrapper compares components with `isSameFluidSameComponents` but we provide a constructed FluidStack with no components at all. all equality checks fail 

this obtains the exact desired tank by id in a dumb for loop and passes it to the `drain`, thus fixing fluid not being pushed when specified via a name filter as described in #2355

### DIFF
--- a/projects/forge/src/main/java/dan200/computercraft/shared/peripheral/generic/methods/FluidMethods.java
+++ b/projects/forge/src/main/java/dan200/computercraft/shared/peripheral/generic/methods/FluidMethods.java
@@ -12,6 +12,7 @@ import dan200.computercraft.api.peripheral.IPeripheral;
 import dan200.computercraft.shared.platform.RegistryWrappers;
 import dan200.computercraft.shared.util.CapabilityUtil;
 import net.minecraft.world.level.block.entity.BlockEntity;
+import net.minecraft.world.level.material.Fluid;
 import net.minecraftforge.common.capabilities.ForgeCapabilities;
 import net.minecraftforge.common.capabilities.ICapabilityProvider;
 import net.minecraftforge.fluids.FluidStack;
@@ -63,7 +64,7 @@ public final class FluidMethods extends AbstractFluidMethods<IFluidHandler> {
 
         return fluid == null
             ? moveFluid(from, actualLimit, to)
-            : moveFluid(from, new FluidStack(fluid, actualLimit), to);
+            : moveFluid(from, fluid, actualLimit, to);
     }
 
     @Override
@@ -88,7 +89,7 @@ public final class FluidMethods extends AbstractFluidMethods<IFluidHandler> {
 
         return fluid == null
             ? moveFluid(from, actualLimit, to)
-            : moveFluid(from, new FluidStack(fluid, actualLimit), to);
+            : moveFluid(from, fluid, actualLimit, to);
     }
 
     @Nullable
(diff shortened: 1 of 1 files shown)


SquidDev:
T'anks for the PR! 

I'm not sure special casing Create's tank here is the right solution (not least because it'll crash when Create isn't installed :)). I think we can probably use this logic for all tank kinds. I don't think than'll cause any problems.

PR #2374, opened 2026-01-26
Fix discrepancy in `type` help

Pokebrouserkat:
to see for yourself, run `help type`, then `type potato`

Image of it:
Screenshot 2026-01-26 at 11 18 34 AM

While "No such path" is capitalized while real types are not, this is hardcoded to be the case, so I assume it is intended and updated the presumably outdated help info instead.

(this was originally going to be an issue, but it's so simple I thought I'd just go do it myself)


SquidDev:
Thanks!

PR #2375, opened 2026-01-26
Add cat, echo, and touch shell programs

Pokebrouserkat:
I've been using startup scripts to make a more unix-like environment for a while now, thought I'd add a few common things. I'll be honest where AI (Claude Code) was used:

- `echo` can be used in combination with `monitor` to print on a monitor without having to first write a program. Claude tried wrapping tArgs and joining them with " " but I just `print(...)` directly, plus `{...}.x` or `{...}[x]` tends to be unreliable in Lua unless cast to a variable first.
- `cat` prints file contents, again, without having to go through `edit`, can be used in combination with a txt file for what echo can do. I have a `less` command on my own computer which uses `textutils.pagedPrint`, that's more helpful for larger files Claude basically rewrote this, my implementation was basically just `print(fs.open(..., "r").readAll())`, which gets the job done I guess.
- `touch` creates an empty file, or does nothing if the file already exists. Again, my implementation was very basic, `fs.open(..., "a").close()`, Claude did everything else in this file.

### DIFF
--- a/projects/core/src/main/resources/data/computercraft/lua/rom/programs/cat.lua
+++ b/projects/core/src/main/resources/data/computercraft/lua/rom/programs/cat.lua
@@ -1,0 +1,26 @@
+-- SPDX-FileCopyrightText: 2025 The CC: Tweaked Developers
+--
+-- SPDX-License-Identifier: MPL-2.0
+
+local tArgs = { ... }
+
+if #tArgs < 1 then
+    local programName = arg[0] or fs.getName(shell.getRunningProgram())
+    print("Usage: " .. programName .. " <paths>")
+    return
+end
+
+for _, v in ipairs(tArgs) do
+    local sPath = shell.resolve(v)
+    if fs.exists(sPath) then
+        if fs.isDir(sPath) then
+            printError(v .. ": Is a directory")
+        else
+            local file = fs.open(sPath, "r")
+            print(file.readAll())
+            file.close()
+        end
+    else
+        printError(v .. ": No such file")
+    end
+end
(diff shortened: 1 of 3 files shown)


Pokebrouserkat:
all comment headers from Claude too

SquidDev:
> Claude Code 

Nah

Pokebrouserkat:
just because Claude mildly (or severely) adjusted a file does not mean the concept is bad, thoughts on at least `echo` or non-AI versions of the other files?
[startup.lua.zip](https://github.com/user-attachments/files/24864883/startup.lua.zip)
this file has no AI-generated components to it, I didn't finish it but it generates a bunch of Unix-like programs I use. basically whenever I tried to run a command and it didn't exist, I made it here so it would exist, making sure to minify everything because the last time I did this I ran out of space. I only used Claude to make sure the programs fit in, I never use Claude (or any AI) on things I do in-game. At absolute most, I have asked ChatGPT things like how to do thing X in Lua, and then it is often wrong, and I just end up reading the docs anyway. I zipped the Lua file because GitHub said it doesn't accept Lua files but does accept zip files

Pokebrouserkat:
I understand if you don't want AI in the codebase, but as I said, I'm honest about where it is used, it seems harsh to reject everything because I used AI once, so I'll point out again that I did not use AI for a single byte or concept of the following files, and provide the zipped `bin` that the startup.lua I attached earlier creates.
[bin.zip](https://github.com/user-attachments/files/24865477/bin.zip)

zyxkad:
Aside from AI, I don't think these unix programs fits in CC.

1. CC:T's shell does not support scroll history, so `cat` won't work better than `edit`.
2. `echo` has basically no use since CC:T does not have a shell script interpreter.
3. For `touch`, it is also commonly used in shell script. I don't think many people will just create a file without writing anything into it.

Pokebrouserkat:
`less` instead of `cat` then, and `which` is still helpful for finding programs without having to manually dig through rom and also determining if a custom override is working (path update, like the aforelinked `startup.lua`)

zyxkad:
If you do want more programs, I suggest you open an issue to discuss what kind of program is necessary in CC:T first. Open a PR to add features without any discussion often won't work.

PR #2386, opened 2026-02-13
Add optional password authentication for rednet/chat

Dewn5228:


projects/core/src/main/resources/data/computercraft/lua/rom/programs/rednet/chat.lua
```diff
@@ -272,8 +293,29 @@ elseif sCommand == "join" then
         sType = "login",
         nUserID = nUserID,
         sUsername = sUsername,
+        sPassword = sPassword,  -- send password (may be nil)
```
fatboychummy: Sending a raw password like this is a bad idea in CC for many reasons...

1. Rednet is just a routing library, built on top of the `modem` peripheral. It is not secure, and you can *very easily* sniff the contents of any rednet message.
2. If anyone does the above, the raw password is leaked.
3. People may assume using a password is secure and reuse passwords they use elsewhere, leading to other personal accounts getting breached.

If you are going to do this, I recommend looking up how hashing works, as well as encryption.

fatboychummy:
The current implementation (since it doesn't use any sort of key exchange or encryption) can also be bypassed by copying the chat.lua file and removing the error if the login response fails.

Dewn5228:
Sorry for the late patch. I had this ready earlier but forgot to push. This update now adds a `crypto` API (PBKDF2, HMAC, AES) and uses it for secure password-based authentication and message encryption. Let me know if adding a new API here is acceptable.

SquidDev:
Thanks for the PR, and sorry for the slow response on this. I'm afraid this is not something  I'm going to merge.

Most of CC's built-in utilities are designed to be simple and "good enough" for the common use-case. I'm definitely open to improvements to them, but I don't want to diverge from that simplicity too much. Encrypted chat is definitely out of scope here — good cryptography is *hard*[^1], and that's not something I want to have to maintain. I think that's better served by community programs.

[^1]: For instance, I'm think there's a risk of a timing side-channel attack on line 222 (`elseif sClientResponse ~= sExpectedResponse then`). I also think this is vulnerable to replay attacks. This stuff is really hard to get right!

PR #2396, opened 2026-03-09
Fix missing reference to data packs in .ignoreme files

ItsNotVingtdeux:
The autorun and module folders can be given files using resource packs, but as of Minecraft 1.13, they are modified using data packs. This PR adds a mention of datapacks to the .ignoreme files in these folders.


SquidDev:
Thanks!

PR #2398, opened 2026-03-11
Fix shebang absolute path handling

graypinkfurball:
**The Problem:**
You cannot write a program that correctly handles user-supplied relative paths and shebangs inside script files without using some kind of workaround.
A user-facing program typically calls `shell.resolve` on any path arguments passed before using the result with functions like `fs.exists` and `fs.open`.
When a shebang passes the absolute path of the script file to the program, it does so without a leading slash, thus it can be easily mangled by `shell.resolve` if the current directory is not `/`.
Some workarounds include using an option flag (e.g. `--absolute` or `--shebang`) to indicate to the program when not to call `shell.resolve`.

**Reproduction Steps:**
1. `mkdir test/`
2. `cd test/`
4. `edit test.txt`:
```
#!type
```
5. `type test.txt` should correctly print `file`
6. `test.txt` should incorrectly print `No such path` despite calling the same program on the same file.

Another example:
`#!edit` on a file `a/b/c.lua` executed with current directory `a/b/` saves to `a/b/a/b/c.lua`

**Rationale:**
The shell program makes the distinction between relative paths and absolute paths, it should also consistently produce explicitly absolute paths (with leading slashes) when passing known absolute paths to programs that may use the shell API.
I understand that the leading slash isn't neccessary when the program uses the underlying filesystem API, but since programs can receive user-passed explicitly absolute and relative paths and convert them to absolute paths with `shell.resolve`, they should be able to handle shebang absolute paths seamlessly.

**Provided Solution:**
Always prepend a leading slash to the program path when adding it to the hashbangArgs table.
It shouldn't break backwards compatibility since the filesystem API ignores leading slashes.

### DIFF
--- a/projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
+++ b/projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
@@ -141,8 +141,11 @@ local function executeProgram(remainingRecursion, path, args)
             return false
         end
 
-        -- Add the path and any arguments to the interpreter's arguments
-        table.insert(hashbangArgs, path)
+        -- Add the program path to the interpreter's arguments
+        -- Always add a leading slash to prevent any shell.resolve calls within the interpreter
+        --   from prepending the current directory since program path is already absolute
+        table.insert(hashbangArgs, "/" .. path)
+        -- Add any arguments passed to program to end of the interpreter's arguments
         for _, v in ipairs(args) do
             table.insert(hashbangArgs, v)
         end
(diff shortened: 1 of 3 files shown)


graypinkfurball:
lmk if i've made any mistakes in how I attempted to make a pull request, this is my first(?) time contributing. I also could've just opened an issue for this but I thought I'd try contributing.

SquidDev:
Thanks for the report and PR! Oooooh, that's nasty, good spot.

I feel ideally what we'd actually do is pass the original user-supplied path instead. So in your example, `type` would be called with `test.txt`, rather than `test/test.txt` or `/test/test.txt`. This means programs using a shebang can, for instance, use `arg[0]` to determine how the program was invoked.

I had a look at this in 3f1ef9a979f4cda97f3cea81b8e768ce10e7a97d. However, this approach is entirely broken, as it doesn't account for shell aliases, or running files without the implicit `.lua`. So I think your fix is actually the only way we can make this work.

PR #2400, opened 2026-03-20
Add Ctrl+key text navigation and editing shortcuts

TheWireLord:
Added Ctrl+key combinations for text navigation and editing in both the shell and the edit program.

**Shell (bios.lua):**
- Ctrl+Left/Right: Navigate by words
- Ctrl+Backspace: Delete word to the left
- Ctrl+Delete: Delete word to the right

**Text Editor (edit.lua):**
- Ctrl+Left/Right: Navigate by words
- Ctrl+Backspace: Delete word to the left
- Ctrl+Delete: Delete word to the right
- Ctrl+Home: Jump to beginning of file
- Ctrl+End: Jump to end of file
- Smart menu behavior: Ctrl combos don't interfere with menu system

### DIFF
--- a/projects/core/src/main/resources/data/computercraft/lua/bios.lua
+++ b/projects/core/src/main/resources/data/computercraft/lua/bios.lua
@@ -166,6 +166,7 @@ function read(_sReplaceChar, _tHistory, _fnComplete, _sDefault)
 
     local tCompletions
     local nCompletion
+    local bCtrlHeld = false  -- Track if Ctrl is currently held
     local function recomplete()
         if _fnComplete and nPos == #sLine then
             tCompletions = _fnComplete(sLine)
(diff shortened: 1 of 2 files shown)


Wojbie:
You appear to be storing if ctrl was pressed as a flag, have you tested how this code would act if someone (or multiple players) were to press both/multiple ctrl keys and released one? Also would require to check how this changes would interact with `AltGr` key presses, as that one is treated as alt+ctrl in some keyboard layouts/languages.

SquidDev:
Thanks for the PR! Afraid I'm going to close this — as Wojbie says, ctrl handling across multiple players can end up quite brittle. I had a look at attaching modifiers to the `key` event a long time ago (e.g. you got something like `key 65 false { ctrl = true, alt = false, shift = false }`), but decided against it at the time (cannot remember why). Maybe should have another look at that…

Issue #2120, opened 2025-03-01
Client and server audio streams can get out-of-sync


SquidDev:
In the current implementation of `speaker.playAudio`, the server keeps track of where it things client playback is up to. Once the client's buffer is running low (<0.5s), we send the next audio packet. On the client side, we then add the audio data to the queue.

However, this requires the server to make several assumptions about the client's audio playback, which don't always hold. For instance, if the speaker sound does not play immediately (e.g. by setting the volume to 0 (#2108), or having too many speaker (#1313)), then the client-side audio buffer will continue to back-up. When the speaker finally plays, it'll start from the beginning of its queue, rather than the server's current position.

I'm not sure what the right solution is here. I think the best option is to attach a timestamp to each audio packet (normalised to the start of the audio stream), and have some client-side handling to either pad/drop packets in order to fit to it.

---

#1874 might be worth investigating at the same time (ughr, I did look into this, and then didn't write anything up, so will have to do that again!). I don't believe the audio desyncs mentioned in #1488 are related.

joviah:
@SquidDev I'm looking into this.  It came up as I was looking at mods I'm using in a potential modpack I'm building and figured it would be nice to give back to the mods I'm playing around with.  I found this when #1874 came up on my radar.  While I haven't experienced the issue personally, I'm happy to look into it.

I think I might even sneak in a related (very minor) performance increase as well.  Nothing major... but every little bit helps.  

If you're ok with me trying to jump in and help, I should be able to get a pull request submitted shortly.  

Fixes I'm looking at should cover #1874, #2120 (your writeup above), #2108 and partially #1313.  For #1313 this fix wouldn't change the Minecraft constraints, but it will cover the desync.

SquidDev:
@joviah Sorry, had missed this message entirely. Yes, if you're happy to look into this, that would be amazing! Thank you!

Issue #1313, opened 2023-01-25
Speaker limit to 8


BuildmasterW:
### Minecraft Version

1.18.x

### Version

cc-tweaked-1.18.2-1.100.8

### Details

There appears to be a limit of 8 speakers playing at one time. I tried to connect a bunch of speakers to one network cable and used the following API https://github.com/MCJack123/AUKit After attempting to play the song "Unforgettable" in a dfpwm file, only 8 speakers played the sound, in order from which was placed first (The first 8 that were placed played the song).

SquidDev:
This is a limitation of Minecraft itself - it only allows 8 streaming sounds to be played at once. You can see this with Jukeboxes as well, if you try to play 9 records at the same time.

Unfortunately I don't think there's anything I can really do to fix this short of bypassing Minecraft's sound system (which I really don't want to do). I'll leave this open in case people have ideas, but I fear this will remain unfixable.

BuildmasterW:
@SquidDev Where is the limitation on Minecraft? Is it client or server sided? Also, do you know if there is any way around it (EG: Mod or changes in a config file)

Thanks!

SquidDev:
It's a client side limitation in the underlying sound library. See `com.mojang.blaze3d.audio.Library`, where it creates the pools of sound channels:

```java
int channels = this.getChannelCount();
int streamingLimit = Mth.clamp((int)Mth.sqrt((float)channels), 2, 8);
int staticLimit = Mth.clamp(channels - streamingLimit, 8, 255);
this.staticChannels = new Library.CountingChannelPool(staticLimit);
this.streamingChannels = new Library.CountingChannelPool(streamingLimit);
```

The limit of 8 is hard coded, so can't be changed via a config file. I'm not aware of any mod which increases it.

MCJack123:
I have some mod source code that changes it!
```java
package com.example.examplemod;

import com.mojang.logging.LogUtils;
import com.mojang.blaze3d.audio.Library;
import net.minecraft.client.Minecraft;
import net.minecraftforge.common.MinecraftForge;
import net.minecraftforge.eventbus.api.SubscribeEvent;
import net.minecraftforge.fml.common.Mod;
import net.minecraftforge.event.server.ServerStartingEvent;
import org.slf4j.Logger;

// The value here should match an entry in the META-INF/mods.toml file
@Mod("examplemod")
public class ExampleMod
{
    // Directly reference a slf4j logger
[...]

PR #2412, opened 2026-04-06
Fix speaker audio desync by removing server-side pacing and adding sample-offset sync

### LINKED ISSUE #1874: speakers.playAudio sometimes plays only on the next call
### Minecraft Version

1.21.x

### Version

1.111.1

### Details

Hello, thank you for your work on this amazing mod.

I discovered a weird issue with speakers. Executing `speakers.playAudio()` with certain timing has a chance to not play sound at all. The audio chunk gets postponed until next function call, at which speakers play 2 chunks one after another.

(speaker is driven properly, waiting for speaker_audio_empty e.t.c., just like in the wiki)
https://pastebin.com/2UDbQaZD

Video that demonstrates the issue:

https://github.com/cc-tweaked/CC-Tweaked/assets/18554021/01bd9332-6b84-425f-8fb3-dacbd3cc9dea

joviah:
## Summary

Fixes #1874 and #2120.

- **Removed server-side wall-clock pacing** in `DfpwmState`. Audio is now sent to the client on the next server tick after Lua pushes it. The existing single-slot buffer design provides natural flow control.
- **Added sample-offset timestamps** to `EncodedAudio` packets. The server tracks cumulative samples sent per stream.
- **Added client-side drift detection** in `DfpwmStream`. The client tracks consumed samples and compares against incoming offsets. Small drift self-corrects; drift beyond 500ms triggers a hard reset (queue clear + resume from latest packet).

### Why not improve the estimation?

The server fundamentally cannot model client playback accurately — network latency, sound engine scheduling, volume=0, and the 8-speaker Minecraft limit all cause the estimate to drift. Rather than building a better estimate, this PR removes the estimation entirely and lets the client self-correct using real data.

### Edge cases covered

- **Volume=0 (#2108):** When the sound engine doesn't play audio, `consumedSamples` stops advancing, drift grows, and a hard reset fires when playback resumes — discarding stale audio.
- **>8 speakers (#1313):** Same mechanism — dropped sounds cause drift, hard reset recovers.
- **Normal play (#1874):** Eager send eliminates the timing gap that caused audio to be held back too long.

## Test plan

- [x] Unit tests for `EncodedAudio` sampleOffset serialization roundtrip
- [x] Unit tests for `DfpwmState` sampleOffset tracking and `isPlaying()` timeout
- [x] Unit tests for `DfpwmStream` drift detection: normal flow, hard reset, stale packet drop, small drift, read stall, stream restart
- [x] Clean boot test on 1.20.1 Forge (no crashes or errors)
- [ ] Manual test: reproduction script from #1874
- [ ] Manual test: volume=0 then increase mid-stream
- [ ] Manual test: >8 speakers

## Files changed

- `EncodedAudio.java` — added `sampleOffset` field + VarLong serialization + `withSampleOffset()`
- `DfpwmState.java` — removed `clientEndTime`/`CLIENT_BUFFER`, added `sampleOffset` counter, simplified `shouldSendPending()`, updated `isPlaying()` to use timeout
- `SpeakerPeripheral.java` — updated `shouldSendPending()` call site
- `DfpwmStream.java` — added sync state (`consumedSamples`, `bufferedSamples`, `streamBaseOffset`), drift detection in `push()`, sample tracking in `read()`, synchronized access, debug logging
- Updated all 3 existing test files + added new test cases


SquidDev:
Ah, I had not realised you were planning to use Claude to write this. CC:T does not accept LLM-generated code — I have updated the contributing documentation to more accurately reflect this.

> The server tracks cumulative samples sent per stream. [...] Removed server-side wall-clock pacing in DfpwmState

I think tagging each audio packet with the expected time it will play, and then dropping/pad the audio buffer as needed makes sense — I've debated doing something similar with global timer in an attempt to fix #2412, but that's a bit trickier than using a per-stream timer. 

However, I think any approach here needs to use the time (or at least the sound engine's notion of time), rather than just the sample count. If the sound engine itself is not reading the the audio stream (which is the underlying problem of #2108 or #1313), then this fix does nothing.

I think the server has to remain the source of truth here — the implementation here allows you the server to send an infinite amount of audio data to the client, potentially OOMing them — it's just the way `DfpwmStream` handles providing data to the sound engine needs improving.

joviah:
No worries! Thanks!

PR #2414, opened 2026-04-09
Remove ability for turtles to be GPS hosts

ItsNotVingtdeux:
If a GPS host ever moves, it would no longer send back accurate data. This was accounted for pocket computers, but not turtles.


Wojbie:
I have seen cases on servers where players used turtles as GPS hosts because they are single block instead of 2 blocks with normal computer+modem. So there are use cases where someone would want turtle to be GPS host, as long as turtle is stationary.

It might be good idea to swap that from error out to a warning message in case of turtles?

ItsNotVingtdeux:
I'm swapping it out for a warning, how should the warning be phrased?

SquidDev:
Had a think about this, and I think I'm going to close this. I don't think this is a problem we see very often in practice, and there's enough times the error (or warning) is wrong, that I don't think it's worth it.

PR #2421, opened 2026-04-27
Recipe sync bug fix

Xenthera:
Fixes stream encoding for stateless custom recipes.
The old codec used `StreamCodec.unit(factory.get())`, which stores a single recipe instance and expects that exact object when encoding. In my dev environment, recipe sync could pass a different instance of the same stateless recipe, causing encoding to fail even though there was no recipe data to serialize.
This changes the stream codec to write no data and create a fresh instance on decode, matching the behavior of stateless recipes.

Steps I used to test:
- `./gradlew :forge:build`
- Tested the Forge jar in a NeoForge dev setup with LaserIO, Patchouli, and JEI installed
- Confirmed recipe sync no longer disconnects the client


Xenthera:
Sorry ignore this, forgot the branch

PR #2426, opened 2026-04-29
fix typo in shell.pathWithExtension

zyxkad:
`#sPath` is the length of global sPath. Its length have nothing to do with the method parameter `_sPath`. So mostly likely it is a typo.

This problem exists in other branches as well.


SquidDev:
Thanks for the PR! Hah, it looks like this has been broken since it was introduced 9(!) years ago in https://github.com/dan200/ComputerCraft/pull/211. However, looking at the surrounding logic, I don't think there's much reason for `pathWithExtension` at all — at this point, the path will never have a trailing slash, so the `if` will never be triggered. Instead, I've just removed it in abe3d3410fd446a2484383c8078bd9ef20b03aa4.

PR #2436, opened 2026-05-08
Fix `min` & `sec` field defaults in os.time

sircfenner:
When os.time is called with a date table, the `min` and `sec` fields are wrongly set to 12 if their values are missing. This looks like it was copied from the `hour` field default. They should both be 0 for parity with stock Lua.

### DIFF
--- a/projects/core/src/main/java/dan200/computercraft/core/apis/LuaDateTime.java
+++ b/projects/core/src/main/java/dan200/computercraft/core/apis/LuaDateTime.java
@@ -77,8 +77,8 @@ final class LuaDateTime {
         var month = getField(table, "month", -1);
         var day = getField(table, "day", -1);
         var hour = getField(table, "hour", 12);
-        var minute = getField(table, "min", 12);
-        var second = getField(table, "sec", 12);
+        var minute = getField(table, "min", 0);
+        var second = getField(table, "sec", 0);
         var time = LocalDateTime.of(year, month, day, hour, minute, second);
 
         var isDst = getBoolField(table, "isdst");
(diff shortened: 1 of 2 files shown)


SquidDev:
Thank you for the fix! Would it be possible to add a test case [to `os_stub.lua`](https://github.com/cc-tweaked/CC-Tweaked/blob/defb713483285a41f9b791458252814376fe9440/projects/core/src/test/resources/test-rom/spec/apis/os_spec.lua#L39), just to check the default values are handled correctly/consistently with PUC Lua?

sircfenner:
I discovered that `os.time` has a further deviation from PUC Lua - it seems to apply a timezone offset (in my case of -3600 for BST). PUC Lua does not appear to do this. This should probably be addressed so that we can add a meaningful and correct test case. Unfortunately I don't have the bandwidth to contribute a fix for that issue.

Here is a test case which addresses both issues:
```Lua
it("returns 0 for unix epoch", function()
	local t = os.time({
		year = 1970,
		month = 1,
		day = 1,
		hour = 0,
	})
	expect(t):eq(0)
end)
```

SquidDev:
I think that's the right behaviour — I get -3600 (also in BST) on PUC Lua. I think it always uses your local timezone, and you can't force UTC, even with setting `isdst = false` does nothing.

I think the easiest thing to do is then to pass it to `os.date` immediately (either with `os.date("%F %T, ...)` or `os.date("*t", ...)`), which also uses the local TZ by default, so should produce a stable result across timezones.

sircfenner:
Thanks for checking this - I think my mistake was testing timezone behavior with an online Lua compiler. I've added now-passing test cases for the default values in date table. Looks like there is already a roundtrip test using the os.date approach you suggested, so I haven't added an extra test for that.

SquidDev (approved):
Amazing, thank you so much!

PR #2458, opened 2026-06-16
rednet: Allow sending through selected modems instead of every modem

sylfn:
## A quick checklist
 - [x] If you're working on CraftOS, try to write a few test cases so we can ensure everything continues to work in the future. Tests live in `src/test/resources/test-rom/spec` and can be run with `./gradlew check`.

---

The description of this PR has been edited to match the contents. See edit history for previous revisions and diffs between them.

No LLMs or "AI" were used in creation of this PR, only natural stupidity. Mistakes are my own.

## rednet: Allow sending through selected modems instead of every modem

Consider the following code (a simplified excerpt from [netbootd](https://github.com/purplesyringa/computercraft-programs/blob/63fe7c4c79f8d87a3f7f4cb6e4276d75f47e42f0/sys/packages/netboot/bin/netbootd.lua#L55), a daemon that broadcasts startup files):

```lua
while true do
    local _, side = os.pullEvent("peripheral")
    rednet.open(side)
    rednet.broadcast("meow!")
end
```

This code broadcasts the message to every opened modem, even though the intention might have been to broadcast the message only to newly-connected one. Sending to every modem instead of one in this case makes code complexity quadratic instead of linear over total number of peripherals.

---

No changes got mentioned in documentation under "Changed in a.b.c:", but I don't know what is the correct way to add these


Wojbie:
> rednet: Track opened modems

This does feel like it would be a breaking change? I am unsure if increasing complexity by adding logic side state that might disagree with real world and adding edge cases is worth this efficiency boost.

> rednet: Allow sending through a single modem instead of every modem

This is interesting addition, on one side, it makes sense for user to desire to only send rednet message out specific modem if they are sure that is side/network their target is on, on other side, send on all modems logic makes sense to ensure message has highest chance to reach recipient either directly, or via routed repeat message thru repeaters. Those duplicated messages are ignored and discarded by both repeat and rednet logic later on via their unique id checks.

sylfn:
Re "Track opened modems":

The only breaking change I know of is outlined in the description (that is, manually opening required channels on the modem. I think the state tracking (which I did with a "cache" approach; or even worse one with "look for `peripheral`/`peripheral_detach` in `rednet.run`") is not a great choice for keeping the code simple.

Re "Allow sending through a single modem instead of every modem":

I have another idea: instead of allowing only a single modem, it might be better to allow supplying a list of modems. This also allows users to track modems themselves (removing the `peripheral.getNames` cost in this case) while allowing the default behavior to be "reach everything" and removing the weird state-tracking that was added in my previous commit.

I agree with "reach everything" logic being sane, and have kept it as the default behavior, airgapping is intended for more niche use cases.

---

I'm going to drop "rednet: Track opened modems" commit (keeping tests for it in place, as I don't think they're harmful) and allow supplying a list of modems to `rednet.send`/`rednet.broadcast`

PR #2488, opened 2026-07-26
changelog: document sleep()'s optional parameter, starting in 1.63

tomodachi94:
Had to do some archaeology for this one! 1.62 is missing this string:

    local timer = os.startTimer( nTime or 0 )

which is only present in 1.63 and later.

## Rationale

Users of older versions (like me!) still use tweaked.cc for documentation. It's good to know what I can and can't use when playing, say, Tekkit 1.6.4[^1]

[^1]: Tekkit uses CC 1.58. Don't ask me why it doesn't use a newer version that also supports 1.6.4 like, say, CC 1.63...

### DIFF
--- a/doc/stub/global.lua
+++ b/doc/stub/global.lua
@@ -37,6 +37,8 @@ nearest multiple of 0.05.
     print("Sleeping for three seconds")
     sleep(3)
     print("Done!")
+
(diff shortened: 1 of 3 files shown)


SquidDev:
Thanks!

PR #2495, opened 2026-08-14
Fix random indentation of two lines in bios.lua

ItsNotVingtdeux:
The "type" field definitions for the "edit.autocomplete" and "lua.autocomplete" settings were both randomly indented, compared to the rest of the setting definitions.


SquidDev:
Thanks!

PR #2505, opened 2026-09-11
Lower division count for vector lib, use string.format for __tostring

9551-Dev:
A small change i thought of recently, i was kinda debating adding some way to edit vectors in-place because that would probably allow for a significant performance improvement, but i feel like my ways of coding and programming are not really fit of base CraftOS (my idea was to include a inplace table on all vectors that would just link to a different list of functions that modify the source vector itself, those would actually be the base functions and the ones that instantiate them would be made using a wrapper loop)

Hopefully using string.format shoulnt cause any back compat issues formatting wise, afaik it just calls to tostring with %s

## A quick checklist


SquidDev:
Thanks for the PR! 

I've no objections with the change, but just wanted to understand the motivation for it. Is this purely for performance, and if so, do you have any benchmarking figures? On a low-level, division *is* faster than multiplication, but for Lua/Cobalt my gut feeling would have been that the overhead of the extra instruction dwarfs any gains!

9551-Dev:
> Thanks for the PR!
> 
> I've no objections with the change, but just wanted to understand the motivation for it. Is this purely for performance, and if so, do you have any benchmarking figures? On a low-level, division _is_ faster than multiplication, but for Lua/Cobalt my gut feeling would have been that the overhead of the extra instruction dwarfs any gains!

this is sadly very possible, i was reading through the code and just found it icked me in the worng way, i still feel like having a way to modify in-place could gain actual large performance for use cases of those who need it, and i often find myself needing to use my own vector library because of that alone

this makes it also more consistent as the reciprocal is already used in normalization calculation

id be willing to implement that along with the ability to do simple matrix multiplication but i am not sure if there would be any interest in that

SquidDev:
Honestly, if you're trying to do high-performance linear algebra (well, as much as you can in CC), `vector` is a terrible place to start. Ignoring in-place updates, it's also using string keys instead of integer keys (slower lookups, higher memory usage), and is hard-coded to 3 dimensions (so if you're doing 3D graphics, you'll need to write your own for 4D vectors anyway).

If you're really trying to juice out performance, I suspect the best thing to do is actually drop using tables entirely, and keep everything in locals.

9551-Dev:
> If you're really trying to juice out performance, I suspect the best thing to do is actually drop using tables entirely, and keep everything in locals.

this is what i did for my own stuff using a custom Lua preprocessor, also are string lookups really slower than numeric ones?, or is it specifically array table portion lookups, 1-to first nil

as for how much you can use linear algebra in CC, there is surprising amount of stuff you can do even without any absurd optimization, well that ive come to be able to write over the years atleast


https://github.com/user-attachments/assets/12990584-5347-4b04-8ad0-23ca42cc8813

SquidDev:
> also are string lookups really slower than numeric ones?, or is it specifically array table portion lookups, 1-to first nil

So *in general* array lookups will be faster than hash lookups, and a non-string key will be faster to lookup in the hash part than a string key. This is because all other types compare a primitive value (either their contents or a pointer), but strings need to compare the contents of the string.

PR #2507, opened 2026-09-20
Sanity checking fixes for shell.run and shell.lua tokenise

9551-Dev:
Should prevent such errors, while keeping backwards compatibility, this technically makes it not match documentation which lists shell.run to only take strings, however i think its better to have it like this than to have it be inconsistent in the docs between shell.run and shell.execute

image


9551-Dev (commented):
<deleted>

projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
```diff
@@ -85,8 +85,21 @@ else
...
 end
 
+local tokenisableTypes = {
+    ["string"] = true,
+    ["number"] = true,
+}
+
 local function tokenise(...)
-    local sLine = table.concat({ ... }, " ")
+    local sArgs = { ... }
```
SquidDev: What's the aim of this bit? If we're sanitising in `shell.run`, then isn't this a no-op?

SquidDev (changes requested):
It's probably worth having a test for this in [`shell_spec.lua`](https://github.com/cc-tweaked/CC-Tweaked/blob/mc-1.20.x/projects/core/src/test/resources/test-rom/spec/programs/shell_spec.lua), like we do for the other `expect` calls (e.g. see `shell.resolveProgram`).

projects/core/src/main/resources/data/computercraft/lua/rom/programs/shell.lua
9551-Dev: Currently yes but i thought it would be good to make tokenise not be able to cause that error at all in case someone were to change something up the ladder or use it in some other part of the code, it would avoid potentially "unexpected" behavior, without having almost any extra cost in code like this

SquidDev: Thinking about it a bit more, I think it's actually a bit cleaner to change tokenise to only accept a single argument, and then handle the concatenation in `shell.run`. I think the other callers only pass a single argument?

9551-Dev: will do

9551-Dev:
I tried to implement those things but i am not entirely sure why the tests are failing, the errors are weird

SquidDev (approved):
> `if not tokenisableTypes[value] then counterLock = true end`

You were missing the `type(value)` here, so the loop immediately started rejected values, thus `shell.run` never actually did anything!

I've simplified the logic a bit here. You don't need to account for `nil`, as that was already broken. Or rather, it's tied to the behaviour of table length, which is unspecified for `nil`s (e.g. `shell.run(1, nil, 3)` errors, but `shell.run(1, 2, nil)` "works").

SquidDev:
> trailing comma is some odd requirement but suuuree 

For reference, this is pretty common in most formatters! The motivation is that it makes the diff more readable when adding new elements. Compare:

```diff
 {
-  "foo"
+  "foo",
+  "bar"
 }
```

vs 

```diff
 {
   "foo",
+  "bar",
 }
```

9551-Dev:
> > `if not tokenisableTypes[value] then counterLock = true end`
> 
> You were missing the `type(value)` here, so the loop immediately started rejected values, thus `shell.run` never actually did anything!
> 
> I've simplified the logic a bit here. You don't need to account for `nil`, as that was already broken. Or rather, it's tied to the behaviour of table length, which is unspecified for `nil`s (e.g. `shell.run(1, nil, 3)` errors, but `shell.run(1, 2, nil)` "works").

Ohh ooops, sorry, ill need to setup a better testing environment next time ^^, and yea it was a bit much probably, i was trying to fully mimic original behavior even if some of it wasnt intented, thanks :>

